PR #1849, opened 2014-03-09
Ensure update_manifests celery task is retried (Bug 928462)

nearlyfreeapps:
As suspected in [Bug 928462](https://bugzilla.mozilla.org/show_bug.cgi?id=928462), it seems like by catching the worker retry exception, the retry is not being properly scheduled for an hour in the future.


ddurst:
I think this is passing now, @robhudson ?

robhudson:
No idea. It's a complicated bug and complicated fix.

PR #2648, opened 2014-10-10
Updates required to work on Django 1.7

robhudson:


mkt/settings.py
```diff
@@ -1001,8 +1001,13 @@ def JINJA_CONFIG():
 READ_ONLY = False
 
 REDIS_BACKENDS = {
```
diox: `REDIS_BACKENDS` is also defined in `sites/*`, and they use `private` settings. I assume thos are not using the dict format either and will fail with the update...

robhudson: Yep, I need to come back to this. I think I will copy the single string parser that Django 1.7 removed into our codebase.

diox: Does the dict-like syntax works now with 1.6 ? If it does, let's convert everything now and then we don't have to have the single string parser ?

robhudson: The nice thing about the single string parse is you can easily override this with an environment variable, which is how it is currently on master for docker. It's something I broke here and need to fix.

manage.py
```diff
@@ -30,16 +30,6 @@
...
 log_configure()
 
-# We need to import waffle here to avoid a circular import with jingo which
-# loads all INSTALLED_APPS looking for helpers.py files, but some of those apps
-# import jingo.
-import waffle  # flake8: noqa
-
-# Hardcore monkeypatching action.
-import jingo.monkey
-jingo.monkey.patch()
```
andymckay: i'm glad to see this go, we should check receiptcheck still works though

mkt/settings.py
andymckay: For docker, all we need is the host, never seen any of other stuff change. Just setting `REDIS_HOST` as an environment variable would work for me.

mkt/developers/views.py
```diff
@@ -553,6 +553,7 @@ def ownership(request, addon_id, addon):
                     thread___addon=addon).delete()
 
         for author in user_form.deleted_objects:
+            author.delete()
```
andymckay: why?

mkt/urls.py
```diff
@@ -18,6 +18,11 @@
 from mkt.users.views import logout
 
 
+# Hardcore monkeypatching action.
+import jingo.monkey
+jingo.monkey.patch()
```
andymckay: will we need to move the import waffle here? thinking of receiptcheck primarily.

mkt/developers/views.py
robhudson: That may be sloppy. I'll have to look at that again. We don't want to delete the UserProfile object, but the object that associates a user with an admin.

The reason I believe is this:
https://docs.djangoproject.com/en/dev/releases/1.7/#miscellaneous (bullet point 2)

mkt/settings.py
robhudson: Ok, cool, thanks @andymckay

mkt/urls.py
robhudson: I'm not sure, we'd have to test. When I did this it worked w/o the extra import but I didn't check receiptcheck.

mkt/developers/views.py
andymckay: Oh this is the AddonUser model that's being deleted. That should be ok, I was worried about deleting something that cascaded with terrible consequences.

robhudson: I'd like to verify that but yes. I should name the variable `app_author` or something instead.

robhudson:
Rebased. I'm going to open this as ready for review.

mkt/settings.py
```diff
@@ -959,7 +959,7 @@ def JINJA_CONFIG():
     # Stops nose always adding lib into the path, which then breaks crypto
     # imports.
     '-P',
-    '--with-fixture-bundling',
+#    '--with-fixture-bundling',
```
diox: Why ? That's probably bad for test perf.

mkt/settings.py
```diff
@@ -1006,8 +1006,13 @@ def JINJA_CONFIG():
 READ_ONLY = False
 
 REDIS_BACKENDS = {
-    'master': os.environ.get('REDIS_URL',
```
diox: I don't remember what was the conclusion of the discussions on this ; don't you need to patch the settings in `sites/` too ?

diox:
Tests pass but with `django.db.utils.OperationalError: (2006, 'MySQL server has gone away')` at the end so they are marked as failed...

robhudson:
Yeah, and I haven't figured out why. But I haven't looked too hard yet.

robhudson:
Closing PR to move branch to the mozilla origin.

PR #2791, opened 2014-12-19
Reviewer onboarding tool (bug 1101096)

washort:
Outstanding issue is IARC rating and further app data.


washort:
Doing this via the API is probably needlessly complicated if the code in https://github.com/mozilla/zamboni/pull/2862 can be used instead.

eviljeff:
If you think its easier to overload the code in #2862 to make it work for this too, then sure. But its a slightly different use case - we need real apps that install (that we can specify the manifest or package for); have screenshots and genuine icons (either specified, or pulled from the manifest/package); different platforms; buchets - the fake QA apps don't appear to need (yet!)

PR #2803, opened 2014-12-23
remove whitelist

andymckay:


ddurst:
I agree this is a bit superfluous and may disagree with the original premise (especially when I think about "whitehat" and "blackhat" conventions), but I do think "allowed" and "blocked" is clearer and thus better, so I'm down with it. Purple r+

PR #2805, opened 2014-12-24
ensure that 404 and 500 have CORS for the API, including DEBUG mode (bug...

andymckay:
... 970641)

### DIFF
--- a/mkt/api/middleware.py
+++ b/mkt/api/middleware.py
@@ -262,17 +262,22 @@ class CORSMiddleware(object):
             getattr(request, 'CORS_HEADERS',
                     ('X-HTTP-Method-Override', 'Content-Type')))
 
-        if fireplacey or getattr(request, 'CORS', None):
+        error_allowed_methods = []
+        if response.status_code >= 400 and request.API:
+            error_allowed_methods = [request.method]
+
+        cors_allowed_methods = getattr(request, 'CORS', error_allowed_methods)
+        if fireplacey or cors_allowed_methods:
             # If this is a request from our hosted frontend, allow cookies.
             if fireplacey:
                 response['Access-Control-Allow-Origin'] = fireplace_url
                 response['Access-Control-Allow-Credentials'] = 'true'
             else:
                 response['Access-Control-Allow-Origin'] = '*'
-            options = [h.upper() for h in request.CORS]
-            if 'OPTIONS' not in options:
-                options.append('OPTIONS')
-            response['Access-Control-Allow-Methods'] = ', '.join(options)
+            methods = [h.upper() for h in cors_allowed_methods]
+            if 'OPTIONS' not in methods:
+                methods.append('OPTIONS')
+            response['Access-Control-Allow-Methods'] = ', '.join(methods)
 
         # The headers that the response will be able to access.
         response['Access-Control-Expose-Headers'] = (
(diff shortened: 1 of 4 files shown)


mkt/api/middleware.py
```diff
@@ -262,17 +262,27 @@ def process_response(self, request, response):
             getattr(request, 'CORS_HEADERS',
                     ('X-HTTP-Method-Override', 'Content-Type')))
 
-        if fireplacey or getattr(request, 'CORS', None):
+        error_allowed_methods = []
+        # If settings.DEBUG is True, these skip the 404 and 500 handlers
```
diox: Why only for debug ? This happens in production as well, so shouldn't we always do this ?

diox: Ah right, had not seen that. Aren't middlewares also run anyway when we go through `handler_*` ? If so, any reason to do the job in different places instead of consolidating everything in the middleware ?

mkt/site/tests/test_views.py
```diff
@@ -66,6 +66,13 @@ def test_404_api(self):
         res = self.client.get('/api/this-should-never-work/')
         eq_(res.status_code, 404)
         eq_(res.content, '')
+        self.assertCORS(res, 'get')
+
+    def test_404_api_debug(self):
+        with self.settings(DEBUG=True):
+            res = self.client.get('/api/this-should-never-work/')
+            eq_(res.status_code, 404)
+            self.assertCORS(res, 'get')
```
kumar303: I've noticed some subtle bugs where options requests are not handled correctly and that's the first thing the browser does before any cors logic. So it would be good to add this in all tests like:

    res = self.client.options('/api/this-should-never-work/')
    self.assertCORS(res, 'options')

mkt/site/views.py
```diff
@@ -43,6 +43,7 @@ def handler403(request):
 def handler404(request):
     if request.path_info.startswith('/api/'):
         # Pass over to API handler404 view if API was targeted.
+        request.CORS = [request.method]
```
kumar303: hmm, I wonder if we need CORS headers too.

mkt/api/middleware.py
andymckay: They are and that's seems a valid point and would reduce the lines of code, I'll make that change.

mkt/site/tests/test_views.py
andymckay: assertCORS already does that: https://github.com/mozilla/zamboni/blob/master/apps/amo/tests/__init__.py#L508

mkt/site/views.py
kumar303: probably not

mkt/site/tests/test_views.py
kumar303: Yes but making an actual options request like ``self.client.options(...)`` can execute completely different code yielding different response headers so it's worth adding that request in a test. For example, DRF won't execute the original view at all when you do an options request so sometimes it won't hit any decorators like you'd expect.

mkt/api/middleware.py
```diff
@@ -262,17 +262,27 @@ def process_response(self, request, response):
...
+        # where we set CORS - this happens down in django.core.handlers.
+        #
+        # By setting them here, we ensure developers get the appropriate
+        # headers when testing locally.
+        if (response.status_code in [404, 500] and settings.DEBUG
```
andymckay: Just wondering, zamboni returns quite a few different error codes, would it make sense to do everything 4xx?

diox: Yes. At least 403, 405.

mkt/site/tests/test_views.py
andymckay: Ok, test updated.

kumar303:
looks good. r+

PR #2813, opened 2014-12-31
remove blacklist

andymckay:
there's two main references remaining, one from app-validator and one in private settings


robhudson:
Scanned it, looked good. r+

PR #2818, opened 2015-01-05
Filter on `lang` in ratings API with `match_lang` (bug 875455)

mstriemer:
This is going to need a migration on the reviews table to store the user's lang, but getting the PR in.

- [x] Create migration for lang on reviews.
- [x] Set lang from user's lang on create.
- [x] Use the lang column in API.
- [ ] Populate the lang from user's lang in task.
- [ ] Update docs/tests to no longer support `match_lang=0`.

### DIFF
--- a/docs/api/topics/ratings.rst
+++ b/docs/api/topics/ratings.rst
@@ -21,6 +21,11 @@ _`List`
 
     :query app: the ID or slug of the app whose ratings are to be returned.
     :query user: the ID of the user or `mine` whose ratings are to be returned.
+    :query lang: a language to filter ratings by if `match_lang` is set.
+    :query match_lang: a boolean to specify to match langauge or not. If unset
+                       all results are returned, if 1 only results matching
+                       `lang` are returned if 0 only results not matching
+                       `lang` are returned. All other values are ignored.
 
     The value `mine` can be used to filter ratings belonging to the currently
     logged in user.
@@ -41,7 +46,8 @@ _`List`
             },
             "info": {
                 "average": "3.4",
-                "slug": "marble-run"
+                "slug": "marble-run",
+                "total_reviews": 391
             },
             "objects": [
                 {
@@ -49,6 +55,7 @@ _`List`
                     "body": "This app is top notch. Aces in my book!",
                     "created": "2013-04-17T15:25:16",
                     "is_author": true,
+                    "lang": "en-US",
                     "modified": "2013-04-17T15:34:19",
                     "rating": 5,
                     "resource_uri": "/api/v2/apps/rating/19/",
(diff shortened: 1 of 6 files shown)


docs/api/topics/ratings.rst
```diff
@@ -21,6 +21,11 @@ _`List`
 
     :query app: the ID or slug of the app whose ratings are to be returned.
     :query user: the ID of the user or `mine` whose ratings are to be returned.
+    :query lang: a language to filter ratings by if `match_lang` is set.
+    :query match_lang: a boolean to specify to match langauge or not. If unset
+                       all results are returned, if 1 only results matching
+                       `lang` are returned if 0 only results not matching
+                       `lang` are returned. All other values are ignored.
```
robhudson: Just wondering if we could get away with a single value here, which is `lang`. If provided, we simply return only ratings matching the language. If not provide, we return all. I'm not sure I see a use-case for `match_lang=0`?

mkt/ratings/tests/test_views.py
```diff
@@ -281,6 +282,42 @@ def test_already_rated_version(self):
...
+        Review.objects.create(addon=self.app, user=self.user, body='yes')
+        other_user = UserProfile.objects.exclude(pk=self.user.pk)[0]
+        other_user.update(lang='pt-BR')
+        Review.objects.create(addon=self.app, user=other_user, body='yes')
+        res, data = self._get_url(self.list_url, app=self.app.app_slug,
+                                  client=self.anon, lang='pt-BR',
+                                  match_lang='0')
+        eq_(res.status_code, 200)
+        eq_(len(data['objects']), 2)
+        eq_(data['info']['total_reviews'], 2)
```
robhudson: Maybe add a test for `match_lang='0'`?

robhudson: Oh wait, that's what the above test is... so why doesn't it return 1 review only since it should exclude the `pt-BR` one?

mkt/ratings/views.py
```diff
@@ -67,17 +68,21 @@ def filter_queryset(self, queryset):
...
+        match_lang = self.request.GET.get('match_lang')
         if app:
             self.app = self.get_app(app)
-            filters['addon'] = self.app
+            filters &= Q(addon=self.app)
         if user:
-            filters['user'] = self.get_user(user)
+            filters &= Q(user=self.get_user(user))
+        elif lang and match_lang == '1':
+            filters &= Q(user__lang=lang)
```
robhudson: I'm looking for the exclude case when `match_lang == '0'`?

docs/api/topics/ratings.rst
mstriemer: It is essentially just there to maintain backwards compatibility. We are currently sending `region` and `lang` on each request from fireplace so we could get some weird behaviour if we start filtering on `lang`.

mkt/ratings/views.py
mstriemer: Yeah, I must have removed it. I was originally using both the `1` and `0` case but now I'm just using the `1` case. I'll update the docs and remove that test since it isn't being used.

diox: AFAIK `user.lang` is not indexed so this will be costly... I think it'd be better to run the migration to populate `lang` on the `Review` model first, and then use that directly in this PR.

PR #2820, opened 2015-01-06
Fix bug in trending per region calculation

robhudson:


mkt/webapps/tasks.py
```diff
@@ -740,7 +745,7 @@ def _score(week1, week3):
         for regional_res in res['aggregations']['region']['buckets']:
             region_slug = regional_res['key']
             week1 = regional_res['week1']['total_installs']['value']
-            week3 = regional_res['week1']['total_installs']['value'] / 3.0
+            week3 = regional_res['week3']['total_installs']['value'] / 3.0
```
robhudson: The bug. Copy-paste fail.

mkt/webapps/tasks.py
```diff
@@ -715,6 +715,10 @@ def _get_trending(app_id):
         return {}
 
     def _score(week1, week3):
+        # If last week app installs are < 100, this app isn't trending.
+        if week1 < 100.0:
```
andymckay: 100.0 seems like a rather magic number, any reason for that choice?

robhudson: I just made it up. We'll likely want to increase it as app installs increase. I modeled all this after AMO's "hotness" calculations which drives its "Up & Coming" add-ons: https://github.com/mozilla/olympia/blob/74364be6dce33d96220493706451951bdbeb8151/apps/addons/cron.py#L322

But I should make that a constant so it's easy to tweak.

robhudson: The idea is you don't want apps with a few installs to be considered as "trending". There's got to be some threshold.

andymckay: Ok, if there's some comments where it came from and that we can tweak it, that's cool. Just wasn't sure myself.

mkt/webapps/tasks.py
```diff
@@ -715,6 +715,10 @@ def _get_trending(app_id):
...
+            return 0.0
+
         score = 0.0
         if week3 > 1:
```
andymckay: This will work just fine, but maybe good to be consistent as float all the way through though.

mkt/webapps/tests/test_crons.py
```diff
@@ -326,11 +335,53 @@ def test_get_trending_negative(self, _mock):
...
+        client = mock.Mock()
+        client.raw.return_value = self._return_value_with_regions(102, 102,
+                                                                  100, 1000)
+        _mock.return_value = client
+
+        # 1st week regional count: 99
+        # Prior 3 week regional count: 99/3 = 33
+        # (99 - 33) / 33 = 2.0 but week1 isn't > 100 so we set to zero.
+        eq_(_get_trending(self.app.id)['all'], 2.0)
+        eq_(_get_trending(self.app.id)['br'], 0.0)
```
andymckay: Thanks for the comments, those really helped me.

andymckay:
r+wc

robhudson:
Merged: https://github.com/mozilla/zamboni/commit/1ff38c75f1423f06edc4b341321f7db1cd8138a9 (and 3 prior commits)

PR #2821, opened 2015-01-06
Use Travis-ci's container infrastructure; bug 1118887

clouserw:


.travis.yml
```diff
@@ -1,20 +1,15 @@
 language: python
+sudo: false
 python:
   - "2.6"
 services:
-  - elasticsearch
   - redis-server
 before_install:
-  - sudo apt-get update
-  - sudo apt-get install swig pngcrush
```
andymckay: there is a test that depends upon pngcrush i believe

clouserw: there is.  this branch is experimentation

diox: It should be fairly simple to install pngcrush locally without sudo - it doesn't require anything fancy, you should be able to find prebuilt binaries.

mkt/feed/tests/test_views.py
```diff
@@ -924,23 +924,6 @@ def test_background_image(self, download_mock, crush_mock):
         eq_(crush_mock.call_args_list[0][1]['set_modified_on'][0], coll)
 
     @mock.patch('mkt.feed.fields.requests.get')
-    def test_background_image_crush(self, download_mock):
```
clouserw: @diox I deleted this test completely.  The other tests in this same file already mock out the pngcrush call so this just seemed redundant.

.travis.yml
clouserw: For posterity:  We decided that mocking pngcrush was more appropriate for our tests.  Thanks to mat for the patch.

requirements/compiled.txt
```diff
@@ -5,7 +8,7 @@ Pillow==2.3.0
 
 # For JWT to sign App receipts:
 # Temporary.
-M2Crypto==0.20.2
+M2Crypto==0.22.3
```
andymckay: Is that on prod boxes? We should use the same as is there.

andymckay:
Other than M2Crypto question, r+wc. Thanks for speeding things up!

requirements/compiled.txt
clouserw: mmmm, yeah.  That's what was available as a .whl so I upgraded it.  The tests pass - we can consider this just upgrading, right?  (So it will be QA'd and pushed as usual, since I changed the requirements file)

clouserw: https://bugzilla.mozilla.org/show_bug.cgi?id=1118999

PR #2822, opened 2015-01-07
Purge old trending records

robhudson:


robhudson:
https://github.com/mozilla/zamboni/commit/5b5a165

PR #2829, opened 2015-01-08
Generate fake user data (bug 1119543)

washort:


mkt/api/models.py
```diff
@@ -60,7 +71,3 @@ class Meta:
         db_table = 'oauth_nonce'
         unique_together = ('nonce', 'timestamp', 'client_key',
                            'request_token', 'access_token')
-
-
-def generate():
-    return os.urandom(64).encode('hex')
```
robhudson: Looks like this broke something and tests failed.

washort: Tests now pass.

robhudson:
Looks like it needs a rebase as well. I can take a close look before finishing up today.

washort:
rebased.

mkt/users/tasks.py
```diff
@@ -0,0 +1,54 @@
+import datetime
+from django.db import transaction
+
+import mkt
+from mkt.access.models import Group, GroupUser
+from mkt.api.models import Access
+from mkt.users.models import UserProfile
+
+
```
robhudson: This is in the `tasks.py` file but it's not using celery tasks?

mkt/users/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+        else:
+            if oauth_key or oauth_secret:
+                raise ValueError("Specify both of oauth_key and oauth_secret, "
+                                 "or neither")
+            Access.create_for_user(profile)
+
+
+def generate_users(n):
+    for i in range(n):
+        create_user('testuser%s@example.com' % (i,), "Test User %d" % (i,))
```
robhudson: It doesn't look like this is used anywhere?

mkt/users/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+from django.db import transaction
+
+import mkt
+from mkt.access.models import Group, GroupUser
+from mkt.api.models import Access
+from mkt.users.models import UserProfile
+
+
+@transaction.commit_on_success
+def create_user(email, group_name=None, overwrite=False,
```
robhudson: Some tests around `create_user` would be good.

robhudson:
Looks ok. Will there be a way later to associate these users with apps or anything?

mkt/users/tasks.py
washort: I'm not particular about where it should go; suggestions?

washort:
What features would you like wrt associating users with apps?

mkt/users/tasks.py
robhudson: maybe mkt/users/utils.py? That exists already. Celery reads in the tasks.py file of each app looking for tasks so it seems odd to be here.

robhudson:
> What features would you like wrt associating users with apps?

I was just thinking that a test developer should own a test app, and then wondering how that factors into the fake data.

washort:
Associating users with fake apps should probably be done in followup to #2862.

washort:
test added

robhudson:
r+

PR #2831, opened 2015-01-08
Respect existing data for load_prices command

kumar303:
@andymckay r?


andymckay:
r+ with the comments in IRC that we should drop this command and stop supporting it, since Docker loads init data which sets it up correctly and @washort has a FakeData framework. This command made sense once, don't think it does any more.

kumar303:
in the case where Docker doesn't work (now or in the future) it's nice to have an option to run things manually though

andymckay:
I don't think have Docker not working is going to be an option when we start deploying to prod using Docker :)

washort:
I expect that whatever fake-data scripting we do will probably just call this code or something very like it.

PR #2836, opened 2015-01-12
stop hiding packaged tab and keep order consistent between free and paid

andymckay:

### DIFF
--- a/media/js/devreg/submit.js
+++ b/media/js/devreg/submit.js
@@ -81,7 +81,6 @@
             $this.find('input').prop('checked', nowSelected);
             $input.val(old).trigger('change');
             $compat_save_button.removeClass('hidden');
-            setTabState();
         })
     );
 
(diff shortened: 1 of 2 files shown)


andymckay:
...what a terrible gif conversion experience that was

media/js/devreg/submit.js
```diff
@@ -105,6 +105,8 @@
         return $('#id_free_platforms option:selected').length ||
             $('#id_paid_platforms option[value="paid-firefoxos"]:selected').length ||
             $('#id_paid_platforms option[value="paid-android-mobile"]:selected').length ||
+            $('#id_paid_platforms option[value="paid-android-tablet"]:selected').length ||
+            $('#id_paid_platforms option[value="paid-desktop"]:selected').length ||
```
muffinresearch: I think you can replace all of these lines to match the first line (as it is for free) now so that rather than checking for all the specfic paid values you look for if any paid platforms are selected.

e.g:

`#id_paid_platforms option:selected`

muffinresearch:
r+wc

media/js/devreg/submit.js
andymckay: Actually is there a condition where we want to hide the packaged app tab any more? Thinking I can just remove this whole function.

muffinresearch: Even better!

PR #2837, opened 2015-01-13
Avoid need for reindexing when adding new buchets (bug 1111893)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1111893

~~Probably needs more tests.~~

@robhudson can you shoot holes in this approach ? We discussed doing it with `missing` queries, but any reason not to do this with `must_not` ? It appears to work for me, missing features are not `True` so they are never taken into consideration.

### DIFF
--- a/mkt/search/tests/test_views.py
+++ b/mkt/search/tests/test_views.py
@@ -863,6 +863,23 @@ class TestSearchViewFeatures(RestOAuth, ESTestCase):
         obj = json.loads(res.content)['objects'][0]
         eq_(obj['slug'], self.webapp.app_slug)
 
+    @patch('mkt.webapps.models.AppFeatures.to_dict')
+    def test_new_unused_feature_doesnt_require_reindex(self, mock_to_dict):
+        """Test that we still show apps even when they don't have the latest
+        list of features in the index, i.e. that we don't need to reindex apps
+        that are not using new features when we add some."""
+        # Mock what's returned when we fetch the features to specifically avoid
+        # having a value for the "new" feature.
+        mock_to_dict.return_value = {}
+        self.webapp.save()
+        self.refresh('webapp')
+        res = self.client.get(self.url, data=self.qs)
+        eq_(res.status_code, 200)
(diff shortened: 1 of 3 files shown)


mkt/webapps/indexers.py
```diff
@@ -532,9 +527,18 @@ def get_app_filter(cls, request, additional_data=None, sq=None,
             should = [es_filter.Terms(id=list(set(app_ids)))]
             sq = sq[0:len(set(app_ids))]
 
+        # MUST NOT.
+        must_not = []
+        if data['profile'] and not no_filter:
```
robhudson: Seems like changing this from a double negative would make it clearer. Except `filter` is a keyword.

mkt/search/tests/test_views.py
```diff
@@ -863,6 +863,23 @@ def test_bad_profile_on_desktop(self):
...
+        # having a value for the "new" feature.
+        mock_to_dict.return_value = {}
+        self.webapp.save()
+        self.refresh('webapp')
+        res = self.client.get(self.url, data=self.qs)
+        eq_(res.status_code, 200)
+        objects = json.loads(res.content)['objects']
+        eq_(len(objects), 1)
+        obj = objects[0]
+        eq_(obj['slug'], self.webapp.app_slug)
```
robhudson: Yeah, a test on `get_app_filter` itself would be nice.

robhudson:
I can dig it. It's simpler than the empty query so if that works I'm :+1:

mkt/webapps/indexers.py
diox: Mmmm feed uses `filtering` query parameter (but does not store it in a variable)

diox: I'll keep `no_filter` for now: there are a few files passing this parameter and I don't want to increase the potential for making a mistake in that commit.

diox:
@robhudson new version for you to look at, with more tests and a refactored `must_not` clause.

robhudson:
Thanks for the heads up. Very nice tests. r+

PR #2839, opened 2015-01-13
Moderation log viewer; soft deleted ratings & undelete (bug 1067726)

eviljeff:
This is mainly back ported from olympia, then fixed, etc.  Could do with some UI love but its not really worth the effort.


mkt/ratings/models.py
```diff
@@ -18,10 +19,30 @@
...
+        qs = qs._clone(klass=ReviewQuerySet)
+        if not self.include_deleted:
+            qs = qs.exclude(deleted=True).exclude(reply_to__deleted=True)
+        return qs
+
     def valid(self):
         """Get all reviews that aren't replies."""
-        # Use extra because Django wants to do a LEFT OUTER JOIN.
-        return self.extra(where=['reply_to IS NULL'])
+        return self.filter(reply_to__isnull=True)
```
eviljeff: I copied this from the olympia patch that added soft deletion - I don't know if its strictly required.

diox: Keep it, it's cool. Django no longer does the useless JOIN when you do `__isnull=True` queries, we don't need the `.extra()` anymore.

diox:
We're going to need to check db indexes (what we have, what we use, what we could modify to help the new queries we are going to make with the new `deleted` column)

mkt/ratings/forms.py
```diff
@@ -36,10 +36,8 @@ def save(self):
                 review = form.instance
                 addon = review.addon
                 if action == REVIEW_MODERATE_DELETE:
-                    review_addon = review.addon
-                    review_id = review.id
                     review.delete()
-                    mkt.log(mkt.LOG.DELETE_REVIEW, review_addon, review_id,
+                    mkt.log(mkt.LOG.DELETE_REVIEW, addon, review,
```
andymckay: should that be `review.id`?

mkt/ratings/forms.py
```diff
@@ -47,11 +45,11 @@ def save(self):
                                          is_flagged=is_flagged))
                     if self.request:
                         ReviewerScore.award_moderation_points(
-                            self.request.user, addon, review_id)
+                            self.request.user, addon, review.id)
                 elif action == REVIEW_MODERATE_KEEP:
                     review.editorreview = False
                     review.save()
-                    mkt.log(mkt.LOG.APPROVE_REVIEW, review.addon, review,
+                    mkt.log(mkt.LOG.APPROVE_REVIEW, addon, review,
```
andymckay: oh its `review` not `review.id` here as well, so I guess either both are right or wrong

mkt/ratings/models.py
```diff
@@ -59,6 +82,16 @@ class Meta:
...
 
+    def delete(self):
+        self.update(deleted=True)
+        # This should happen in the `post_save` hook.
+        # self.refresh(update_denorm=True)
+
+    def undelete(self):
+        self.update(deleted=False)
+        # This should happen in the `post_save` hook.
+        # self.refresh(update_denorm=True)
```
andymckay: should we just do that then?

mkt/ratings/tests/test_models.py
```diff
@@ -38,3 +38,15 @@ def test_review_aggregates_triggers_reindex(self, index_webapps):
         index_webapps.reset_mock()
         addon_review_aggregates(self.app.pk)
         assert index_webapps.called
+
+    def test_soft_delete(self):
+        eq_(Review.objects.count(), 2)
+        eq_(Review.with_deleted.count(), 2)
```
andymckay: These two lines just show you the fixtures set up correctly, I don't think they contribute to the test.

mkt/ratings/tests/test_models.py
```diff
@@ -38,3 +38,15 @@ def test_review_aggregates_triggers_reindex(self, index_webapps):
...
+
+        Review.objects.all()[0].delete()
+        eq_(Review.objects.count(), 1)
+        eq_(Review.with_deleted.count(), 2)
+
+        Review.objects.all()[0].delete()
+        eq_(Review.objects.count(), 0)
+        eq_(Review.with_deleted.count(), 2)
```
andymckay: I'm not sure the second set of assertions show you much here either. Unless you are expecting something wierd on the second deletion.

mkt/reviewers/forms.py
```diff
@@ -26,8 +25,30 @@
     STATUS_CHOICES.append((mkt.STATUS_CHOICES_API[status],
                            mkt.STATUS_CHOICES[status]))
 
+MODERATE_ACTION_FILTERS = (('', ''), ('approved', _lazy(u'Approved reviews')),
+                           ('deleted', _lazy(u'Deleted reviews')))
 
-log = logging.getLogger('z.reviewers.forms')
+MODERATE_ACTION_DICT = dict(approved=mkt.LOG.APPROVE_REVIEW,
+                            deleted=mkt.LOG.DELETE_REVIEW)
```
andymckay: just doing an actual dict `{approved:mkt.LOG.APPROVE_REVIEW...}` is faster and generally better :tm:

mkt/reviewers/models.py
```diff
@@ -142,9 +142,10 @@ def award_points(cls, user, addon, status, **kwargs):
         return score
 
     @classmethod
-    def award_moderation_points(cls, user, addon, review_id):
+    def award_moderation_points(cls, user, addon, review_id, undo=False):
         """Awards points to user based on moderated review."""
-        event = mkt.REVIEWED_APP_REVIEW
+        event = (mkt.REVIEWED_APP_REVIEW if not undo else
+                 mkt.REVIEWED_APP_REVIEW_UNDO)
```
andymckay: ideally there would be a test for this on the model

mkt/reviewers/forms.py
```diff
@@ -26,8 +25,30 @@
...
+
+
+class ModerateLogForm(happyforms.Form):
```
andymckay: ideally there would be test_forms for this, especially for the `clean` method

mkt/reviewers/templates/reviewers/moderatelog_detail.html
```diff
@@ -0,0 +1,46 @@
...
+          <dd>{{ log.created|babel_datetime }}</dd>
+          <dt>{{ _("App title") }}</dt>
+          <dd>{{ log.details.addon_title }}</dd>
+          <dt>{{ _("Review title") }}</dt>
+          <dd>{{ log.details.title }}</dd>
+          <dt>{{ _("Review text") }}</dt>
+          <dd>{{ log.details.body }}</dd>
+        </dl>
+        {% if log.details.is_flagged %}
+          <p><strong>{{ _("Review was flagged prior to moderation.") }}</strong></p>
```
andymckay: there's a mix of ' and " in this template for translated strings, I think we aim to stick to '?

mkt/reviewers/tests/test_views.py
```diff
@@ -4146,3 +4146,60 @@ def test_comm_url_no_encode(self):
...
+
+    def setUp(self):
+        super(TestModerateLog, self).setUp()
+        self.login_as_editor()
+        self.url = reverse('reviewers.apps.moderatelog')
+        self.review = Review.objects.create(addon=app_factory(), body='body',
+                                            user=user_factory(), rating=4,
+                                            editorreview=True)
+        mkt.set_user(UserProfile.objects.get(username='editor'))
+
```
andymckay: do we need a test for unauthorized?

mkt/reviewers/tests/test_views.py
```diff
@@ -4146,3 +4146,60 @@ def test_comm_url_no_encode(self):
...
+    def test_log(self):
+        r = self.client.get(self.url)
+        eq_(r.status_code, 200)
+
+    def test_start_filter(self):
+        r = self.client.get(self.url, dict(start='2011-01-01'))
+        eq_(r.status_code, 200)
```
andymckay: not sure what this tests too much, that start can be parsed?

mkt/reviewers/tests/test_views.py
```diff
@@ -4146,3 +4146,60 @@ def test_comm_url_no_encode(self):
...
+        """
+        for i in xrange(2):
+            mkt.log(mkt.LOG.APPROVE_REVIEW, self.review, self.review.addon)
+            mkt.log(mkt.LOG.DELETE_REVIEW, self.review, self.review.addon)
+        r = self.client.get(self.url, dict(search='deleted'))
+        eq_(pq(r.content)('tbody tr').length, 2)
+
+    def test_no_results(self):
+        r = self.client.get(self.url, dict(end='2004-01-01'))
+        assert '"no-results"' in r.content, 'Expected no results to be found.'
```
andymckay: those are useful tests

mkt/reviewers/urls.py
```diff
@@ -36,7 +36,13 @@
         name='reviewers.apps.review.attachment'),
     url(r'^apps/review/%s/abuse$' % mkt.APP_SLUG, views.app_abuse,
         name='reviewers.apps.review.abuse'),
+
     url(r'^apps/logs$', views.logs, name='reviewers.apps.logs'),
+    url(r'^apps/moderatelogs$', views.moderatelog,
```
andymckay: maybe `apps/logs/moderate`, having `logs` and `moderatelogs` feels ungainly

mkt/reviewers/views.py
```diff
@@ -1213,3 +1213,44 @@ class ReviewerScoreViewSet(CORSMixin, MarketplaceView, viewsets.ModelViewSet):
...
+    review = None
+
+    if len(log.arguments) > 1 and isinstance(log.arguments[1], Review):
+        review = log.arguments[1]
+    is_admin = acl.action_allowed(request, 'ReviewerAdminTools', 'View')
+    can_undelete = review and review.deleted and (
+                    is_admin or request.user.pk == log.user.pk)
+
+    if request.method == 'POST':
+        # A Form seems overkill for this.
```
andymckay: forms are never overkill :)

mkt/site/models.py
```diff
@@ -322,7 +322,9 @@ def update(self, **kw):
                 if attrs[k] != v:
                     kw[k] = v
                     setattr(self, k, v)
-        cls.objects.filter(pk=self.pk).update(**kw)
+        # We want this to not fail mysteriously for soft deleted objects.
+        objects = getattr(cls, 'with_deleted', cls.objects)
+        objects.filter(pk=self.pk).update(**kw)
```
andymckay: test?

mkt/reviewers/views.py
```diff
@@ -1213,3 +1213,44 @@ class ReviewerScoreViewSet(CORSMixin, MarketplaceView, viewsets.ModelViewSet):
...
+        if request.POST['action'] == 'undelete':
+            if not can_undelete:
+                raise PermissionDenied
+            ReviewerScore.award_moderation_points(
+                log.user, review.addon, review.id, undo=True)
+            review.undelete()
+            return redirect('reviewers.apps.moderatelog.detail', id)
+    data = context(request, log=log, review=review, can_undelete=can_undelete)
```
andymckay: It feels like there's about 5 code paths in here and I only see one test. Another reason forms are awesome, they let you test logic without testing the whole view code.

mkt/ratings/forms.py
eviljeff: I thought so, but the logging code didn't output a useful link without the object itself (not an issue before, when the object was already deleted).

eviljeff:
In defence of the lack of tests and odd comments, I copied all this from olympia and didn't do much more than the minimum to make it work :stuck_out_tongue: But, sure I'll add some more.

eviljeff:
@diox I wasn't sure what changes you wanted with the "We're going to need to check db indexes" comment.

mkt/ratings/models.py
eviljeff: The comment means the following line occurs in post_save.  (I guessed and then confirmed with Kris).  It seems redundant to me so I can just remove the comments.

diox:
Not sure yet - I can't check from here. I'm worried about the potential negative impact on the consumer side. Proper db indexes will solve this.

We need to:
- Look at existing indexes for this table
- Look at how queries we are making currently perform with those indexes (through `EXPLAIN ...`)
- Look at if we are going to need to modify an existing index (to add the `deleted` column in an existing composite index) or add another one

ops should be able to help.

mkt/reviewers/urls.py
eviljeff: @andymckay 'apps/reviewlogs' and 'apps/moderatelogs'?  I thought having /moderate under /logs would imply it was a log of moderation of reviewer reviews, rather than moderation of user ratings.  ( /apps/logs/review and /apps/logs/moderate is a possibility too but would have to redirect /logs to something.)

eviljeff:
@diox the with_deleted object for Review is referenced minimally fwiw (https://github.com/mozilla/olympia/pull/374/files#diff-971684285f553add130befa48e053329R391 was the only place added, outside of tests.  ActivityLog uses it too but that's only exposed to app reviewers, via this patch).  If you think someone in ops would understand your comment and be able to provide the answers I'll ping someone in #remora though.

diox:
The default manager uses `qs.exclude(deleted=True).exclude(reply_to__deleted=True)` so that completely changes the query used for the consumer pages.

mkt/reviewers/tests/test_views.py
eviljeff: Yeah.  There are tests for start being provided; end being provided; and the filter.  Maybe a bit simplistic I suppose.

eviljeff:
logged https://bugzilla.mozilla.org/show_bug.cgi?id=1124222 for optimization

eviljeff:
I believe I've addressed the comments and there are now more tests.
After some investigation with @diox we've decided not to add any further indexes now and defer instead to bug 1124222
r? @andymckay or @diox

mkt/developers/models.py
```diff
@@ -494,7 +494,7 @@ def to_string(self, type_=None):
 
         try:
             kw = dict(addon=addon, review=review, version=version, group=group,
-                      collection=collection, tag=tag, user=self.user.username)
+                      collection=collection, tag=tag, user=self.user.display_name)
```
diox: Is there an XSS test somewhere that would cover this change ?

eviljeff: Everything is returned via self.f on the next line, and the function is actually only used for the log viewer introduced in this patch (its leftover from AMO).

diox: Yes, my point was, is there an test checking that there is no possibility of XSS through the user display name in the log viewer.

eviljeff: Added.

mkt/reviewers/views.py
```diff
@@ -1213,3 +1214,50 @@ class ReviewerScoreViewSet(CORSMixin, MarketplaceView, viewsets.ModelViewSet):
...
+        if form.cleaned_data['search']:
+            modlog = modlog.filter(action=form.cleaned_data['search'].id)
+
+    pager = paginate(request, modlog, 50)
+    data = context(request, form=form, pager=pager, tab='moderatelog')
+    return render(request, 'reviewers/moderatelog.html', data)
+
+
+@reviewer_required
+def moderatelog_detail(request, id):
```
diox: nit: `id` is a built-in function, so I like to avoid using it as a variable/argument name when possible

mkt/reviewers/tests/test_views.py
```diff
@@ -4146,3 +4139,114 @@ def test_comm_url_no_encode(self):
...
+        super(TestModerateLogDetail, self).setUp()
+        # AccessMixin needs a url property.
+        self.url = self._url(0)
+
+    def _url(self, id):
+        return reverse('reviewers.apps.moderatelog.detail', args=[id])
+
+    def test_detail_page(self):
+        mkt.log(mkt.LOG.APPROVE_REVIEW, self.review.addon, self.review)
+        id = ActivityLog.objects.editor_events()[0].id
```
diox: nit: id is a built-in function, so I like to avoid using it as a variable/argument name when possible (also happens below in the view, and in other tests)

mkt/reviewers/views.py
```diff
@@ -1213,3 +1214,50 @@ class ReviewerScoreViewSet(CORSMixin, MarketplaceView, viewsets.ModelViewSet):
...
+    form = ModerateLogDetailForm(request.POST or None)
+    is_admin = acl.action_allowed(request, 'ReviewerAdminTools', 'View')
+    can_undelete = review and review.deleted and (
+                    is_admin or request.user.pk == log.user.pk)
+
+    if (request.method == 'POST' and form.is_valid() and
+        form.cleaned_data['action'] == 'undelete'):
+        if not can_undelete:
+            if not review:
+                raise RuntimeError('Review doesn`t exist.')
```
diox: How about a 404 instead here and just below ? raising a `RuntimeError` would result in a 500, which wouldn't be nice to reviewers. Also, these 2 cases need tests.

diox:
Sorry this review took so long, I had forgotten about it. I left a couple comments but it looks good. **r+wc**, preferably after the tag because we are a bit late for QA.

mkt/reviewers/views.py
eviljeff: They're catching cases that shouldn't exist - a POST submission attempting to undelete something that isn't deleted or doesn't exist.  If zamboni is rendering these forms its a bug somewhere else so we'd want it in sentry I figured.
I could just take out these lines and award_moderation_points() or review.undelete() would fail instead, at the expense of a slightly longer stacktrace?

PR #2841, opened 2015-01-14
Configure fxpay in Fireplace views (bug 1113801)

kumar303:
Fireplace side: https://github.com/mozilla/fireplace/pull/907

### DIFF
--- a/mkt/commonplace/views.py
+++ b/mkt/commonplace/views.py
@@ -116,11 +116,11 @@ def commonplace(request, repo, **kwargs):
     # need it.
     fxa_auth_state, fxa_auth_url = fxa_auth_info()
     site_settings = {
+        'dev_pay_providers': settings.DEV_PAY_PROVIDERS,
         'fxa_auth_state': fxa_auth_state,
-        'fxa_auth_url': fxa_auth_url
+        'fxa_auth_url': fxa_auth_url,
+        'fxa_css_path': settings.FXA_CSS_PATH,
     }
-
-    site_settings['fxa_css_path'] = settings.FXA_CSS_PATH
 
     ctx = {
         'BUILD_ID': BUILD_ID,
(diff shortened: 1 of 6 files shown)


mkt/settings.py
```diff
@@ -1213,3 +1213,19 @@ def JINJA_CONFIG():
...
+DEV_PAY_PROVIDERS = {
+    'mozilla-dev/payments/pay/v1':
+        'https://marketplace-dev.allizom.org/mozpay/?req={jwt}',
+    'mozilla-stage/payments/pay/v1':
+        'https://marketplace.allizom.org/mozpay/?req={jwt}',
+    'mozilla-alt/payments/pay/v1':
+        'https://payments-alt.allizom.org/mozpay/?req={jwt}',
+    'mozilla-local/payments/pay/v1':
+        'http://mp.dev/mozpay/?req={jwt}',
+}
```
andymckay: would we want that per server? e.g: stage has mozilla-stage, but not the others?

kumar303: are you asking if we should only allow a specific server in each site setting? Like, only add stage to the allowed list in zamboni settings for stage?

andymckay: Yup, being able to access payments-alt payments on -dev seems unneccessary.

Would probably allow you to remove the boolean too.

andymckay:
r+

mkt/commonplace/views.py
```diff
@@ -116,12 +116,12 @@ def commonplace(request, repo, **kwargs):
     # need it.
     fxa_auth_state, fxa_auth_url = fxa_auth_info()
     site_settings = {
+        'dev_pay_providers': settings.DEV_PAY_PROVIDERS,
```
ngokevin: We only need this for the site, right? Because if we want it for package, we'll want it in the site_config endpoint also.

kumar303: anything that runs on Firefox OS won't need this because all that is handled by device settings. It would be nice if we could change device settings but you have to do that at build time (settings similar to these are baked into engineering builds).

PR #2844, opened 2015-01-14
Remove less2stylus() and jquery_migrated(), they are not used any more (bug 1058976)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1058976


ngokevin:
r+

PR #2846, opened 2015-01-15
Update loop dev URL (bug 1121580)

diox:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1121580#c10 loop dev URL is now `https://loop-webapp-dev.stage.mozaws.net/`.

### DIFF
--- a/mkt/commonplace/views.py
+++ b/mkt/commonplace/views.py
@@ -61,7 +61,7 @@ def get_allowed_origins(request, include_loop=True):
         # On dev, include loop dev origin as well.
         if development_server:
             allowed.extend([
-                'http://loop-webapp.dev.mozaws.net',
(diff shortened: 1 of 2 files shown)


mkt/commonplace/views.py
```diff
@@ -61,7 +61,7 @@ def get_allowed_origins(request, include_loop=True):
         # On dev, include loop dev origin as well.
         if development_server:
             allowed.extend([
-                'http://loop-webapp.dev.mozaws.net',
+                'https://loop-webapp-dev.stage.mozaws.net/',
```
eviljeff: Is the '/' an issue?  With origins they normally need to be without the ending /

diox: Good call, it's my mistake.

eviljeff:
r+

PR #2850, opened 2015-01-16
remove legacy from countries and make consistent on 3 letter ISO codes

andymckay:


robhudson:
What are we doing for the API which take the 2-letter codes? e.g. `?region=us`?

andymckay:
I wasn't planning on make such a breaking change and just keeping this as an internal zamboni cleanup.

robhudson:
First pass looked good. r+

PR #2851, opened 2015-01-16
remove DecimalField

andymckay:
not sure what crappy data this field we were trying to store was, but the only source now is solitude and that's pretty good


robhudson:
r+ with passing tests.

PR #2852, opened 2015-01-17
replace assertRedirects with assert3xx (bug 1025277)

andymckay:


robhudson:
r+. When Django 1.7 comes out we can rewrite `assert3xx` to just call `assertRedirects(..., fetch_redirect_response=False)`.

andymckay:
Thats good to know!

PR #2855, opened 2015-01-19
remove mcc from the API and zamboni (bug 1058257)

andymckay:
* fireplace maintains its own list of mccs https://github.com/mozilla/fireplace/blob/66a132ecf3e72fafde7c3e096ae0d37ff475d2c2/src/media/js/mobilenetwork.js#L32
* the mcc can be more than one per region, so what zamboni has is incorrect
* this is actually a *backwards incompatible change* for the API, should we do anything different?

### DIFF
--- a/docs/api/topics/apps.rst
+++ b/docs/api/topics/apps.rst
@@ -115,13 +115,11 @@ App
             "regions": [
                 {
                     "adolescent": true,
-                    "mcc": 310,
                     "name": "United States",
                     "slug": "us"
                 },
                 {
                     "adolescent": true,
-                    "mcc": null,
                     "name": "Rest of World",
                     "slug": "restofworld"
                 }
@@ -272,9 +270,6 @@ App
         volume of data to calculate ratings and rankings independent of
         worldwide data.
     :type regions.adolescent: boolean
-    :param regions.mcc: represents the region's ITU `mobile
-        country code`_.
-    :type regions.mcc: string|null
     :param regions.name: The region name.
     :type regions.name: string
     :param regions.slug: The region slug.
--- a/docs/api/topics/apps.rst
+++ b/docs/api/topics/apps.rst
@@ -326,6 +326,15 @@ App
          15   Blocked
     =======  ============================
 
+    .. warning:: Version differences.
+
+    The field `regions.mcc` was removed in v2. In v1 it was:
+
+    :param regions.mcc: represents the region's ITU `mobile
+        country code`_.
+    :type regions.mcc: int|null
+
+
 .. http:get:: /api/v2/apps/(int:id)|(string:slug)/privacy/
 
     **Response**
@@ -413,7 +422,8 @@ Versions
         public version of the application.
     :type is_current_version: boolean
     :param features: each item represents a
-        :ref:`device feature <features>` required to run the application.
+        :ref:`feature profile <feature-profile-label>` required to run the
+        application.
     :type features: array
     :param release_notes: the release notes for that version.
     :type release_notes: string|object|null
@@ -440,8 +450,8 @@ Versions
         }
 
     :param features: each item represents a
-        :ref:`device feature <features>` required to run the application.
-        Features not present are assumed not to be required.
+        :ref:`feature profile <feature-profile-label>` required to run the
+        application. Features not present are assumed not to be required.
     :type features: array
 
     **Response**
--- a/mkt/api/base.py
+++ b/mkt/api/base.py
@@ -159,8 +159,8 @@ class MarketplaceView(object):
             except ValueError:
                 pass  # Swallow and ignore invalid input.
 
-        return super(MarketplaceView, self).paginate_queryset(queryset,
-            page_size=page_size)
+        return super(MarketplaceView, self).paginate_queryset(
+            queryset, page_size=page_size)
 
     def get_region_from_request(self, request):
         """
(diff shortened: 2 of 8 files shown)


diox:
Well, technically, we're supposed to keep v1 of the API working as it was when it was frozen... But if what it is exposing is incorrect as you point out, I'm not sure...

andymckay:
Added in different serializers per request, as stolen from @mstriemer

clouserw:
Do any shipping versions of FxOS use this?  Old versions of the marketplace that never updated?  If there are, can we tell what will happen to them if this goes away?  Hopefully fail gracefully.  Lastly, before this change goes out we should email marketplace-api-announce@

mstriemer:
There are some test failures but many of them will likely be solved by #2861. There's at least one to do with MCC.

docs/api/topics/apps.rst
```diff
@@ -115,13 +115,11 @@ App
             "regions": [
                 {
                     "adolescent": true,
-                    "mcc": 310,
```
mstriemer: I guess we don't version our API docs? Would be nice to do that somehow. Maybe add something like `Changed in V2: "mcc" was removed from the output.` below this?

mkt/api/base.py
```diff
@@ -172,6 +172,16 @@ def get_region_from_request(self, request):
         """
         return get_region_from_request(request)
 
+    def get_serializer_class(self):
+        """
+        Return a serializer for the requested API version. Versioned
+        serializers are registered on the default serializer with their version
+        number.
```
mstriemer: This should probably have an example.

```py
class MySerializer(Serializer):
    some_field = CharField()
class MySerializerV1(MySerializer):
    # some_field is inherited.
    deprecated_field = CharField()
MySerializer.V1 = MySerializerV1
```

mkt/api/base.py
```diff
@@ -172,6 +172,16 @@ def get_region_from_request(self, request):
...
+        """
+        api_version = 'V{v_int}'.format(v_int=self.request.API_VERSION)
```
diox: We use lowercase `v{int}` elsewhere (namespace, url name, prefix) - any reason to use uppercase V here ?

mkt/api/tests/test_handlers.py
```diff
@@ -678,7 +678,18 @@ def test_banner_message(self):
...
+        eq_(data['banner_regions'],
+            [mkt.regions.ARG.slug, mkt.regions.BRA.slug])
+
+    def test_region_v1(self):
+        res = self.client.get(self.get_url)
+        eq_(res.status_code, 200)
+        eq_(res.json['regions'][0]['mcc'], '722')
+
+    def test_region_v2(self):
+        res = self.client.get(self.get_url.replace('v1', 'v2'))
```
mstriemer: This seems fragile. The v1 test too. Not sure what to do about it though. Could add an assert that the right version is in the URL I guess.

diox: Yeah, instead, you should pass a different namespace when reversing `'app-detail'`, IMHO.

mkt/constants/regions.py
```diff
@@ -108,12 +108,22 @@ class RESTOFWORLD(REGION):
     if country.get('ratingsbody'):
         country['ratingsbody'] = getattr(ratingsbodies, country['ratingsbody'])
 
+    # This if check can be removed once marketplace-constants is updated.
```
mstriemer: Is marketplace-constants changing `mcc` to `mccs`? I'm not sure what has to change to remove this.

mkt/api/base.py
mstriemer: I used an uppercase V because it's a class.

mkt/webapps/tests/test_serializers.py
```diff
@@ -321,6 +322,15 @@ def test_missing_price(self):
         eq_(res['price_locale'], None)
 
 
+class TestRegionSerializer(mkt.site.tests.TestCase):
+
+    def test_v1(self):
+        eq_(RegionSerializerV1(regions.ARG).data['mcc'], '722')
```
mstriemer: I think this test is failing with `722 != '722'`.

mstriemer:
lgtm. I'll leave the final r? someone else though since I'm probably biased.

mkt/api/base.py
diox: Technically it's just an attribute :) It's not very different from, say, `_meta` attribute on models.

mstriemer: Fair enough, lowercase works for me.

andymckay: Only reason was it looks nicer in a camel case class name `AppSerializerV1` is more obvious than `AppSerializerv1` and then wanted to be consistent.

mkt/api/base.py
andymckay: There's tests for this, seemed good enough.

mkt/constants/regions.py
andymckay: Its returning mcc as a list in the next version. So once 0.5.0 is out, it will always be a list and we remove this.

mkt/webapps/tests/test_serializers.py
andymckay: Ah thanks.

mkt/api/base.py
diox: `AppSerializerV1` is fine. It's just `AppSerializer.V1` (with the dot) that feels weird to me.

mstriemer:
If `mcc` was a string before it should probably stay a string. That would still be a backwards incompatible change.

andymckay:
Its not though, the docs are wrong. Scroll up the docs to see that the example in the API contradicts the docs:

'''
[1058257] zamboni $ curl -a https://marketplace.firefox.com/api/v1/apps/app/twitter/ | python -m json.tool | a mcc
            "mcc": 722,
            "mcc": 470,
...
'''

andymckay:
This is not worth it for a field no one uses.

PR #2857, opened 2015-01-20
Disable cache in travis builds, it's actually slower

diox:
(Stolen from mozilla/olympia#427)


diox:
Results are not as dramatic as in olympia. In fact, it varies from 2 minutes faster to 2 minutes slower. Closing this for now.

PR #2858, opened 2015-01-20
Fix requirements for dev.txt

robhudson:
pip was throwing an error about duplicate packages b/c we were including
prod and compiled twice when using dev.txt


andymckay:
r+, thanks for fixing my bugs!

PR #2860, opened 2015-01-21
run through supervisord (bug 1122190)

andymckay:


Dockerfile
```diff
@@ -15,7 +15,8 @@ RUN yum install -y redis \
     gcc-c++ \
     npm \
     wget \
-    totem
+    totem \
+    supervisor
```
muffinresearch: The recommended way to install supervisord is via pypi according to their docs. You can do that directly in this dockerfile though to save on putting the dep in requirements. If the version via yum is recent enough maybe that won't matter?

muffinresearch:
As noted in the related PR a bin dir seems odd for a conf file. 

Otherwise r+wc.

Dockerfile
andymckay: Its 3.0b3 as opposed to 3.1.3, but it seems to work just fine. I'm not sure its a big deal but it means that in my head - yum is all the libraries for running zamboni and pip are all the python deps the code needs.

PR #2861, opened 2015-01-21
Turn DEBUG back to False in tests

diox:
(This was changed in 2e4183c6 but it looks like an accident)


mstriemer:
Makes the tests pass so I like it. Should probably get confirmation from @robhudson though.

robhudson:
I think I changed it to fix something about tests related to templates failing but if this now works (which it looks like it does), r+.

PR #2863, opened 2015-01-22
Barebone model for LangPack (bug 1122255)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1122255

- No Version/File model, keeping it simple - I don't think we need to deal with old versions so I don't see why we would have a model for them

- No soft-deletion

- Re-uses from_upload() static method and FileUpload in order to be compatible with existing submit API

Only accepts a single language and a single Firefox OS version. Might change that depending on the discussions regarding those topics.


migrations/882-add-langpacks.sql
```diff
@@ -0,0 +1,14 @@
...
+    `language` varchar(10) NOT NULL,
+    `fxos_version` varchar(255) NOT NULL,
+    `version` varchar(255) NOT NULL,
+    `filename` varchar(255) NOT NULL,
+    `hash` varchar(255) NOT NULL,
+    `size` integer UNSIGNED NOT NULL,
+    `active` bool NOT NULL
+)
+;
+CREATE INDEX `langpacks_langpack_914f9139` ON `langpacks_langpack` (`fxos_version`, `language`, `active`);
```
robhudson: I typically try to rename these to something more reasonable.

mkt/langpacks/models.py
```diff
@@ -0,0 +1,108 @@
...
+
+class LangPack(ModelBase):
+    # Primary key is a uuid in order to be able to set it in advance (we need
+    # something unique for the filename, and we don't have a slug).
+    uuid = UUIDField(primary_key=True, auto=True)
+
+    # Fields for which the manifest is the source of truth - can't be
+    # overridden by the API.
+    language = models.CharField(choices=LANGUAGE_CHOICES,
+                                default=settings.LANGUAGE_CODE, max_length=10)
```
robhudson: if someone tries to submit a language that isn't in our choices, they get a validation error?

mkt/langpacks/tests/test_models.py
```diff
@@ -0,0 +1,69 @@
...
+        eq_(langpack.language, 'de')
+        eq_(langpack.fxos_version, '2.2')
+        eq_(langpack.filename, '%s-%s.zip' % (langpack.uuid, langpack.version))
+        ok_(langpack.filename in langpack.file_path)
+        ok_(os.path.exists(langpack.file_path))
+        eq_(langpack.hash[0:23], 'sha256:f0fa5a4f5c0edf2d')
+        eq_(langpack.size, 499)
+        ok_(LangPack.objects.no_cache().get(pk=langpack.uuid))
+        eq_(LangPack.objects.count(), 1)
+        return langpack
```
robhudson: you need this `return` here?

mkt/langpacks/tests/test_models.py
```diff
@@ -0,0 +1,69 @@
...
+        eq_(langpack.language, 'de')
+        eq_(langpack.fxos_version, '2.2')
+        eq_(langpack.filename, '%s-%s.zip' % (langpack.uuid, langpack.version))
+        ok_(langpack.filename in langpack.file_path)
+        ok_(os.path.exists(langpack.file_path))
+        eq_(langpack.hash[0:23], 'sha256:f0fa5a4f5c0edf2d')
+        eq_(langpack.size, 499)
+        ok_(LangPack.objects.no_cache().get(pk=langpack.uuid))
+        eq_(LangPack.objects.count(), 1)
+
```
robhudson: trailing newline

mkt/langpacks/tests/test_models.py
```diff
@@ -0,0 +1,69 @@
...
+        ok_(langpack.filename in langpack.file_path)
+        ok_(os.path.exists(langpack.file_path))
+        eq_(langpack.hash[0:23], 'sha256:f0fa5a4f5c0edf2d')
+        eq_(langpack.size, 499)
+        ok_(LangPack.objects.no_cache().get(pk=langpack.uuid))
+        eq_(LangPack.objects.count(), 1)
+        return langpack
+
+    def test_upload_existing(self):
+        langpack = self.test_upload_new()
```
robhudson: ah, you use the return here, nvm.

mkt/langpacks/utils.py
```diff
@@ -0,0 +1,43 @@
...
+        # At the moment, we don't care about most of the manifest for
+        # langpacks so we simply extract the minimum we require ; in the future
+        # we might want to call the parent's parse() method and have a way to
+        # deal with extra stuff in the child.
+        data = self.get_json_data(upload)
+
+        languages_provided = data.get('languages-provided', {}).keys()
+        if len(languages_provided) < 1:
+            raise forms.ValidationError(_(u'Your language pack must contain '
+                    'one language in `languages-provided` object'))
```
robhudson: funky wrapping here. Not sure that's pep8.

mkt/langpacks/utils.py
```diff
@@ -0,0 +1,43 @@
...
+        if len(languages_provided) > 1:
+            # For now, refuse langpacks containing more than one language.
```
robhudson: :+1:

mkt/langpacks/utils.py
```diff
@@ -0,0 +1,43 @@
...
+                    'many languages. Only one language per pack is supported'))
+
+        languages_target = data.get('languages-target', {}).values()
+        if len(languages_target) < 1:
+            raise forms.ValidationError(_(u'Your language pack must contain '
+                    'one language in `languages-target` object'))
+        if len(languages_target) > 1:
+            # For now, refuse langpacks containing more than one target.
+            raise forms.ValidationError(_(u'Your language pack contains too '
+                    'many targets. Only one target per pack is supported'))
```
robhudson: So we're ok with the wildcards in the target keys?

mkt/langpacks/utils.py
```diff
@@ -0,0 +1,43 @@
...
+    def parse(self, upload):
+        """Parse the FileUpload passed in argument and return langpack data.
+        May raise forms.ValidationError()"""
+        # At the moment, we don't care about most of the manifest for
+        # langpacks so we simply extract the minimum we require ; in the future
+        # we might want to call the parent's parse() method and have a way to
+        # deal with extra stuff in the child.
+        data = self.get_json_data(upload)
+
+        languages_provided = data.get('languages-provided', {}).keys()
```
robhudson: What about multiple apps in the languages provided? is that fine?

mkt/langpacks/utils.py
```diff
@@ -0,0 +1,43 @@
...
+# FIXME: needs test, by mocking get_json_data() in various situations.
+class LanguagePackParser(WebAppParser):
+    def parse(self, upload):
+        """Parse the FileUpload passed in argument and return langpack data.
+        May raise forms.ValidationError()"""
+        # At the moment, we don't care about most of the manifest for
+        # langpacks so we simply extract the minimum we require ; in the future
+        # we might want to call the parent's parse() method and have a way to
+        # deal with extra stuff in the child.
+        data = self.get_json_data(upload)
```
robhudson: We should throw a validation error `if role != 'langpack'`.

mkt/langpacks/utils.py
```diff
@@ -0,0 +1,43 @@
+ # -*- coding: utf-8 -*-
+from django import forms
+
+from tower import ugettext as _
+
+from mkt.files.utils import WebAppParser
+
+
+# FIXME: needs test, by mocking get_json_data() in various situations.
```
robhudson: I agree. :)

robhudson:
So far so good. Great start.

mkt/langpacks/utils.py
diox: Yep, good call. And forbid `role='langpack'` apps in `WebAppParser.parse()`.

mkt/langpacks/utils.py
diox: I think I'm going to add a check that the target is exactly `"app://*.gaiamobile.org/manifest.webapp"` and refuse anything else. And not care about what's in `"languages-provided"` beside the language itself.

mkt/langpacks/utils.py
diox: See below

mkt/langpacks/models.py
diox: Right now they don't, because we bypass the choices, but yeah, that's the plan. Going to add it.

mkt/langpacks/utils.py
diox: I'm open to suggestions, I'm not sure what would be best here

robhudson: I'd try starting the `_(u'...` on a new line with 4-space indent. That gives you a lot of room for the string and wrapping.

robhudson:
r+ from me. I saw some "FIXME"s so if there's more I'm happy to re-review but this is great so far.

mkt/langpacks/utils.py
diox: I've added the role check below, but still need to add the opposite one in `WebAppParser` and write the tests.

mkt/langpacks/models.py
diox: Added.

mkt/langpacks/utils.py
diox: Added tests

mkt/langpacks/utils.py
diox: Done.

diox:
@robhudson please take a second look - there are 2 FIXME left, but I don't know what to do about those yet. Everything else should be working and tested.

mkt/langpacks/tests/test_utils_.py
```diff
@@ -0,0 +1,67 @@
...
+            'languages-provided': {
+                'es': {},
+                'de': {},
+            },
+            'languages-target': {
+                'app://*.gaiamobile.org/manifest.webapp': '2.2'
+            }
+        }
+        expected = [u'Your language pack contains too many languages. '
+                    u'Only one language per pack is supported.']
```
robhudson: I wonder if these error messages could be a variable that is pulled from the same place so we don't have to keep them in sync if they change?

diox: I debated doing that, but existing `WebAppParser` tests hardcoded them too, so I followed that logic. That'll force us to be careful and run the tests before making changes :)

PR #2865, opened 2015-01-23
Remove FxA migration code (bug 1117875)

mstriemer:


andymckay:
I hope never to see the HTML in those emails again.

Thanks for doing this.

r+ if tests pass

PR #2866, opened 2015-01-23
Fix popup closure on developer pages (bug 1124781)

muffinresearch:


mstriemer:
Makes sense, what spurred on the change from the window closing itself to the app closing the window?

mstriemer:
r+

kumar303:
@mstriemer we need to share the popup across fxa and payments now -- the client needs to be in control the opening and closing

PR #2869, opened 2015-01-26
Close popup only if it exists

muffinresearch:
Just to be a bit more defensive we should close the popup only if it exists.


eviljeff:
r+

PR #2870, opened 2015-01-26
add in receipt verifier to supervisord (bug 1109334)

andymckay:


kumar303:
looks like a legit test failure

mkt/settings.py
```diff
@@ -381,7 +381,7 @@
 # The verification URL, the addon id will be appended to this. This will
 # have to be altered to the right domain for each server, eg:
 # https://receiptcheck.addons.mozilla.org/verify/
-WEBAPPS_RECEIPT_URL = '/verify/'
+WEBAPPS_RECEIPT_URL = 'http://mp.dev/receipt-verifier/'
```
kumar303: Can't you use [SITE_URL](https://github.com/andymckay/zamboni/blob/1109334/mkt/settings.py#L367) since it is overridable by env var and should default to ``mp.dev``?

kumar303: Oh, hmm, shouldn't this default be changed from ``localhost`` to ``mp.dev``? https://github.com/andymckay/zamboni/blob/1109334/mkt/settings.py#L104 I think Docker sets it anyway via the environment.

scripts/docker/supervisor.conf
```diff
@@ -10,3 +10,12 @@ redirect_stderr=true
 stdout_logfile=logs/docker.log
 stdout_logfile_maxbytes=1MB
 stopsignal=KILL
+
+[program:receipt-verifier]
+command=gunicorn --log-level=DEBUG -c wsgi/receiptverify.py -b 0.0.0.0:9000 --debug verify:application
```
kumar303: For other apps in Docker we've been avoiding the use of default ports so how about making this something non-default (9001 or whatever) ? I believe 9000 is a default used by Pyramid which probably doesn't matter much but there could be other frameworks that use it.

kumar303:
r+wc. Thanks for adding this!

scripts/docker/supervisor.conf
andymckay: Pyramid defaults to 8080, don't know of anything that goes to 9000 by default.

mkt/settings.py
andymckay: You are right, static_url prepends SITE_URL on to it. Although its set to a full URL on prod, we should just make it relative and let SITE_URL fix it up.

PR #2873, opened 2015-01-27
Add scrollbars=yes and increase popup height (bug 1125964)

muffinresearch:
The devhub login popup creation didn't allow scrollbars and the height didn't match the fireplace one. This updates the devhub login popup to allow scrollbars and to match the height used in fireplace et al [1] 

[1] https://github.com/mozilla/marketplace-core-modules/blob/master/utils.js#L188


eviljeff:
r+

PR #2874, opened 2015-01-27
Adjusted popularity measure (bug 930902, 1116564)

robhudson:
This changes how we calculate popularity. It stop using the `Installed` table, which only records installs from logged in users, and moves to using Monolith which tracks installs from all users. It also does a rolling 90 day popularity so newer apps that become popular can also come into play.

I'm removing `total_downloads` and `weekly_downloads` here as well as they weren't being used any longer.


migrations/886-rm-weekly-downloads.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE addons DROP COLUMN weeklydownloads;
```
washort: +1 less stuff in `addons`

migrations/887-rm-total-downloads.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE addons DROP COLUMN totaldownloads;
```
washort: i guess that's +2 then

mkt/lookup/views.py
```diff
@@ -447,9 +447,10 @@ def _app_downloads(app):
              'alltime': 0}
 
     # Webapps populate these fields via Monolith.
+    # TODO: Update this to use monolith or link to stats or use 90d numbers?
     stats['last_24_hours'] = 'N/A'
-    stats['last_7_days'] = app.weekly_downloads
-    stats['alltime'] = app.total_downloads
+    stats['last_7_days'] = 'N/A'
```
eviljeff: Um... TODO to implement something this patch removed?  What's the blocker?

robhudson: @eviljeff I guess just knowing what would be good info there? Also, I removed the totals from the reviewer tools. If you'd like to have the 90-day popularity count I can replace it with that.

washort: this is the only part that confuses me.

washort:
r+

mkt/lookup/templates/lookup/app_summary.html
```diff
@@ -108,12 +108,6 @@
             {% endfor %}
           </ul>
         </dd>
-        <dt>{{ _('Downloads') }}</dt>
-        <dd class="pretty-list">
-          {% for key, heading in summaries %}
-            <div>{{ heading }}: {{ downloads[key] }}</div>
-          {% endfor %}
-        </dd>
```
robhudson: @eviljeff Updated to just remove this section in the lookup tool. We have a link to statistics on this page already and that will be more accurate than this ever was.

robhudson:
Merged: https://github.com/mozilla/zamboni/commit/7388ab8d90839d6c3483eb15a783e05903e0b707

PR #2878, opened 2015-01-28
Re-organize trending and popularity tasks

robhudson:
This combines the calculating of values and reindexing on the same
chunks of apps. We also change the final task in the chain to be a purge
task that removes records that were not updated and also reindex those
so the values in ES are synced for purged records.


robhudson:
Closing in favor of the approach in https://github.com/mozilla/zamboni/pull/2880

PR #2879, opened 2015-01-28
Sign language packs when they are uploaded (bug 1122277)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1122277


mkt/langpacks/models.py
```diff
@@ -81,21 +84,55 @@ def handle_file_operations(self, upload):
         """Handle file operations on an instance by using the FileUpload object
         passed to set filename, size, hash on the LangPack instance, and moving
         the temporary file to its final destination."""
-        # FIXME: store validation result ?
         upload.path = smart_path(nfd_str(upload.path))
         if not self.uuid:
             self.reset_uuid()
         self.filename = self.generate_filename()
```
robhudson: Does `generate_filename` need to be a model method?

mkt/langpacks/serializers.py
```diff
@@ -36,7 +39,11 @@ def restore_object(self, attrs, instance=None):
         try:
             return LangPack.from_upload(attrs['upload'], instance=instance)
         except serializers.ValidationError, e:
-            raise exceptions.ParseError(e.messages)
+            raise ParseError(e.messages)
+        except SigningError, e:
+            # A SigningError could mean the package is corrupted but it often
+            # means something is wrong on our end, so return a 503.
+            raise ServiceUnavailable([e.message])
```
robhudson: When does `restore_object` get called via DRF?

mkt/langpacks/models.py
diox: Like with `genereate_hash()`, we could pass what it needs (`uuid`, `version`) as arguments, but I don't see a strong need to do so. Do you ?

mkt/langpacks/tests/test_models.py
```diff
@@ -98,6 +108,88 @@ def test_upload_language_validation(self, get_json_data_mock):
...
+        )
+        sign_app_mock.assert_called_once_with(*expected_args)
+
+    @patch('mkt.langpacks.models.sign_app')
+    def test_upload_sign_error(self, sign_app_mock):
+        sign_app_mock.side_effect = SigningError
+        eq_(LangPack.objects.count(), 0)
+        upload = self.upload('langpack')
+        with self.assertRaises(SigningError):
+            LangPack.from_upload(upload)
```
robhudson: How about a test for a pre-existing langpack, and someone uploads a new one, but there's a signing error. To verify that the old one is still served, etc?

mkt/langpacks/serializers.py
diox: At the end of the deserialization process. It's the method that creates (or updates) the model instance, which is why it's the one we are overriding here, since we want a completely custom way of creating our model instances.

mkt/langpacks/models.py
robhudson: I couldn't find anything else calling it and it's only a one-liner so it seems like it could just be inlined here. But if there's future plans or other things calling it I didn't find, no biggie.

mkt/langpacks/tests/test_models.py
diox: Good idea, I added a test.

mkt/langpacks/models.py
diox: Ah, I didn't want to inline it to keep the API similar to `File` and keep the code readable. Now I actually use it in the tests to generate a real filename when creating fake data.

robhudson:
r+

PR #2880, opened 2015-01-28
Re-organize trending and popularity tasks

robhudson:
This moves everything from tasks to cron. The tasks were hitting a
recursion error b/c we were chaining too many tasks together.

This also combines the calculating of values and reindexing on the same
chunks of apps and adds a records purge so the values in ES are synced
for purged records.


robhudson:
/cc @diox or @washort 

Could either of you get this reviewed in the next 24h? I'd like to get this on -dev and tested before this code ships next week. Thanks.

washort:
Lotta code motion, but looks plausible to me; let's merge it and try it on -dev. r+

PR #2881, opened 2015-01-28
Send a strong signal to delete feed item memberships when deleting a Webapp instance to prevent them from getting garbage collected (bug 1124319).

chuckharmston:
This was a fun one.

### DIFF
--- a/mkt/feed/models.py
+++ b/mkt/feed/models.py
@@ -474,13 +474,11 @@ models.signals.pre_save.connect(
 
 
 # Delete membership instances when their apps are deleted.
-def remove_deleted_app_on(cls):
-    def inner(*args, **kwargs):
-        instance = kwargs.get('instance')
+def remove_memberships(*args, **kwargs):
+    instance = kwargs.get('instance')
+    for cls in [FeedBrandMembership, FeedCollectionMembership,
+                FeedShelfMembership]:
         cls.objects.filter(app_id=instance.pk).delete()
-    return inner
 
-for cls in [FeedBrandMembership, FeedCollectionMembership,
-            FeedShelfMembership]:
-    post_delete.connect(remove_deleted_app_on(cls), sender=Webapp,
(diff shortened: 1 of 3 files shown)


mkt/feed/tests/test_serializers.py
```diff
@@ -284,6 +284,19 @@ def test_is_published(self):
...
         assert data['is_published']
 
+    def test_shelf_get_apps_deleted_app(self):
+        """
+        Regression test for bug 1124319; ensures that deleted apps also delete
+        any respective FeedShelfMembership objects.
+        """
+        app_id = self.app_ids.pop()
+        Webapp.objects.get(pk=app_id).delete()
+        eq_(FeedShelfMembership.objects.filter(app__id=app_id).count(), 0)
```
ngokevin: Can we assert this is 1 beforehand?

ngokevin:
r+

mkt/feed/tests/test_serializers.py
chuckharmston: Done.

chuckharmston:
Thanks!

mkt/feed/models.py
```diff
@@ -474,13 +474,11 @@ def delete_search_index(sender, instance, **kw):
...
+                FeedShelfMembership]:
         cls.objects.filter(app_id=instance.pk).delete()
-    return inner
 
-for cls in [FeedBrandMembership, FeedCollectionMembership,
-            FeedShelfMembership]:
-    post_delete.connect(remove_deleted_app_on(cls), sender=Webapp,
-                        dispatch_uid='apps_collections_cleanup')
+post_delete.connect(remove_memberships, sender=Webapp, weak=False,
+                    dispatch_uid='cleanup_feed_membership')
```
robhudson: That makes more sense. This is one signal checking 3 model classes instead of the other way around.

PR #2882, opened 2015-01-30
Create minifest and download endpoints for language packs (bug 1122279)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1122279


docs/api/topics/submission.rst
```diff
@@ -75,14 +75,14 @@ Validate
 
 .. _validation-response-label:
 
-.. http:get:: /api/v2/apps/validation/(int:id)/
+.. http:get:: /api/v2/apps/validation/(string:id)/
```
diox: I didn't change the API - the docs were wrong.

mkt/downloads/urls.py
```diff
@@ -1,11 +1,14 @@
...
+from mkt.langpacks.views import download as download_langpack
 
 
 urlpatterns = patterns('',
     # .* at the end to match filenames.
     # /file/:id/type:attachment
     url('^file/(?P<file_id>\d+)(?:/type:(?P<type>\w+))?(?:/.*)?',
-        views.download_file, name='downloads.file'),
+        download_file, name='downloads.file'),
+    url('^langpack/(?P<langpack_id>[0-9a-f]{32})(?:/type:(?P<type>\w+))?(?:/.*)?',
```
robhudson: That's a pretty crazy regexp.

mkt/langpacks/models.py
```diff
@@ -66,6 +78,44 @@ def path_prefix(self):
...
+        url = ('%s/langpack.zip' %
+            reverse('downloads.langpack', args=[unicode(self.pk)]))
+        return absolutify(url)
+
+    @property
+    def manifest_url(self):
+        """Return URL to the minifest for the langpack"""
+        if self.active:
+            return absolutify(
+                reverse('langpack.manifest', args=[unicode(UUID(self.pk))]))
```
robhudson: if `self.pk` is a uuid I'm curious why you have to pass it through `UUID` and `unicode`?

mkt/langpacks/views.py
```diff
@@ -60,3 +71,48 @@ def get_serializer_class(self):
...
+            raise Exception(
+                u'Attempting to download langpack %s, '
+                u'which does not have a filename.' % langpack.pk)
+        return HttpResponseSendFile(request, langpack.file_path,
+                                    content_type='application/zip',
+                                    etag=langpack.hash.split(':')[-1])
+    else:
+        raise Http404
+    log.info('Downloading package: %s from %s' % (langpack.pk,
+                                                  langpack.file_path))
```
robhudson: This log looks like it'll never get run.

mkt/langpacks/views.py
```diff
@@ -60,3 +71,48 @@ def get_serializer_class(self):
...
+def manifest(request, uuid):
+    """Returns the "mini" manifest for a langpack."""
+    try:
+        uuid_hex = UUID(uuid).hex
+    except ValueError:
+        raise Http404
+
+    langpack = get_object_or_404(LangPack, pk=uuid_hex)
+
+    if langpack.active or action_allowed(request, 'LangPacks', '%'):
```
robhudson: Is this permission added somewhere? I may have missed that.

mkt/langpacks/models.py
andymckay: Whats the advantage of using a uuid over a pk? And isn't the pk the incrementing id in the db?

diox: `pk` is whatever field has the `primary_key=True` set (django creates one by default if there isn't one). 

As mentioned in the models file, I needed an `UUIDField` to be able to determine the filename before saving the instance. I thought it was redundant to have another field as the primary key when I already had something unique, so I set `uuid` as the primary key.

The reason it's passed through `UUID()` here is to have manifest urls that look like the ones we use for apps. (`/12345678-1234-5678-1234-567812345678/` vs `/12345678123456781234567812345678/`).

mkt/langpacks/views.py
diox: Oops, good catch.

mkt/langpacks/views.py
diox: I was thinking it'd have to be given out manually. Should I create a group that has it ?

robhudson: Yeah, typically we add a migration. And then we manually add users to the group.

robhudson:
r+wc

mkt/langpacks/views.py
diox: Added.

PR #2883, opened 2015-01-30
Fix gitignore for supervisord

muffinresearch:
r? @andymckay 

To prevent:

\#\# master...upstream/master
?? logs/docker.log.1
?? logs/docker.log.2
?? logs/docker.log.3
?? supervisord.pid

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,4 +36,5 @@ media/marketplace-stats
 media/transonic
 *.signed.zip
 *.po~
-logs/*.log
+logs/*
+supervisord.pid


andymckay:
r+

andymckay:
Wont logs/* mean that .keep is ignored which we need to keep the directory around?

muffinresearch:
> Wont logs/* mean that .keep is ignored which we need to keep the directory around?

Should be ok because we're already tracking the .keep file. Updating the .gitignore after that should change that.

PR #2884, opened 2015-01-31
some pep8 fixes oh yeah

andymckay:


.travis.yml
```diff
@@ -8,9 +8,11 @@ before_install:
   - scripts/travis_es.sh
   - /tmp/elasticsearch/elasticsearch-1.3.7/bin/elasticsearch -d -D es.path.data=/tmp -D es.gateway.type=none -D es.index.store.type=memory -D es.discovery.zen.ping.multicast.enabled=false
 install:
+  - pip install flake8
```
diox: we should just have it in `test.txt`

.travis.yml
```diff
@@ -8,9 +8,11 @@ before_install:
...
   - pip install --no-deps --exists-action=w --download-cache=/tmp/pip-cache -r requirements/test.txt --find-links https://pyrepo.addons.mozilla.org/wheelhouse/
 before_script:
   - mysql -e 'create database zamboni;'
+  - flake8 . --exclude=./docs/conf.py
```
robhudson: At least for me personally, I usually `--ignore=E125,E126,E127,E128`.

mkt/site/context_processors.py
```diff
@@ -14,7 +14,7 @@
 def i18n(request):
     return {'LANGUAGES': settings.LANGUAGES,
             'LANG': settings.LANGUAGE_URL_MAP.get(translation.get_language())
-                    or translation.get_language(),
+            or translation.get_language(),
```
robhudson: This is less readable to me :-1:

mkt/__init__.py
```diff
@@ -1,18 +1,18 @@
...
-from mkt.constants.platforms import *
-from mkt.constants.search import *
-from mkt.constants.submit import *
+                           ratingsbodies)  # flake8: noqa
+from mkt.constants.applications import *  # flake8: noqa
+from mkt.constants.base import *  # flake8: noqa
+from mkt.constants.payments import *  # flake8: noqa
+from mkt.constants.platforms import *  # flake8: noqa
+from mkt.constants.search import *  # flake8: noqa
+from mkt.constants.submit import *  # flake8: noqa
```
robhudson: Not crazy about all the `noqa` droppings everywhere.

mkt/users/helpers.py
```diff
@@ -30,7 +30,7 @@ def emaillink(email, title=None, klass=None):
 
 @register.function
 def user_data(user):
-    anonymous, currency, pre_auth, email = True, 'USD', False, ''
+    anonymous, currency, _, email = True, 'USD', False, '' # flake8: noqa
```
robhudson: perhaps avoid `_` as that's usually used for gettext?

mkt/webapps/query.py
```diff
@@ -82,13 +82,14 @@ def get_from_clause(self):
...
-                name, alias, join_type, lhs, join_cols, _, join_field = self.query.alias_map[alias]
+                name, alias, join_type, lhs, join_cols, _, join_field = (
+                    self.query.alias_map[alias])
             except KeyError:
                 # Extra tables can end up in self.tables, but not in the
                 # alias_map if they aren't in a join. That's OK. We skip them.
                 continue
             alias_str = (alias != name and ' %s' % alias or '')
-            ### jbalogh wuz here. ###
+            # jbalogh wuz here.
```
robhudson: good comment to keep

robhudson:
I'm good with this change. I like that we do it once and require it for future commits so it doesn't drift.

mkt/api/urls.py
```diff
@@ -41,7 +41,8 @@ def include_version(version):
     return include('mkt.api.v%d.urls' % version, **kwargs)
 
 
-urlpatterns = patterns('',
+urlpatterns = patterns(
+    '',
```
diox: I find the previous version more readable :(

I wonder if, here and in the various places where we call `patterns()` with `''` as the prefix, we actually need the function call at all. All `patterns()` does is add a prefix and wrap tuples of url+function in `url()` calls, if you are already using `url()` everywhere and you don't set a prefix, you could just do `urlpatterns = [...]`.

mkt/site/context_processors.py
mstriemer: I think if you wrap the statement in parens it uses the formatting most people would want.

```python
    'LANG': (settings.LANGUAGE_URL_MAP.get(translation.get_language())
             or translation.get_language()),

mkt/__init__.py
mstriemer: Yeah, it would be nice if there's a way to allow `import *` in an `__init__.py` file.

.travis.yml
mstriemer: Agreed, we should probably add the `flake8 . --exclude=./docs/conf.py` command to the Makefile also so that you can easily check this locally.

mkt/users/helpers.py
andymckay: yeah i've normally advocated for that will do

mkt/__init__.py
andymckay: me neither, but then again, the import * is just wrong in the first place, i think we should stopping do that

mkt/site/context_processors.py
andymckay: yeah that works and is more readable, thanks

kumar303:
I am slightly concerned at how you spend your weekends :eyes:

mkt/api/urls.py
andymckay: We could, its not great that the first argument is always ''. Does this block the PR for you? Trying to avoid changing too much code in the PR as it is.

diox: Doesn't block. I'll file a bug to replace the superfluous `patterns()` calls.

andymckay: Cool, thanks.

.travis.yml
andymckay: Chatted to rob and we thought that since its all working with these rules in, we'd leave them in. In the future we won't ignore these rules.

robhudson:
I'm good with this. :ship: it. :boom:

PR #2887, opened 2015-02-03
Move langpack name to its own property, expose it to the API (bug 1122256)

diox:
Needed for https://bugzilla.mozilla.org/show_bug.cgi?id=1122256

### DIFF
--- a/mkt/langpacks/tests/test_models.py
+++ b/mkt/langpacks/tests/test_models.py
@@ -2,6 +2,7 @@
 import json
 import os
 
+from django.conf import settings
 from django.core.files.storage import default_storage as storage
 from django.forms import ValidationError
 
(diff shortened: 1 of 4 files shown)


mkt/langpacks/tests/test_models.py
```diff
@@ -44,6 +44,10 @@ def test_get_minifest_contents(self):
              'package_path': langpack.download_url,
              'developer': {'name': 'Mozilla'}})
 
+    def test_name(self):
+        langpack = LangPack(fxos_version='2.5.1')
+        eq_(langpack.name, u'English (US) language pack for Firefox OS 2.5.1')
```
diox: There is a comment above that also applies here: we should test this with translations, but need the weekly l10n extraction to be done and translations to be submitted first.

robhudson: Is there a bug? Or maybe add an extra `TODO`?

robhudson:
r+wc

mkt/langpacks/tests/test_models.py
diox: I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1129409

PR #2889, opened 2015-02-04
Fix region iterations on devhub (1129169)

muffinresearch:
When Boku accounts were added we're bailing early on the loop when we should be continuing to go through the rest of the data.


diox:
r+

PR #2890, opened 2015-02-04
Only include valid apps in search with no_filter (bug 1128783)

robhudson:


mkt/search/tests/test_views.py
```diff
@@ -796,6 +796,25 @@ def test_no_filter(self):
         res = self.client.get(self.url, data={'filtering': 0})
         eq_(len(res.json['objects']), 1)
 
+    def test_no_filter_excludes_deleted_or_disabled(self):
```
chuckharmston: This should probably be two different tests. My rubric: if a failure in different parts would be  of a failure of a different code path, it should be housed in a different test.

diox:
There are useful existing tests in `mkt/search/tests/test_filters.py`, maybe look if it's appropriate to add/modify a test there as well ?

chuckharmston:
Great tests, I think this is the right fix for a kind of confusing bug. Nice job! r+

I think there are some other tests that might be affected in `mkt.search`, make sure the build passes before merging.

robhudson:
Yep, I was waiting for Travis to tell me if I missed anything.

mkt/search/tests/test_views.py
robhudson: Ok. I'll separate the tests.

PR #2891, opened 2015-02-06
generate unique icons/previews for fake apps (bug 1065620)

washort:
this uses `pydenticon` to create patterned images instead of solid colors.

### DIFF
--- a/mkt/webapps/fakedata.py
+++ b/mkt/webapps/fakedata.py
@@ -4,13 +4,12 @@ import itertools
 import os
 import json
 import random
-import StringIO
 import tempfile
 from zipfile import ZipFile, ZIP_DEFLATED
 
 from django.conf import settings
 
-from PIL import Image
+import pydenticon
 
 import mkt
 from mkt.constants.applications import DEVICE_TYPES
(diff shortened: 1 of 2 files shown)


requirements/prod.txt
```diff
@@ -71,6 +71,7 @@ polib==1.0.3
 protobuf==2.5.0
 pyasn1==0.1.7
 PyBrowserID==0.6
+pydenticon==0.2
```
kumar303: is this in our repo? I forget what we use now. I assume it's no longer https://pyrepo.addons.mozilla.org/ because that cert has been revoked :)

mkt/webapps/fakedata.py
```diff
@@ -57,28 +56,36 @@ def generate_app_data(num):
...
-                               caption="screenshot " + str(n),
-                               position=n)
-    f = tempfile.NamedTemporaryFile()
-    im.save(f, 'png')
-    resize_preview(f.name, p)
+def generate_previews(app, n=1):
+    gen = pydenticon.Generator(8, 12, foreground=foreground,
+                               digest=hashlib.sha512)
+    for i in range(n):
+        img = gen.generate(unicode(app.name).encode('utf8') + chr(i), 320, 480,
```
kumar303: ``chr(i)`` ? Wouldn't that add line breaks, tabs, or NULLs in some cases?

kumar303:
r+wc

mkt/webapps/fakedata.py
washort: Sure. but that's just used to generate a hash, nothing else uses this string.

requirements/prod.txt
washort: haven't uploaded it yet.

mkt/webapps/fakedata.py
kumar303: oh, that works then. I thought it was the file name.

PR #2892, opened 2015-02-06
Store manifest when upload a langpack and simpliy model (bug 1122273)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1122273

We need to keep some keys in the minifest identical to the manifest, before the platform compares the two and will complain if some keys like name, developer, locales are different.

This means we can drop the custom name property. In addition, some refactoring was needed to be able to share webapps manifest parsing and minifest generation, and as a result we're able to drop a few fields.

### DIFF
--- a/migrations/894-simplify-langpacks.sql
+++ b/migrations/894-simplify-langpacks.sql
@@ -1,0 +1,11 @@
+-- filename is now automatically generated.
+ALTER TABLE `langpacks_langpack` DROP COLUMN `filename`;
+
+-- hash was only used for the etag, which is calculated differently now.
+ALTER TABLE `langpacks_langpack` DROP COLUMN `hash`;
+
+-- size was only used in the manifest, it's now automatically generated.
+ALTER TABLE `langpacks_langpack` DROP COLUMN `size`;
+
+-- original manifest is important to be able to serve a correct minifest.
+ALTER TABLE `langpacks_langpack` ADD COLUMN `manifest` longtext NOT NULL;
--- a/mkt/langpacks/models.py
+++ b/mkt/langpacks/models.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-import hashlib
 import json
 import os.path
 from uuid import UUID
@@ -9,11 +8,9 @@ from django.core.files.storage import default_storage as storage
 from django.core.urlresolvers import reverse
 from django.db import models
 from django.utils.functional import lazy
-from django.utils import translation
 
 import commonware.log
 from django_statsd.clients import statsd
-from tower import ugettext as _
 from uuidfield.fields import UUIDField
 
 from lib.crypto.packaged import sign_app, SigningError
@@ -23,6 +20,7 @@ from mkt.translations.utils import to_language
 from mkt.site.helpers import absolutify
 from mkt.site.models import ModelBase
 from mkt.site.utils import smart_path
+from mkt.webapps.models import get_cached_minifest
 
 
 log = commonware.log.getLogger('z.versions')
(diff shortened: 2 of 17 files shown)


mkt/webapps/models.py
```diff
@@ -1803,57 +1803,20 @@ def get_cached_manifest(self, force=False):
...
         If the addon is not a packaged app, this will not cache anything.
 
+        Ensure that the calling method checks various permissions if needed.
+        E.g. see mkt/detail/views.py. This is also called as a task after
+        reviewer approval so we can't perform some checks here.
         """
         if not self.is_packaged:
             return
 
-        key = 'webapp:{0}:manifest'.format(self.pk)
```
diox: I moved the contents of this method to `utils.py` to share it with langpacks. There was a big ugly bug (that is now fixed and tested) : caching never worked, because `key` here was overwritten in the loop on the manifest below, right before saving.

migrations/894-simplify-langpacks.sql
```diff
@@ -0,0 +1,11 @@
...
+ALTER TABLE `langpacks_langpack` DROP COLUMN `filename`;
+
+-- hash was only used for the etag, which is calculated differently now.
+ALTER TABLE `langpacks_langpack` DROP COLUMN `hash`;
+
+-- size was only used in the manifest, it's now automatically generated.
+ALTER TABLE `langpacks_langpack` DROP COLUMN `size`;
+
+-- original manifest is important to be able to serve a correct minifest.
+ALTER TABLE `langpacks_langpack` ADD COLUMN `manifest` longtext NOT NULL;
```
robhudson: The comments here are really useful. :+1:

mkt/langpacks/models.py
```diff
@@ -201,6 +178,8 @@ def from_upload(cls, upload, instance=None):
         instance.handle_file_operations(upload)
         # Save!
         instance.save()
+        # Bust caching of manifest by passing force=True.
+        instance.get_minifest_contents(force=True)
```
robhudson: Should this happen only when active? Or does it become active here?

mkt/webapps/models.py
robhudson: Wow, nice find. That bug is mine, and it is big and ugly.

mkt/webapps/utils.py
```diff
@@ -79,3 +85,57 @@ def iarc_get_app_info(app):
...
+        if data:
+            return data
+
+    sign_if_packaged = getattr(app_or_langpack, 'sign_if_packaged', None)
+    if sign_if_packaged is None:
+        # Langpacks are already signed when we generate the manifest and have
+        # a file_path attribute.
+        signed_file_path = app_or_langpack.file_path
+    else:
+        # sign_if_packed() will return the signed path. But to call it, we need
```
robhudson: `sign_if_packaged()`

mkt/langpacks/models.py
diox: Since we are completely replacing the langpack everytime, it doesn't matter.

mkt/webapps/utils.py
```diff
@@ -79,3 +85,57 @@ def iarc_get_app_info(app):
...
+        # and they don't have release notes.
+        data['version'] = app_or_langpack.version
+
+    for key in ['developer', 'icons', 'locales', 'name']:
+        if key in manifest:
+            data[key] = manifest[key]
+
+    data = json.dumps(data, cls=JSONEncoder)
+    cache.set(cache_key, data, None)
+    return data
```
robhudson: Some of this is complicated with `if` statements b/c of the differences between apps and langpacks. I'm curious if passing in a `type='app'` or `type='langpack'` would help?

mkt/webapps/tests/test_utils_.py
```diff
@@ -31,3 +39,50 @@ def test_short_locale(self):
...
+        get_cached_minifest(self.webapp)
+
+        ok_(not cache.get(
+            'langpack:12345678123456781234567812345678:manifest'))
+        langpack = LangPack(pk='12345678123456781234567812345678',
+                            manifest='{}')
+        get_cached_minifest(langpack)
+
+        ok_(cache.get('webapp:337141:manifest'))
+        ok_(cache.get('langpack:12345678123456781234567812345678:manifest'))
```
robhudson: :+1:

robhudson:
r+wc

PR #2894, opened 2015-02-06
Better --noop for load_prices

kumar303:


washort:
nice, i'd forgotten about `--delete`, that's handy. r+

kumar303:
``--delete`` rarely actually works because there are typically too many related foreign keys (apps and purchase history). Now that this command doesn't overwrite existing prices you don't really need it though.

PR #2895, opened 2015-02-06
Override flake8 defaults, which adds E226, E241 and E242

robhudson:


Makefile
```diff
@@ -88,3 +88,7 @@ deploy_release:
 
 release: tag_release deploy_release
 
+# Make sure the following matches what is in the .travis.yml file.
+flake:
+	flake8 . --exclude=./docs/conf.py --ignore=E123,E133
+
```
andymckay: according to the docs we could make a setup.cfg in the root of zamboni and then "flake8" from the command line in the root would work

http://flake8.readthedocs.org/en/latest/config.html#default

andymckay:
r+wc. stricter is good, although i think having flake8 work out of the box would be nice

robhudson:
Added a setup.cfg.

robhudson:
https://github.com/mozilla/zamboni/commit/99435aa

PR #2896, opened 2015-02-07
Use regional popularity and trending (bug 1127465)

robhudson:


mkt/search/views.py
```diff
@@ -109,18 +115,20 @@ def _sort_search(request, sq, data):
     """
     from mkt.api.base import get_region_from_request
```
diox: You didn't add this import, but can it be moved to the top ?

mkt/search/views.py
```diff
@@ -109,18 +115,20 @@ def _sort_search(request, sq, data):
...
 
+    region = get_region_from_request(request)
+
     # When querying we want to sort by relevance. If no query is provided,
     # i.e. we are only applying filters which don't affect the relevance,
     # we sort by popularity descending.
```
diox: Also something you didn't add, but can you add in this comment a "By default" or "If no sort param is provided", something like that ? It's confusing to read considering what happens below when a sort parameter is provided.

mkt/search/views.py
robhudson: should be fine, yeah, there's already `mkt.api.base` stuff up there.

mkt/search/views.py
```diff
@@ -101,6 +101,12 @@ def name_query(q):
     return query.Bool(should=should)
 
 
+def _get_regional_sort(region, field):
+    if region and not region.adolescent:
```
diox: Please add docstring explaining that mature regions get their own field to sort by.

mkt/search/views.py
```diff
@@ -101,6 +101,12 @@ def name_query(q):
...
+        return ['-%s_%s' % (field, region.id)]
+    return ['-%s' % field]
```
diox: That wasn't here before, so, before this PR, it was impossible for an adolescent region to get explicit sorting of a query by popularity or trending ?

robhudson:
https://github.com/mozilla/zamboni/commit/9bf68dd

PR #2898, opened 2015-02-09
Only show paid regions if supported (bug 1126002)

muffinresearch:
In order to only move the checkboxes if they match the account this branch:

* Exposes the enabled payment account providers
* Use that to check what regions to move
* Fires a custom event on account deletion so that the checkbox UI gets updated if the associated account was deleted
* As a bonus this fixes a bug where the overlayed class wasn't removed during the confirm deletion flow.

### DIFF
--- a/media/js/devreg/payments-manage.js
+++ b/media/js/devreg/payments-manage.js
@@ -56,18 +56,32 @@ define('payments-manage', ['payments'], function(payments) {
             });
             $ul.append($el);
         });
+
         $confirm_delete_overlay.on('click', 'a.payment-account-delete-confirm', _pd(function() {
             $.post(data['delete-url']).then(function () {
-                refreshAccountForm(data)
+                refreshAccountForm(data);
             });
+
             $confirm_delete_overlay.remove();
+            z.body.removeClass('overlayed');
+            var accountName = data.name;
             var currentAppName = data['current-app-name'];
             var paymentAccounts = data['app-payment-accounts'][currentAppName];
             // If this app is associated with the account and it only has one
             // account show the warning.
             if (paymentAccounts && paymentAccounts.length === 1) {
                 $('#paid-island-incomplete').removeClass('hidden');
+                $('.no-payment-regions').removeClass('hidden');
             }
+
+            if (paymentAccounts.indexOf(accountName) > -1 && data.provider) {
+                // Fire a custom event for the account deletion so we can update
+                // The regions UI as necessary.
+                console.log('Firing a custom event for the payment account deletion: ' +
+                    JSON.stringify({account: accountName, provider: data.provider}));
+                $('body').trigger('app-payment-account-deletion', {provider: data.provider});
+            }
+
         }));
     }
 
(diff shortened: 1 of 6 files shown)


media/js/devreg/payments-manage.js
```diff
@@ -56,18 +56,32 @@ define('payments-manage', ['payments'], function(payments) {
...
             if (paymentAccounts && paymentAccounts.length === 1) {
                 $('#paid-island-incomplete').removeClass('hidden');
+                $('.no-payment-regions').removeClass('hidden');
             }
+
+            if (paymentAccounts.indexOf(accountName) > -1 && data.provider) {
+                // Fire a custom event for the account deletion so we can update
+                // The regions UI as necessary.
+                console.log('Firing a custom event for the payment account deletion: ' +
+                    JSON.stringify({account: accountName, provider: data.provider}));
```
kumar303: probably better to use a custom log object (with a prefix) here

media/js/devreg/payments.js
```diff
@@ -186,6 +193,11 @@ define('payments', [], function() {
                     var regionId = price.region;
                     var regionSeen = seenRegions.indexOf(regionId) > -1;
 
+                    if (enabledProviderIds.indexOf(price.provider) === -1) {
+                        console.log('Continuing as provider is not found in enabled provider ids list');
```
kumar303: ditto about the log

media/js/devreg/payments-manage.js
kumar303: oh maybe that's fireplace only? Ah, no worries

media/js/devreg/payments.js
```diff
@@ -355,6 +370,19 @@ define('payments', [], function() {
         }
     }
 
+    $('body').on('app-payment-account-deletion', function(e, data) {
+        var providerId = providerIdLookup[data.provider];
+        if (providerId) {
+            var idx = enabledProviderIds.indexOf(parseInt(providerId, 10));
+            if (idx > -1) {
+                enabledProviderIds.splice(idx, 1);
+                //$regions.data('enabledProviderIds', enabledProviderIds);
```
kumar303: still need this?

mkt/developers/views_payments.py
```diff
@@ -180,6 +180,9 @@ def payments(request, addon_id, addon, webapp=False):
                    'all_paid_region_ids_by_name': paid_region_ids_by_name,
                    'providers': providers,
                    'provider_regions': provider_regions,
+                   'enabled_provider_ids':
+                       [acct.payment_account.provider
```
kumar303: shouldn't this be ``acct.payment_account.provider.pk`` to get the ID?

kumar303:
might be nice to add a django test for ``data-enabled-provider-ids`` in the template so that no one deletes it accidentally

kumar303:
r+wc

muffinresearch:
> might be nice to add a django test for data-enabled-provider-ids in the template so that no one deletes it accidentally

Yep, I was planning to do that.

media/js/devreg/payments.js
muffinresearch: Well spotted.

mkt/developers/views_payments.py
muffinresearch: This is directly the id because providers are constants. https://github.com/mozilla/zamboni/blob/master/mkt/developers/models.py#L77

kumar303: Huh. That's confusing. That model field would be better named ``provider_id`` :fried_shrimp:

PR #2900, opened 2015-02-10
Update prices for no-edit cases too (bug 1006474)

muffinresearch:


kumar303:
Seems like a good idea. r+

PR #2905, opened 2015-02-10
Pass L10n app data to Webpay in JWT (bug 1131708)

kumar303:
@diox r?

### DIFF
--- a/mkt/webpay/tests/test_webpay_jwt.py
+++ b/mkt/webpay/tests/test_webpay_jwt.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 import json
 import urlparse
 from urllib import urlencode
@@ -7,6 +8,7 @@ from django.core.urlresolvers import reverse
 from django.test.utils import override_settings
 
 import jwt
+import mock
 from mozpay.verify import verify_claims, verify_keys
 from nose.tools import eq_, ok_, raises
 
@@ -59,6 +61,7 @@ class TestPurchaseJWT(PurchaseTest):
         request = token_data['request']
         eq_(request['id'], self.product.external_id())
         eq_(request['name'], self.product.name())
+        eq_(request['defaultLocale'], self.product.default_locale())
         eq_(request['icons'], self.product.icons())
         eq_(request['description'], self.product.description())
         eq_(request['pricePoint'], self.product.price().name)
@@ -82,6 +85,19 @@ class TestPurchaseJWT(PurchaseTest):
         request = token_data['request']
         product = urlparse.parse_qs(request['productData'])
         ok_('buyer_email' not in product)
+
+    def test_locales(self):
+        with mock.patch.object(self.product, 'localized_properties') as props:
+            loc_data = {
+                'es': {
+                    'name': 'El Mocoso',
+                    'description': u'descripción de la aplicación',
+                }
+            }
+            props.return_value = loc_data
+            token_data = self.decode_token()
+            # Make sure the JWT passes through localized_properties() data.
+            eq_(token_data['request']['locales'], loc_data)
 
 
 class BaseTestWebAppProduct(PurchaseTest):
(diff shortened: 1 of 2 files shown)


mkt/webpay/tests/test_webpay_jwt.py
```diff
@@ -125,6 +145,30 @@ def test_webapp_product(self):
         eq_(product_data['addon_id'], self.product.addon().pk)
         eq_(product_data['application_size'], self.product.application_size())
 
+    @mock.patch.object(settings, 'AMO_LANGUAGES', ('en-US', 'es', 'fr'))
```
mstriemer: You can use [`django.test.utils.override_settings`](https://docs.djangoproject.com/en/1.6/topics/testing/tools/#django.test.utils.override_settings) instead of `mock`. Looks like it's already imported.

```py
@override_settings(AMO_LANGUAGES=('en-US', 'es', 'fr'))
def test_localized_properties(self):
  # ...

File: mkt/webpay/webpay_jwt.py [hunk 2]
```diff
@@ -18,7 +19,14 @@
 
 
 def get_product_jwt(product, contribution):
-    """Prepare a JWT for paid products to pass into navigator.pay()"""
+    """
+    Prepare a JWT describing the item about to be purchased when
+    working with navigator.mozPay()
+
+    See the MDN docs for details on the JWT fields:
+    https://developer.mozilla.org/en-US/Marketplace/Monetization
```
mstriemer: Does pyflakes get mad if you leave this URL on one line? I feel like it should be okay to leave this as one line in a docstring.

mkt/webpay/webpay_jwt.py
```diff
@@ -18,7 +19,14 @@
 
 
 def get_product_jwt(product, contribution):
-    """Prepare a JWT for paid products to pass into navigator.pay()"""
+    """
+    Prepare a JWT describing the item about to be purchased when
+    working with navigator.mozPay()
```
mstriemer: This should have a `.` at the end.

mkt/webpay/webpay_jwt.py
kumar303: yeah, it says line is too long :)

mkt/webpay/tests/test_webpay_jwt.py
```diff
@@ -83,6 +86,22 @@ def test_no_user(self):
...
+        }
+        self.addon.description = {
+            'fr': u"ceci est une description d'application",
+            'es': u'se trata de una descripción de la aplicación',
+        }
+        self.addon.save()
+
+        token_data = self.decode_token()
+        eq_(token_data['request']['locales'],
+            self.product.localized_properties())
```
mstriemer: This test knows a lot about `product.localized_properties()`. I'd rather tests like these just use stubs rather than requiring all of our business logic and django database stuff but there's a lot going on here and that might be too much of a refactor at this point.

I'd rather see something like this instead of setting `addon.name` and `addon.description` which seem irrelevant in this context:

```py
with mock.patch.object(self.product, 'localized_properties') as local_props:
    local_props.return_value = 'foobar'
    token_data = self.decode_token()
    eq_(token_data['request']['locales'], 'foobar')

Comment (kumar303) [inline mkt/webpay/tests/test_webpay_jwt.py]: oh yeah, good idea

File: mkt/webpay/webpay_jwt.py [hunk 5]
```diff
@@ -77,6 +87,20 @@ def external_id(self):
...
 
+    def default_locale(self):
+        return self.webapp.default_locale
+
+    def localized_properties(self):
+        props = {}
+
+        for attr in ('name', 'description'):
+            tr_object = getattr(self.webapp, attr)
+            for tr in Translation.objects.filter(id=tr_object.id):
```
mstriemer: Is there no way to retrieve this from `tr_object` directly? This seems fragile if we add support for activating/deactivating locales for example.

mstriemer:
r+wc

mkt/webpay/webpay_jwt.py
kumar303: mmm, yeah, lots of worms, lots of cans. the FIXMEs in ``translations/models.py`` are scary. I couldn't see of a more direct way to do this but I was hoping @diox might be able to take a look. When manifests are updated we do at least [attempt to delete transactions](https://github.com/mozilla/zamboni/blob/master/mkt/webapps/models.py#L871).

kumar303:
Thanks for the r!

PR #2906, opened 2015-02-11
Replace custom search filtering with DRF filter classes (bug 1128783)

robhudson:


mkt/reviewers/forms.py
```diff
@@ -200,7 +200,7 @@ def __init__(self, *args, **kwargs):
     def clean_status(self):
         status = self.cleaned_data['status']
         if status == 'any':
-            return 'any'
+            return None
```
cvan: any implication from changing this?

robhudson: We used to explicitly check for not `'any'` to apply the filter (https://github.com/mozilla/zamboni/pull/2906/files#diff-9a26e6efe2570209f70b06294f23ca39L1007) but now we check for not `None` (https://github.com/mozilla/zamboni/pull/2906/files#diff-26e55b38d2cbc410af290d209506f3d7R120)

mkt/search/views.py
```diff
@@ -170,43 +32,29 @@ class SearchView(CORSMixin, MarketplaceView, GenericAPIView):
...
-            request, search_form_to_es_fields(form_data), no_filter=no_filter)
-
-        # Sort.
-        sq = _sort_search(request, sq, form_data)
-
-        # Done.
-        page = self.paginate_queryset(sq)
-        return self.get_pagination_serializer(page), form_data.get('q', '')
-
     def get(self, request):
```
diox: Do you need a specific `get` method anymore or can you use DRF's `ListModelMixin` ? Also, can you drop `ReviewersSearchView.get` and just let it inherit from whatever the base `SearchView` class has ?

mkt/search/views.py
```diff
@@ -170,43 +32,29 @@ class SearchView(CORSMixin, MarketplaceView, GenericAPIView):
...
 
 class FeaturedSearchView(SearchView):
+
     def get(self, request, *args, **kwargs):
-        serializer, _ = self.search(request)
+        queryset = self.filter_queryset(self.get_queryset())
+        page = self.paginate_queryset(queryset)
+        serializer = self.get_pagination_serializer(page)
         data = self.add_featured_etc(request, serializer.data)
         response = Response(data)
```
diox: I think you could just call super() and add stuff to response.data

mkt/search/views.py
```diff
@@ -229,14 +77,17 @@ class SuggestionsView(SearchView):
     serializer_class = SuggestionsESAppSerializer
 
     def get(self, request, *args, **kwargs):
-        results, query = self.search(request)
+        query = request.GET.get('q', '')
+        queryset = self.filter_queryset(self.get_queryset())
+        page = self.paginate_queryset(queryset)
+        serializer = self.get_pagination_serializer(page)
```
diox: Not sure about this one, but also worth trying using `super()` here.

mkt/search/forms.py
```diff
@@ -177,8 +179,17 @@ def clean_device_and_dev(self):
...
     def clean_author(self):
-        return self.cleaned_data.get('author', '').lower()
+        author = self.cleaned_data.get('author')
+        if author:
+            return author.lower()
 
     def clean(self):
         self.clean_device_and_dev()
+
+        # Convert empty things to `None`s.
```
diox: Is his always going to be right ? it seems like it is now, but we may want to differentiate between `0` and `None` in the future. Maybe not worth caring about now though.

robhudson: I could add an explicit check for `0` also, like I do below for False. I agree, this is shady. The other option is to adjust (or add) the `clean_*` methods to return `None` for their specific case.

mkt/search/views.py
robhudson: I'm sure you know DRF better than I do. If that's something I should look at I will.

mkt/search/views.py
robhudson: Oh nice. I'll update it.

mkt/search/views.py
robhudson: You're right. I used `ListAPIView` which brings in the `ListModelMixin` which does exactly what our `get` was doing so I removed it.

robhudson:
@diox Mind another pass? I've left the follow up commits separate if that's easier. I'll squash before merging.

mkt/api/paginator.py
```diff
@@ -39,13 +39,13 @@ def page(self, number):
         number = self.validate_number(number)
         bottom = (number - 1) * self.per_page
         top = bottom + self.per_page
-        page = Page(self.object_list[bottom:top], number, self)
 
         # Force the search to evaluate and then attach the count. We want to
         # avoid an extra useless query even if there are no results, so we
         # directly fetch the count from hits.
         # Overwrite `object_list` with the list of ES results.
```
diox: That comment is no longer true, needs updating - we are not directly overwriting. I am pretty surprised it still works without needing the overwrite, btw. Can you double check that we are not making extra ES queries ? (I don't fully trust our tests here)

mkt/reviewers/views.py
```diff
@@ -992,43 +993,11 @@ class ReviewersSearchView(SearchView):
...
-    if data.get('is_escalated') is not None:
-        qs = qs.filter('term', is_escalated=data['is_escalated'])
-    is_tarako = data.get('is_tarako')
-    if is_tarako is not None:
-        if is_tarako:
-            qs = qs.filter('term', tags='tarako')
-        else:
-            qs = qs.filter(~F('term', tags='tarako'))
-    return qs
+    paginator_class = ESPaginator
```
diox: Do you really need this? It is already defined to the same value by the base class, `SearchView`.

mkt/api/paginator.py
robhudson: Calling `Page(result.hits, ...)` sets `object_list` to `result.hits` on the returned `page` object. Effectively that comment is still true but indirectly. I'll update it to be more correct.

mkt/reviewers/views.py
robhudson: If it's in the super class, then no. I'll remove it.

mkt/api/paginator.py
diox: I wonder why we manually set it before, since we were already instantiating the `Page` object...

robhudson: I wrote the original and find this change better and more straight forward. It's executing the ES query then creating the `Page` object from the results. Previously we created the `Page` object passing in the es-dsl `Search` object and then reaching inside and replacing `object_list` with the search results, which doesn't quite make sense.

diox: Yeah, the new version totally makes more sense now ; It was just confusing when reading the diff.

diox:
r+wc

robhudson:
Squashed and merged in https://github.com/mozilla/zamboni/commit/004911b

PR #2907, opened 2015-02-11
Fix star rating display and anonymous review flagging (bug 890175)

eviljeff:


mkt/ratings/helpers.py
```diff
@@ -13,15 +12,7 @@ def stars(num, large=False):
...
-        t = jingo.env.get_template('ratings/reviews_rating.html')
-        # These are getting renamed for contextual sense in the template.
-        return jinja2.Markup(t.render({'rating': num, 'detailpage': large}))
-
-
-@jingo.register.function
-def impala_reviews_link(addon):
-    t = jingo.env.get_template('ratings/reviews_link.html')
-    return jinja2.Markup(t.render({'addon': addon}))
+        return _('Rated %s out of 5 stars' % num)
```
robhudson: One says "Not yet reviewed" and the other says "Rated...". I wonder if we should be consistent in calling them ratings or reviews?

mkt/reviewers/tests/test_views.py
```diff
@@ -3230,6 +3230,16 @@ def _post(self, action):
     def _get_logs(self, action):
         return ActivityLog.objects.filter(action=action.id)
 
+    def test_anonymous_flagger(self):
+        # raise SkipTest
```
robhudson: leftovers?

mkt/reviewers/templates/reviewers/includes/reviews_link.html
```diff
@@ -0,0 +1,11 @@
+{% set reviewbase = product.get_url_path()|urlparams('reviews') %}
```
robhudson: Since this isn't reused I would just inline it into the anchor tag.

robhudson:
Nice cleanup. r+wc

mkt/ratings/helpers.py
eviljeff: I think we can blame that on the schizophrenic rating/review terminology on Marketplace front end (and back end).  I'm tempted to use 'Review' as that's at what we're labelling them for consumers, for now at least.

PR #2908, opened 2015-02-11
Some jshint fun. Or terror. One of those.

andymckay:
This should get the easy ones. The harder problems are left now:
* [x] media/js/common/buckets.js
* [ ] media/js/common/upload-base.js
* [ ] media/js/devreg/devhub.js
* [ ] media/js/devreg/ecosystem.js
* [ ] media/js/devreg/inapp_payments.js
* [ ] media/js/devreg/init.js
* [x] media/js/devreg/payments-enroll.js 

For help, chat, status please see: https://etherpad.mozilla.org/jshint-zamboni

### DIFF
--- a/media/js/common/buckets.js
+++ b/media/js/common/buckets.js
@@ -5,7 +5,7 @@ define('buckets', [], function() {
 
     // Compatibilty with PhantomJS, which doesn't implement canPlayType
     if (!('canPlayType' in aelem)) {
-        function noop() {return '';};
+        function noop() {return '';}
         velem = aelem = {canPlayType: noop};
     }
 
@@ -25,7 +25,7 @@ define('buckets', [], function() {
         // Camel-case it.
         property = property[0].toUpperCase() + property.substr(1);
 
-        for (var i = 0, e; e = prefixes[i++];) {
+        for (var i = 0; e; e = prefixes[i++]) {
             try {
                 if ((e + property) in context) {
                     return context[e + property];
@@ -52,14 +52,14 @@ define('buckets', [], function() {
         'mozApps' in navigator && navigator.mozApps.installPackage,
         'mozPay' in navigator,
         // FF 18 and earlier throw an exception on this key
-        (function() {try{return !!window.MozActivity} catch(e) {return false;}})(),
+        (function() {try{return !!window.MozActivity;} catch(e) {return false;}})(),
         'ondevicelight' in window,
         'ArchiveReader' in window,
         'battery' in navigator,
         'mozBluetooth' in navigator,
         'mozContacts' in navigator,
         'getDeviceStorage' in navigator,
-        (function() { try{return window.mozIndexedDB || window.indexedDB} catch(e) {return false}})(),
+        (function() { try{return window.mozIndexedDB || window.indexedDB;} catch(e) {return false}})(),
         'geolocation' in navigator && 'getCurrentPosition' in navigator.geolocation,
         'addIdleObserver' in navigator && 'removeIdleObserver' in navigator,
         'mozConnection' in navigator && (navigator.mozConnection.metered === true || navigator.mozConnection.metered === false),
--- a/media/js/common/formsets.js
+++ b/media/js/common/formsets.js
@@ -15,7 +15,7 @@
  *   </div>
  *   <a href="#" id="add">{{ _('Add formset row') }}</a>
  * </div>
- * 
+ *
  * $('#formset').formset();
  * $('#add').on('click', function(evt) {
  *   $('#formset').formset('addRow');
@@ -31,7 +31,7 @@
  *   'totalForms': '-TOTAL_FORMS',  // End of `name` attr of hidden field indicating total number of
  *                                  // forms (useful if you use multiple formsets in a single <form>)
  * })
- *  
+ *
  */
 
 (function($) {
@@ -75,7 +75,7 @@ var methods = {
         return $(this).data('options');
     },
 
-    addRow: function() { 
+    addRow: function() {
 
         var $this = $(this),
             options = $this.data('options'),
@@ -93,7 +93,7 @@ var methods = {
                     oldVal = $elem.attr(attr),
                     newVal = oldVal.replace(options.prefix, newNum);
                 $elem.attr(attr, newVal);
-            })
+            });
         });
 
         // Insert new row
--- a/media/js/devreg/devhub.js
+++ b/media/js/devreg/devhub.js
@@ -23,7 +23,7 @@ $(document).ready(function() {
     // Save manifest if passed in
     var params = z.getVars();
     if('manifest' in params && z.capabilities.sessionStorage) {
-        window.sessionStorage['manifest_url'] = params['manifest'];
+        window.sessionStorage.manifest_url = params.manifest;
     }
 
     //Ownership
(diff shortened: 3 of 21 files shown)


media/js/devreg/devhub.js
```diff
@@ -1249,18 +1249,18 @@ var imageStatus = {
 function multipartUpload(form, onreadystatechange) {
     var xhr = new XMLHttpRequest(),
         boundary = "BoUnDaRyStRiNg",
-        form = $(form),
-        serialized = form.serializeArray(),
+        $form = $(form),
```
muffinresearch: nice

media/js/devreg/l10n.js
```diff
@@ -225,11 +225,12 @@ $(document).ready(function () {
         $("#change-locale").text(current.text());
         $(".trans").each(function () {
             var $el = $(this),
+                $ni = null,
```
muffinresearch: This is style more than a jshint thing but these should all be vars on separate lines.

media/js/devreg/slugify.js
```diff
@@ -27,9 +27,9 @@ function show_slug_edit(e) {
 }
 
 function slugify() {
-    if (z == null || z.unicode_letters) {
+    if (z === null || z.unicode_letters) {
         var slug = $('#id_slug');
-        url_customized = slug.attr('data-customized') == 0 ||
+        url_customized = slug.attr('data-customized') === 0 ||
```
muffinresearch: This might break. See this example for the background: http://jsbin.com/torivi/2/watch?html,js,console

I think we might want to swich to using `slug.data('customized') === 0` instead of attr.

media/js/devreg/validator.js
```diff
@@ -26,7 +26,7 @@ function initValidator($doc) {
 
     function ResultsTier($suite, tierId, options) {
         if (typeof options === 'undefined')
-            options = {}
+            options = {};
```
muffinresearch: You can kill all these ifs and replace with:

```js
options = options || {};
options.app = options.app || null;
options.testsWereRun = options.testsWereRun || true;
```

media/js/zamboni/files.js
```diff
@@ -15,7 +16,7 @@ if (typeof diff_match_patch !== 'undefined') {
                    the end. We can't just set lines.length - 1, because this
                    will just chop off lines. But if we don't trim these empty
                    lines we'll end up with lines between each diff. */
-                if ((t + 1) == lines.length && lines[t] == '') {
+                if ((t + 1) == lines.length && lines[t] === '') {
```
muffinresearch: What about the first '=='?

media/js/zamboni/files.js
```diff
@@ -283,7 +287,7 @@ function bind_viewer(nodes) {
                 html.push('</div>');
                 var $dom = $(html.join(''));
 
-                if (message.line != null && $line.length) {
+                if (message.line === null && $line.length) {
```
muffinresearch: Presumably you mean !==

media/js/zamboni/files.js
```diff
@@ -403,8 +407,8 @@ function bind_viewer(nodes) {
                         }
 
                         if (metadata.sub_packages) {
-                            for (var prefix in metadata.sub_packages) {
-                                process_files(prefix, metadata.sub_packages[prefix]);
+                            for (var sb_prefix in metadata.sub_packages) {
+                                process_files(prefix, metadata.sub_packages[sb_prefix]);
```
muffinresearch: What's the 'sb'? If it's sub maybe 'subPrefix'? 

Imho we really should try and stop doing underscores for js (apart from JSON keys)

media/js/zamboni/files.js
```diff
@@ -500,11 +504,12 @@ function bind_viewer(nodes) {
                     $diffbar = this.$diffbar,
                     gr = $gutter[0].getBoundingClientRect(),
                     gh = gr.bottom - gr.top,
-                    height = Math.max(0, Math.min(gr.bottom, $(window).height())
-                                       - Math.max(gr.top, 0));
+                    vp_height = Math.max(0,
```
muffinresearch: vpHeight

media/js/common/buckets.js
```diff
@@ -25,7 +25,7 @@ define('buckets', [], function() {
         // Camel-case it.
         property = property[0].toUpperCase() + property.substr(1);
 
-        for (var i = 0, e; e = prefixes[i++];) {
+        for (var i = 0; e; e = prefixes[i++]) {
```
muffinresearch: This doesn't look right to me. Running this gives a ReferenceError e is not defined.

I think we should remove the assignment. Here's why this sort of thing is bad. http://jsbin.com/macuya/1/watch?js,console

media/js/common/formsets.js
```diff
@@ -31,7 +31,7 @@
  *   'totalForms': '-TOTAL_FORMS',  // End of `name` attr of hidden field indicating total number of
  *                                  // forms (useful if you use multiple formsets in a single <form>)
  * })
- *  
+ *
  */
 
 (function($) {
```
muffinresearch: We could add a 'use strict' after this line. We'll need to watch those in-case there's other things that get broken by use of strict.

media/js/devreg/validator.js
```diff
@@ -274,14 +274,14 @@ function initValidator($doc) {
             if (!$('.msg', res).length)
                 $(res).hide();
         });
-        if (this.allCounts.error == 0 && this.allCounts.warning == 0) {
+        if (this.allCounts.error === 0 && this.allCounts.warning === 0) {
```
muffinresearch: Are we sure `this.allCounts.error` is going to be 0 and not "0"?

media/js/zamboni/files.js
```diff
@@ -1,4 +1,5 @@
-"use strict";
```
muffinresearch: This will make every file concatentated with this strict. We might not want that.

andymckay: Good job its being removed.

media/js/devreg/validator.js
andymckay: I have no idea on any of this because: no tests.

media/js/zamboni/files.js
andymckay: Thanks

media/js/devreg/validator.js
muffinresearch: Changing the comparison could break it if was relying on the coercion.

media/js/zamboni/files.js
muffinresearch: Sorry. my bad. Must be color blind!

media/js/devreg/l10n.js
andymckay: not keen on making changes beyond what we need for jshint, otherwise the number of changes becomes too big and scary

media/js/devreg/slugify.js
andymckay: Hmm interesting.

media/js/devreg/validator.js
andymckay: again, tempted not to make any changes beyond jshint

media/js/zamboni/files.js
andymckay: /me wonders how that passes jshint...

media/js/common/buckets.js
andymckay: @diox is suggesting to just delete this whole file, which goes against the "don't make more than jshint changes", but deleting code works. Got to figure out this file actually is for.

diox: I removed this file entirely in a different PR: #2932

andymckay:
The pull request is there, leaving it to someone braver.

PR #2909, opened 2015-02-12
Create mapping conditionally (bug 1132264)

muffinresearch:
Fixes the regression I caused :(

More evidence basic tests would be good to cover this sort of thing. I'm beginning to think something basic is worth doing unless refactoring the devhub is coming soon?


diox:
r+

I agree we should try doing basic tests now. Even if it means trashing them later when the refactor finally takes place (I have been hearing about this refactor since I joined Mozilla...)

PR #2913, opened 2015-02-12
Tidy up dev/device form processing and docs

robhudson:


docs/api/topics/search.rst
```diff
@@ -20,12 +20,12 @@ Search
...
-    :param optional device: Filters by supported device. One of 'desktop',
-        'mobile', 'tablet', or 'firefoxos'.
-    :type device: string
-    :param optional dev: Enables filtering by device profile if either
-                         'firefoxos' or 'android'.
+    :param optional dev: Filters by supported device. One of 'desktop',
+        'android', or 'firefoxos'.
     :type dev: string
+    :param optional device: Enables additional filtering by device profile
+    if device is 'android'. One of 'mobile', 'tablet', or 'firefoxos'.
```
diox: I would stop documenting 'firefoxos', as we only accept it because of backwards-compatibility.

mkt/search/forms.py
```diff
@@ -166,7 +166,7 @@ def clean_languages(self):
 
     def clean_device_and_dev(self):
         device = self.cleaned_data.get('dev')
-        device_type = self.cleaned_data.get('device')
```
diox: I'd also `pop()` this one (instead of doing that below)

diox:
r+wc

docs/api/topics/search.rst
robhudson: Oh? We don't accept it anymore or is there an alternative?

diox: You only need `dev` for Firefox OS - that's what fireplace sends. `device=firefoxos` pretty much does nothing, `device` is only used when `dev=android`.

robhudson: Ah, I'm with you now. Stop documenting but still accept it in the API. Ok. Will do.

robhudson:
https://github.com/mozilla/zamboni/commit/d823f37

PR #2914, opened 2015-02-13
Change tests to use email address rather than username where possible. (assist bug 1131717)

eviljeff:


diox:
r+ (change the PR name to reflect the commit name)

PR #2915, opened 2015-02-13
Handle native FxA logout in reviewer tools (bug 1132899)

washort:


media/js/devreg/login.js
```diff
@@ -160,20 +160,41 @@ define('login', ['notification', 'storage'], function(notification, storage) {
...
+    function init_native_fxa() {
         var user = z.body.data('user');
         var email = user ? user.email : '';
         console.log('detected user', email);
-        navigator.id.watch({
-            loggedInUser: email,
-            onlogin: function(assertion) {
-                if (!email) {
-                    gotVerifiedEmail(assertion);
+        var packaged_origin = "app://packaged." + window.location.host;
```
eviljeff: I don't believe any of the non-fireplace pages (e.g. dev hub; reviewer tools) run directly in the packaged app, but instead launch out into a browser instance.

media/js/devreg/login.js
```diff
@@ -160,20 +160,41 @@ define('login', ['notification', 'storage'], function(notification, storage) {
...
+        var isFirefoxOS = (navigator.mozApps && navigator.mozApps.installPackage &&
+            navigator.userAgent.indexOf('Android') === -1 &&
+            (navigator.userAgent.indexOf('Mobile') !== -1 ||
+             navigator.userAgent.indexOf('Tablet') !== -1));
+        var hasFxA = isFirefoxOS && navigator.userAgent.match(/rv:(\d{2})/)[1] >= 34;
+        var isYulelogFxA = window.top !== window.self && hasFxA;
+        if (isYulelogFxA) {
+            window.addEventListener('message', function (msg) {
+                if (!msg.data || !msg.data.type || msg.origin !== packaged_origin) {
```
diox: `packaged.` in the origin was only valid for dev and stage iframed package historically ; it has never been used in prod. 

The origins has since been cleaned up and we are only using `packaged.` for the true packaged app. The full list of origins can be found in https://github.com/mozilla/fireplace/blob/master/config.js#L59 : for each key, the yulelog origin is `iframe_origin`, the true packaged app origin (not used atm, except for testing) is `origin`.

We already have code in zamboni that allows origins ; it's in python, but then passed to the commonplace templates and then JS code, so maybe you could find a way to re-use it ? See https://github.com/mozilla/zamboni/blob/master/mkt/commonplace/views.py#L26

PR #2918, opened 2015-02-17
Remove extra columns from UserProfile model (bug 1131717)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1131717

BOOM

- [x] Rename `username` to `fxa_uid`
- [x] Drop `last_login_attempt`, `last_login_attempt_ip` and `failed_login_attempts`
- [x] Remove all code dealing with `username` except the stuff that sets `fxa_uid`
- [x] Remove dead authentication code
- [x] Fix tests and fixtures


mkt/account/tests/test_views.py
```diff
@@ -228,17 +227,15 @@ def test_put(self):
...
 
     def test_patch_extra_fields(self):
         res = self.client.patch(self.url,
                                 data=json.dumps({'display_name': 'foo',
-                                                 'username': 'bob'}))
+                                                 'name': 'bob'}))
         eq_(res.status_code, 200)
         user = UserProfile.objects.get(pk=self.user.pk)
         eq_(user.display_name, 'foo')  # Got changed successfully.
-        eq_(user.username, self.user.username)  # Did not change.
```
eviljeff: is there another non-user updatable field that can be tested instead?

mkt/site/tests/__init__.py
```diff
@@ -754,17 +754,15 @@ def req_factory_factory(url='', user=None, post=False, data=None, **kwargs):
...
+    If not provided, email will be 'factoryuser<number>@mozilla.com'.
     If email has no '@' it will be corrected to 'email@mozilla.com'
     """
     global user_factory_counter
-    username = kw.pop('username', '%030x' % random.randrange(16 ** 30))
     email = kw.pop('email', 'factoryuser%d' % user_factory_counter)
     if '@' not in email:
         email = '%s@mozilla.com' % email
 
-    user = UserProfile.objects.create(username=username, email=email, **kw)
```
eviljeff: should this function be generating a fxa id instead now?  (the 32 hex string was a generated version of it)

diox: I voted not to bother with it if we don't need to.

mkt/account/tests/test_views.py
diox: Yeah, I should update the test to use `fxa_uid` actually

mkt/users/models.py
```diff
@@ -147,7 +137,7 @@ def is_developer(self):
 
     @property
     def name(self):
-        return smart_unicode(self.display_name or self.username)
+        return smart_unicode(self.display_name or '-')
```
eviljeff: I do wonder if we should be using a default of the name part of name@domain in the email, or a pseudo random string.  (Or even something like 'user-%d' % self.id).  Otherwise the places where this is used might start to look a little odd if there are many users all with the same display name '-'.

diox: I like `'user-%d' % self.id` actually, that's a good suggestion. I didn't want to have the name part of the email in there, because of the potential privacy concern. But exposing the user id is fine, I'll do that.

mkt/zadmin/templates/admin/base.html
```diff
@@ -73,15 +73,6 @@
                       </li>
                     </ul>
                   </li>
-                {% else %}
```
eviljeff: Won't this mean there isn't a login link on admin pages?  (That wouldn't be the end of the world tbh)

diox: I tried loading `/admin/` without being logged on, and I ended up on a login page. I don't think this link even worked correctly.

eviljeff: Okay, the pages are wrapped in @require_admin decorators or similar I guess.  It would only be useful if there were any public /admin pages (e.g. status pages) but I think they're in /monitor

mkt/users/models.py
diox: Implemented.

mkt/account/tests/test_views.py
diox: Done. Added a couple more tests while I was there.

mkt/lookup/views.py
```diff
@@ -387,7 +387,7 @@ def _expand_query(q, fields):
 def user_search(request):
     results = []
     q = request.GET.get('q', u'').lower().strip()
-    search_fields = ('username', 'display_name', 'email')
+    search_fields = ('fxa_uid', 'display_name', 'email')
```
robhudson: Above you remove `display_name` from the template, should we remove/change it here as well? Is it used and worthy of querying?

mkt/settings.py
```diff
@@ -180,7 +180,6 @@
     'django.contrib.messages.middleware.MessageMiddleware',
     'mkt.api.middleware.APIBaseMiddleware',
     'mkt.api.middleware.AuthenticationMiddleware',
-    'commonware.log.ThreadRequestMiddleware',
```
robhudson: Tell me more?

mkt/users/models.py
```diff
@@ -63,24 +62,17 @@ def widget_attrs(self, widget):
                 'data-src': lazy_reverse('users.ajax')}
 
 
-AbstractBaseUser._meta.get_field('password').max_length = 255
```
robhudson: This would affect the SQL schema but there's nothing in the migration for this.

mkt/lookup/views.py
diox: Actually, I kept it in the template, just made the template use `name` which is a wrapper around `display_name` with a fallback added.

mkt/settings.py
diox: This middleware does 2 things:

- Sets the username of the user from the current request in thread local
- Sets the remote addr from the current request in thread local

It also exposes 2 functions to play with that data, `get_remote_addr()` and `get_username()`. We don't use the latter, and the former was only useful because we were trying to access the remote addr in models. I moved the call to the login view, which was the only place where the model method was used, and as a result we're free to simply use `request.META.get('REMOTE_ADDR')` and we don't need the methods or the middleware anymore.

diox: And actually, I just realized that's not true and `commonware.log` actually depends on the middleware to work :( I'll revert that part.

Worse, I'll need to expose a fake username property to keep the log stuff working :(

mkt/users/models.py
diox: We have no data anymore in this field, thanks to https://github.com/mozilla/zamboni/blob/master/migrations/839-cleanup-users.sql

You're right though.

mkt/account/views.py
```diff
@@ -256,7 +253,7 @@ def create_action(self, request, serializer):
                         msg='User created a new account (from FxA)')
                 record_action('new-user', request)
             auth.login(request, profile)
-            profile.log_login_attempt(True)
+            profile.update(last_login_ip=request.META.get('REMOTE_ADDR', ''))
```
eviljeff: I thought last_login_ip got dropped above?

diox: Nope, only `last_login_attempt_ip` (and `last_login_attempt`) was.

diox:
I submitted a PR to commonware that was merged, which fixes the dependency on username that it has, so I'll be able to put the middleware back, just need to wait on a release. Will try to come up with a test for that too.

eviljeff:
r+ from me... but my insight into FxA login flow is somewhat limited

diox:
@washort or @mstriemer mind taking a look ?

mstriemer:
We used the `last_login_attempt` field to detect if we wanted to send users a migration email for FxA. It looks like we no longer will have a way of detecting if the user has logged in (or tried, anyway) anymore. Do you know if there's any way to retrieve the last login time if we ever need it again?

diox:
last *successful* login fields are kept - `last_login` for the date, `last_login_ip` for, well, the ip.

mkt/users/models.py
diox: Migration added, reverting `password` to its default length, 128.

robhudson:
r+. Thanks for all the cleanup.

PR #2919, opened 2015-02-17
remove slug from webapp model (bug 1036900)

eviljeff:


mkt/webapps/models.py
```diff
@@ -499,14 +496,10 @@ def __unicode__(self):
...
 
     def save(self, **kw):
-        self.clean_slug(slug_field='app_slug')
+        self.clean_slug()
         creating = not self.id
         super(Webapp, self).save(**kw)
         if creating:
-            # Set the slug once we have an id to keep things in order.
-            # This breaks test_change_called_on_new_instance_save
-            self.update(slug='app-%s' % self.id)
```
diox: Glad to get rid of such a weird call (`.update()` inside `.save()` ??)

diox:
r+

PR #2927, opened 2015-02-19
Add file_size to fireplace serializer, display it in human-readable format (bug 1132138)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1132138

`file_size` was missing from fireplace serializer. I added it and cheated a little bit by changing the output to let the API to the formatting instead of doing it on the consumer pages side. Let me know what you think, @ngokevin.


ngokevin:
r+ thanks forgot to put it in the serializer

PR #2928, opened 2015-02-19
Fix key clash in search query rules (bug 998070)

robhudson:


robhudson:
The bug was to add a match query on name but it turned out we already had one but it wasn't showing up in the query b/c of the dictionary key clash, so this bug ended up being easy.

mkt/search/filters.py
```diff
@@ -34,18 +34,20 @@ def filter_queryset(self, request, queryset, view):
...
 
         # Only add fuzzy queries if q is a single word. It doesn't make sense
         # to do a fuzzy query for multi-word queries.
         if ' ' not in q:
-            rules[query.Fuzzy] = {'value': q, 'boost': 2, 'prefix_length': 1}
+            rules.append(
+                (query.Fuzzy, {'value': q, 'boost': 2, 'prefix_length': 1}))
 
-        for k, v in rules.iteritems():
+        for k, v in rules:
```
chuckharmston: Ha, good catch. I would have missed this.

mkt/search/filters.py
```diff
@@ -34,18 +34,20 @@ def filter_queryset(self, request, queryset, view):
...
-        rules = {
-            query.Match: {'query': q, 'boost': 3, 'analyzer': 'standard'},
-            query.Match: {'query': q, 'boost': 4, 'type': 'phrase', 'slop': 1},
-            query.Prefix: {'value': q, 'boost': 1.5}
-        }
+        rules = [
+            (query.Match, {'query': q, 'boost': 3, 'analyzer': 'standard'}),
+            (query.Match, {'query': q, 'boost': 4, 'type': 'phrase',
+                           'slop': 1}),
+            (query.Prefix, {'value': q, 'boost': 1.5}),
```
chuckharmston: I guess this is legit now that Bastacorp has filed for chapter 11.

chuckharmston:
r+

PR #2929, opened 2015-02-20
add in package id as well (bug 1123470)

andymckay:


mkt/lookup/views.py
```diff
@@ -139,6 +139,15 @@ def _transaction_summary(tx_uuid):
         log.warning('Transaction not found in solitude: {0}'.format(tx_uuid))
         lookup['transaction'] = False
 
+    if pay.get('provider') == PROVIDER_BANGO:
+        # If we are processing a Bango refund, then support would also like to
+        # know the package id.
+        try:
+            pay['package_id'] = (client.api.by_url(pay['seller'])
+                                 .get_object_or_404()['bango']['package_id'])
```
robhudson: That's some complicated nest of calls. What kind of object does `by_url` return?

robhudson:
r+

mkt/lookup/views.py
andymckay: A wrapper to an object in curling, its the same as `client.api.generic.seller(...)`, but saving the interpolation of the string `/generic/seller/.../`.

PR #2932, opened 2015-02-23
Remove device queue in reviewer tools (bug 1098333)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1098333

Feature detection does not work correctly and that queue does not match how the reviewers are doing their work anyway.


media/js/common/buckets.js
```diff
@@ -1,114 +0,0 @@
...
-    profile += '.' + capabilities.length;
-    // Add a version number.
-    profile += '.4';
-
-    return {
-        get_profile: function() {return profile;},
-        capabilities: capabilities
-    };
-
-});
```
robhudson: Remove this from `mkt/asset_bundles.py` also.

robhudson:
r+wc

media/js/common/buckets.js
diox: Good catch, surprised this didn't cause a test failure. I'll look into that.

PR #2933, opened 2015-02-23
use transaction.atomic instead of old transaction.TransactionMiddleware

eviljeff:


mkt/settings.py
```diff
@@ -94,6 +94,7 @@
 DATABASES['default']['OPTIONS'] = {'init_command': 'SET storage_engine=InnoDB'}
 DATABASES['default']['TEST_CHARSET'] = 'utf8'
 DATABASES['default']['TEST_COLLATION'] = 'utf8_general_ci'
+DATABASES['default']['ATOMIC_REQUEST'] = 'True'
```
eviljeff: Do I need to add this to the settings.py files in /sites/ also?

mkt/translations/tests/test_forms.py
```diff
@@ -13,6 +13,7 @@ class TestForm(forms.TranslationFormMixin, ModelForm):
 
     class Meta:
         model = TranslatedModel
+        exclude = []
```
eviljeff: Not strictly anything to do with the rest of this patch... but harmless.

eviljeff:
transaction stuff in `app_review` worries me a little still.

mkt/reviewers/views.py
```diff
@@ -419,25 +419,21 @@ def _review(request, addon, version):
     return render(request, 'reviewers/review.html', ctx)
 
 
-@transaction.commit_manually
```
diox: While you are at it, can you remove the following in `mkt/reviewers/tests/test_views.py`:
```
if os.environ.get('JENKINS_HOME'):
    raise SkipTest()
```

mkt/reviewers/views.py
```diff
@@ -448,17 +444,12 @@ def app_review(request, addon):
         # them to index the app or call update_version() if that wasn't done
         # before already.
         if request.method == 'POST':
-            try:
+            with transaction.atomic():
```
diox: you can remove this. The entire view will be under transaction management anyway, and will be rolled back if those queries fail

diox: (Also, we should file a bug to replace this code, the post_request tasks system we have now make this obsolete)

mkt/settings.py
diox: Yes, since they defined their own `DATABASES` dict, overriding this one.

mkt/reviewers/views.py
eviljeff: Is this bug 1105507?

eviljeff: Thinking about this method more, do we *want* the whole transaction to be rolled back if there is an exception in the post_save handlers?  At that point (I think) we've already signed the zip, send the emails, etc.

diox: That's not 1105507, no. Without moving anything to celery, this bit of code is obsolete and could probably be replaced by something doing signals normally in the code being executed before.

mkt/settings.py
eviljeff: okay, added.

mkt/reviewers/views.py
```diff
@@ -419,25 +419,21 @@ def _review(request, addon, version):
...
-        resp = _review(request, addon, version)
+        with transaction.atomic():
+            resp = _review(request, addon, version)
     except SigningError, exc:
-        transaction.rollback()
         messages.error(request, 'Signing Error: %s' % exc)
-        transaction.commit()
         return redirect(
             reverse('reviewers.apps.review', args=[addon.app_slug]))
     except Exception:
```
diox: Remove this `except` and `raise` as it's now superfluous

mkt/settings.py
diox: It should be a boolean, not a string.

eviljeff:
changes made btw.

diox:
I want to test this more but it looks ok. Wait till Monday to merge it so that we can look at how dev behaves with it.

mkt/settings.py
```diff
@@ -94,6 +94,7 @@
 DATABASES['default']['OPTIONS'] = {'init_command': 'SET storage_engine=InnoDB'}
 DATABASES['default']['TEST_CHARSET'] = 'utf8'
 DATABASES['default']['TEST_COLLATION'] = 'utf8_general_ci'
+DATABASES['default']['ATOMIC_REQUEST'] = True
```
diox: There and in all settings, it should be `ATOMIC_REQUEST*S*`.

Which means we need a test proving that the views are correctly wrapped into a transaction, since tests were passing despite this typo.

eviljeff:
tests pass now, with the bare minimum of fudging so everything still works.

diox:
r+, merge on Monday.

PR #2934, opened 2015-02-23
Update some outdated requirements around docs/tests/dev (bug 846349)

diox:

### DIFF
--- a/mkt/api/forms.py
+++ b/mkt/api/forms.py
@@ -153,7 +153,10 @@ class SchemeURLValidator(URLValidator):
 
     def __call__(self, value):
         super(URLValidator, self).__call__(value)
-        supports_scheme = lambda scheme: value.startswith(scheme + '://')
+
+        def supports_scheme(scheme):
+            return value.startswith('{0}://'.format(scheme))
+
         if not any(supports_scheme(scheme) for scheme in self.schemes):
             raise ValidationError('Scheme should be one of {0}.'.format(
                                   ', '.join(self.schemes)))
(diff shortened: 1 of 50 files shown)


mkt/api/forms.py
```diff
@@ -153,7 +153,10 @@ def __init__(self, *args, **kwargs):
 
     def __call__(self, value):
         super(URLValidator, self).__call__(value)
-        supports_scheme = lambda scheme: value.startswith(scheme + '://')
+
+        def supports_scheme(scheme):
+            return value.startswith(scheme + '://')
```
robhudson: I always like `'%s://' % scheme` better

robhudson: Or even `'{0}://'.format(scheme)` for future compat

requirements/test.txt
```diff
@@ -7,11 +7,11 @@
...
 flake8==2.3.0
 pyflakes==0.8.1
-pep8==1.5.7
+pep8==1.6.2
 mccabe==0.3
 
 fudge==1.0.3
 mock==1.0.1
-psutil==0.2.0
+psutil==2.2.1
```
robhudson: Is this used anywhere? I'm not seeing it imported but perhaps it's a dependency for something else. If so, should we document when a requirement is added but only as a dependency?

robhudson:
r+wc

requirements/test.txt
diox: We should document that, but it's a pretty big task...

This particular package does not seem to be used anywhere. I ran a reverse dependency check and could not find a package requiring it. It was added in fe6836bc2ec68cc09fc5c2da63fca3596e13b047 (thanks, `git bisect`!) for qunit tests, which we no longer do. I'll remove it!

robhudson: Yay, requirements--.

PR #2937, opened 2015-02-24
Drastically reduce the number of SQL queries made by the ratings API (bug 1135979)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1135979

### DIFF
--- a/mkt/ratings/serializers.py
+++ b/mkt/ratings/serializers.py
@@ -3,7 +3,7 @@ from django.core.urlresolvers import reverse
 from rest_framework import serializers
 from rest_framework.exceptions import PermissionDenied
 
-from mkt.account.serializers import AccountSerializer
+from mkt.account.serializers import UserSerializer
 from mkt.api.fields import SlugOrPrimaryKeyRelatedField, SplitField
 from mkt.api.exceptions import Conflict
 from mkt.ratings.models import Review, ReviewFlag
@@ -20,7 +20,7 @@ class RatingSerializer(serializers.ModelSerializer):
         serializers.HyperlinkedRelatedField(view_name='app-detail',
                                             read_only=True, source='addon'))
     body = serializers.CharField()
-    user = AccountSerializer(read_only=True)
+    user = UserSerializer(read_only=True)
     report_spam = serializers.SerializerMethodField('get_report_spam_link')
     resource_uri = serializers.HyperlinkedIdentityField(
         view_name='ratings-detail')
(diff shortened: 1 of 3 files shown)


mkt/ratings/views.py
```diff
@@ -27,25 +26,16 @@
 log = commonware.log.getLogger('z.api')
 
 
-class RatingPaginator(Paginator):
-    # FIXME: This is only right when ?app= filtering is applied, if no
-    # filtering or if ?user= filtering is done, it's completely wrong.
-    @property
-    def count(self):
```
diox: `count` is used 4 times on the serializer. (The default paginator uses `count` to calculate the results once). This means that 4 times, we loaded the whole object list, and 4 times, we loaded the related `addon` instance. Because the our models have transformers loading extra stuff, that's roughly 25 extra queries *per review* in the list.

mkt/ratings/serializers.py
```diff
@@ -49,6 +52,15 @@ def __init__(self, *args, **kwargs):
             # Don't let users modify 'app' field at edit time
             self.fields['app'].read_only = True
 
+    def to_native(self, obj):
+        # When we have an `app` set on the serializer, we know it's because the
+        # view was filtering on this app, so we can safely overwrite the
+        # `addon` property on the instance with it, saving some costly queries.
+        app = getattr(self, 'app', None)
+        if app is not None:
+            obj.addon = app
```
diox: Saves us ~ 16 queries per review.

mkt/ratings/views.py
```diff
@@ -27,25 +26,18 @@
...
-
-
 class RatingViewSet(CORSMixin, MarketplaceView, ModelViewSet):
     # Unfortunately, the model class name for ratings is "Review".
-    queryset = Review.objects.valid().all()
+    # We prefetch 'version' because it's often going to be similar, and select
+    # related 'user' to avoid extra queries.
+    queryset = (Review.objects.valid()
+                .prefetch_related('version')
+                .select_related('user'))
```
diox: `.prefetch_related('version')` saves us 2 queries per review, assuming the same version is going to appear multiple times.
`.select_related('user')` does a `JOIN` and saves us one query per review.

diox:
We're saving a lot of queries there, this should improve performance quite a bit. We still need to check indexes, but that can be done separately.

I still need to test whether cache-machine will like the `prefetch_related`/`select_related` or not.

mkt/ratings/tests/test_views.py
```diff
@@ -65,38 +88,81 @@ def test_get_empty_with_app(self):
...
+                              version=first_version,
+                              body=u'I also lôve this app',
+                              rating=4)
+        self.app.total_reviews = 2
+        version_factory(addon=self.app, version='2.0',
+                        file_kw=dict(status=mkt.STATUS_PUBLIC))
+        self.app.update_version()
+
+        reset_queries()
+        with self.assertNumQueries(5):
```
diox: When I started, this test did more than a hundred queries.

robhudson:
I read it over a few times. r+

diox:
Checked with cache-machine enabled and it seems to be doing the right thing.

PR #2938, opened 2015-02-24
Bug 967921 - User message is not true

gurumukhi:
User message 'Changing to Free will put your app back into review.' is not true. Removed the message.


robhudson:
Thanks for the patch!

PR #2942, opened 2015-02-25
Try compressing assets before running tests (bug 1136655)

diox:


mkt/asset_bundles.py
```diff
@@ -221,6 +221,7 @@
     ),
     'mkt/reviewers': (
         'js/lib/moment-with-langs.min.js',  # JS date lib.
+        'js/common/buckets.js',
```
diox: This is temporary, to test that the build does fail if `compress_assets` returns an error.

eviljeff:
r+

PR #2944, opened 2015-02-26
Adjust permissions on mini-manifest view (bug 1132137)

robhudson:
This also refactors the mini manifest view to use DRF to re-use the
permission classes.


robhudson:
There were a couple tests failing for me here and I couldn't quite figure out why. When I checked `request.user` it was always `AnonymousUser` even thought this is using `self.client` to make the API call.

I'm hoping for general comments about making the manifest view use DRF and also if anyone has ideas about why these tests are not getting the right user.

mkt/detail/tests/test_views.py
```diff
@@ -120,20 +114,20 @@ def test_app_unlisted_author(self, _mock):
     def test_app_private(self):
         self.app.update(status=mkt.STATUS_APPROVED)
         res = self.client.get(self.url)
-        eq_(res.status_code, 404)
+        eq_(res.status_code, 403)
```
robhudson: I'm not sure if I like that a `403` could give a hint that the manifest exists but the user doesn't have access to it, but really, this is the manifest URL with the UUID. We should be more concerned about this on the app detail API but we're also returning 403s there so this probably doesn't matter much. It is more correct.

mkt/detail/tests/test_views.py
```diff
@@ -104,7 +98,7 @@ def test_app_unlisted(self, _mock):
     @mock.patch('mkt.webapps.models.Webapp.get_cached_manifest')
     def test_app_unlisted_reviewer(self, _mock):
         _mock.return_value = self._mocked_json()
-        self.login_as_reviewer()
+        self.grant_permission(self.profile, 'Apps:Review')
```
robhudson: I need to make this `Apps:Staff`.

mkt/api/authorization.py
```diff
@@ -135,6 +136,18 @@ def has_object_permission(self, request, view, object):
         return object.is_public() and self.has_permission(request, view)
 
 
+class AllowStaffReadOnlyIfReviewed(BasePermission):
+    """
+    Allow users in the Staff group to view apps that have any published status.
+    """
+    def has_permission(self, request, view):
+        return (request.method in SAFE_METHODS and
+                acl.action_allowed(request, 'Apps', '*'))
```
eviljeff: isn't it `acl.action_allowed(request, 'Apps', '%'))` ?

mkt/api/authorization.py
```diff
@@ -177,6 +165,40 @@ def __call__(self, *a):
...
+    action = '%'
+
+    def has_permission(self, request, view):
+        return (request.method in SAFE_METHODS and
+                super(AllowStaffReadOnlyIfReviewed,
+                      self).has_permission(request, view))
+
+    def has_object_permission(self, request, view, object):
+        return (self.has_permission(request, view) and
+                object.status in mkt.REVIEWED_STATUSES)
```
eviljeff: Will this make the detail page completely inaccessible once an app has been rejected?  Or just the mini-manifest?

mkt/api/authorization.py
```diff
@@ -177,6 +165,40 @@ def __call__(self, *a):
...
+                super(AllowReviewerReadOnly, self).has_permission(request,
+                                                                  view))
+
+
+class AllowStaffReadOnlyIfReviewed(GroupPermissionBase):
+    """
+    Allow users in the Staff group to view apps that have any published status.
+    """
+    app = 'Apps'
+    action = '%'
```
eviljeff: iirc `"Apps:%"`  is shorthand for *any* `Apps:...` permission, of which many of the holders won't be staff (reviewers, moderators, etc). 'Apps:Edit' is probably the closest 'staff' permission.

robhudson: The Staff group has `Apps:*`. Reviewers have `Apps:Review`. Support staff has `Apps:Configure`. You're saying this is too broad of a permission?

mkt/api/authorization.py
robhudson: This PR updates the app detail api view to also use this so yes, it would be 403 if rejected. Which is a change to previously allowing reviewers always.

robhudson:
Closing for reasons cited in the bug. I'll re-work the permissions and drop the DRF stuff done here. (I may copy over some of these tests and permission changes, tho)

mkt/api/authorization.py
eviljeff: From the name of the class I thought the intention was to have a more restrictive than AllowReviewerReadOnly group, where it was only Staff.  If it was supposed to be really inclusive then its fine (I didn't look to see where it was used)

mkt/api/authorization.py
eviljeff: At the very least Lisa and I should have the same level as access as the developer.

PR #2947, opened 2015-02-26
Fix fixture json.

jasonthomas:
This fixes error 
'django.core.serializers.base.DeserializationError: Problem installing fixture 'zamboni/mkt/site/fixtures/init.json': Expecting property name enclosed in double quotes: line 1703 column 9 (char 45632)' when running python manage.py loaddata ini.

r?


andymckay:
r+

PR #2950, opened 2015-02-27
Remove simplejson now that we are using "json" everywhere (bug 915256)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=915256


ngokevin:
r+

PR #2952, opened 2015-03-02
Fix JS error on paid app submission (bug 1137913)

muffinresearch:
This also fixes another related JS error when the submission is made and there's no platforms selected.


diox:
r+

PR #2958, opened 2015-03-03
Add a search view without region filtering for curation tools (bug 1136460)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1136460#c6

Also remove trailing garbage after search API urls, to end possible confusion between `/api/v1/apps/search/suggest` and `/api/v1/apps/search/suggest/`. The former is no longer tolerated (was never documented, only used by accident in transonic)


diox:
Note: when merging, **transonic** needs to be updated at the same time to use the new API.

mkt/search/views.py
```diff
@@ -101,6 +102,20 @@ class NonPublicSearchView(SearchView):
...
+    """
+    A search view that allows searching for public apps regardless of region
+    exclusions, protected behind a permission class.
+
+    """
+    authentication_classes = [RestSharedSecretAuthentication,
+                              RestOAuthAuthentication]
+    permission_classes = [GroupPermission('Feed', 'Curate')]
+    filter_backends = [SearchQueryFilter, PublicSearchFormFilter,
+                       PublicAppsFilter, DeviceTypeFilter,
```
robhudson: I *think* they also didn't want device filtering. I would imagine profile filtering doesn't make sense here also for feed curation search?

mkt/search/tests/test_views.py
```diff
@@ -1050,6 +1050,81 @@ def test_not_finds_invalid_statuses(self):
...
             eq_(len(res.json['objects']), 0)
 
+    def test_not_finds_excluded_region(self):
+        self.app2.addonexcludedregion.create(region=mkt.regions.USA.id)
+        self.reindex(Webapp, 'webapp')
+        res = self.client.get(self.url, data={
+            'q': 'second', 'lang': 'en-US', 'region': 'us'
+        })
+        eq_(res.status_code, 200)
+        eq_(len(res.json['objects']), 0)
```
robhudson: :+1:

mkt/search/tests/test_views.py
```diff
@@ -1050,6 +1050,81 @@ def test_not_finds_invalid_statuses(self):
...
+        eq_(objs[0]['name'], self.app2.name)
+
+    def test_finds_public(self):
+        res = self.client.get(self.url, data={
+            'q': 'second', 'lang': 'en-US', 'region': 'us'
+        })
+        eq_(res.status_code, 200)
+        objs = res.json['objects']
+        eq_(len(objs), 1)
+        eq_(objs[0]['name'], self.app2.name)
```
robhudson: This is the exact same as `test_with_permission`. You could probably just get rid of that one.

mkt/api/v2/urls.py
```diff
@@ -48,10 +49,13 @@
...
         name='apps-recommend'),
-    url(r'^apps/search/rocketbar/', RocketbarViewV2.as_view(),
+    url(r'^apps/search/rocketbar/$', RocketbarViewV2.as_view(),
         name='rocketbar-search-api'),
-    url(r'^apps/search/non-public/', NonPublicSearchView.as_view(),
+    url(r'^apps/search/non-public/$', NonPublicSearchView.as_view(),
         name='non-public-search-api'),
+    url(r'^apps/search/no-region/$',
+        NoRegionSearchView.as_view(),
+        name='no-region-search-api'),
```
robhudson: Thanks for cleaning up the URLs.

mkt/search/views.py
ngokevin: Yeah, I think the only thing we want filtered is for Public Apps. That might mean this endpoint might need to be renamed.

diox: The client is not sending the device or profile parameter though, so do we care ?

ngokevin: I guess not. It'd be up to the client to decide if they want it or not

mkt/search/views.py
```diff
@@ -101,6 +102,20 @@ class NonPublicSearchView(SearchView):
                        ProfileFilter, SortingFilter]
 
 
+class NoRegionSearchView(SearchView):
+    """
+    A search view that allows searching for public apps regardless of region
+    exclusions, protected behind a permission class.
+
+    """
+    authentication_classes = [RestSharedSecretAuthentication,
```
ngokevin: this is inherited from SearchView right? or do we just want to be explicit

ngokevin:
lgtm

mkt/search/views.py
robhudson: In that case, why not update curation tools to not send the region? Or is that trickier b/c then it would do geo-ip lookups?

diox: mmm, good point. we don't do /automatic/ geoip lookups in v2... I'll check that out.

diox: So what happens is, we disable geoip lookup in v2, but we default to `restofworld`. If enabled, the `RegionFilter` picks up that region (even if it's not present in the query string - giving us compatibility with the automatic geoip lookups v1 does), and that's why we need a separate endpoint without that filter.

mkt/search/views.py
diox: Just being explicit because the other class nearby does the same :)

mkt/search/views.py
diox: I added a comment explaining why we have a specific view.

mkt/search/tests/test_views.py
diox: I removed this one and kept `test_with_permission`.

diox:
Since we changed our minds a couple times concerning what kind of filtering is done where, I preferred to follow what @robhudson and @chuckharmston did for operator tools (the search view operator tools is `NonPublicSearchView` - it *does* filter on region, but shows non-public apps as well).

PR #2959, opened 2015-03-03
Revert use table cleanup (bug 1131717)

kumar303:
This migration is too slow for prod right now

WIP, DO NOT MERGE

### DIFF
--- a/migrations/897-remove-username-and-extra-fields.sql
+++ b/migrations/897-remove-username-and-extra-fields.sql
@@ -1,29 +1,36 @@
+-- This migration was too slow to run on prod as is.
+-- In addition to this commit we're going to manually run this SQL on
+-- stage to test the removal:
+-- https://gist.github.com/kumar303/aff8caedf59da3459bdb
+
+
+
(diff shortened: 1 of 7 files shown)


clouserw:
Thanks for space change.  The revert part (the new .sql file) shouldn't be a part of this patch.  That's just something jason runs on stage manually - that way it won't mess anyone else up and when we test the real fix later it will work as expected on stage.

kumar303:
Well, we need a way to see what code will break too when the database migration is not run so this patch will run the suite hopefully in a manner similar to prod (in the case where the removal migration is disabled)

clouserw:
r+

kumar303:
This was not merged to master because it will break local installs for each developer. Instead I pushed it to this branch: https://github.com/mozilla/zamboni/compare/master...revert-1131717 It has also been cherry picked into the release tag: https://github.com/mozilla/zamboni/compare/2015.02.24...2015.03.03

kumar303:
The user table cleanup will be reintroduced in https://bugzilla.mozilla.org/show_bug.cgi?id=1139188

PR #2961, opened 2015-03-03
Generate fake data for feed (bug 1131903)

washort:
This should generate the same sort of data as the feed support in https://github.com/mozilla/marketplace-api-mock/

### DIFF
--- a/mkt/feed/management/commands/generate_feed.py
+++ b/mkt/feed/management/commands/generate_feed.py
@@ -1,0 +1,17 @@
+from django.core.management.base import BaseCommand
+
+from mkt.feed.fakedata import generate_feed_data
+
+
+class Command(BaseCommand):
+    """
+    Usage:
+
+        python manage.py generate_feed
+
+    """
+
+    help = 'Generate example feed data for development/testing'
+
+    def handle(self, *args, **kwargs):
+        generate_feed_data()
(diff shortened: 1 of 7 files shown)


mkt/feed/management/commands/generate_feed.py
```diff
@@ -0,0 +1,17 @@
...
+    Usage:
+
+        python manage.py generate_feed
+
+    """
+
+    help = 'Generate example feed data for development/testing'
+
+    def handle(self, *args, **kwargs):
+        generate_feed_data()
```
robhudson: Should we add an indexing call here? Or does the model saving trigger indexing via signals?

robhudson:
I glanced briefly, looks fine. I think someone like @ngokevin should look since he's more familiar with feed models.

mkt/feed/management/commands/generate_feed.py
washort: Good question, I was running reindex each time due to the ES problems we discussed earlier. I believe it should index as it goes but I'll check.

ngokevin:
It would be cool if we could share some modules from Marketplace API Mock, but I guess the intention is to get rid of the API Mock. Although the code looks a bit cleaner in the API Mock.

Anyhow, looks fine. Looks like you followed https://github.com/mozilla/marketplace-api-mock/blob/master/factory/feed.py#L217 pretty well.

PR #2968, opened 2015-03-06
Reviewer onboarding tool (bug 1101096)

washort:
Second try, this time using the existing fake-data generation tools, extended to cover more fields.


washort:
screen shot 2015-03-06 at 3 54 08 pm

data/apps/reviewer_apps.json
```diff
@@ -0,0 +1,26 @@
+[{"type": "hosted",
+  "status": "pending",
+  "manifest_url": "http://reviewer-app1.testmanifest.com/manifest.webapp",
+  "categories": ["books", "business"],
+  "device_types": ["firefoxos", "desktop"],
+  "privacy_policy": "good privacy policy",
+  "preview_files": ["images/screenshot1.png"],
+  "manifest_file": "manifests/app1.json"
```
eviljeff: would the json in manifest_file be the same as the one on testmanifest.com?

washort: Yes; I just wanted to set this up to load from a file so we're not dependent on an external resource. Besides, it's easier to customize this way.

mkt/webapps/fakedata.py
```diff
@@ -247,26 +272,47 @@ def generate_apps(hosted=0, packaged=0, privileged=0, versions=(4,)):
...
-        app = generate_hosted_app(appname, cat_slug, developer_name)
+        app = generate_hosted_app(name, categories, developer_name,
+                                  status=status, **spec)
     else:
         app = generate_packaged_app(
-            appname, type, cat_slug, developer_name, **spec)
+            name, type, categories, developer_name,
+            status=status, **spec)
     generate_icon(app)
     generate_previews(app, num_previews)
```
eviljeff: redundant if preview_files specified?

eviljeff:
I'm not super familiar with the fake apps code but it seems to do what we'd need it to.  So r+wc from me.

PR #2969, opened 2015-03-10
fix commbage emails for mozilla_contacts (bug 1132010)

eviljeff:


ngokevin:
r+ thanks. Let's cherry pick this today

PR #2970, opened 2015-03-10
Handle free in-app products (bug 1138609)

muffinresearch:
* [x] Needs tests.

### DIFF
--- a/mkt/purchase/tests/test_webpay.py
+++ b/mkt/purchase/tests/test_webpay.py
@@ -274,6 +274,33 @@ class TestPostback(PostbackTest):
 
         assert not self.tasks.send_purchase_receipt.delay.called
 
+    def test_free_inapp(self):
+        response_trans_id = '<free-uuid>'
+        (self.solitude.api.generic.buyer.get_object_or_404
+                                        .return_value) = {
+            'email': self.buyer_email,
+        }
+        inapp = InAppProduct.objects.create(
+            logo_url='logo.png', name=u'Free Inapp Product',
+            price=Price.objects.get(price=0),
+            webapp=self.addon)
+        self.contrib.update(inapp_product=inapp, addon=inapp.webapp)
+        jwt_dict = self.jwt_dict()
+        jwt_dict['response']['transactionID'] = response_trans_id
+        jwt_dict['response']['solitude_buyer_uuid'] = '<buyer:uuid>'
+        jwt_dict['request']['pricePoint'] = '0'
+        jwt_encoded = self.jwt(req=jwt_dict)
+        self.decode.return_value = jwt_dict
+        resp = self.post(req=jwt_encoded)
+        (self.solitude.api.generic.buyer
+             .get_object_or_404.assert_called_with)(uuid='<buyer:uuid>')
+        eq_(resp.status_code, 200)
+        eq_(resp.content, response_trans_id)
+        cn = Contribution.objects.get(pk=self.contrib.pk)
+        eq_(cn.type, mkt.CONTRIB_PURCHASE)
+        eq_(cn.user.email, self.buyer_email)
+        self.tasks.send_purchase_receipt.delay.assert_called_with(cn.pk)
+
     def test_user_created_after_purchase(self):
         self.contrib.user = None
         self.contrib.save()
--- a/mkt/purchase/webpay.py
+++ b/mkt/purchase/webpay.py
@@ -87,6 +87,26 @@ def pay_status(request, addon, contrib_uuid):
     return {'status': 'complete' if qs.exists() else 'incomplete'}
 
 
+def _get_user_profile(request, buyer_email):
+    user_profile = UserProfile.objects.filter(email=buyer_email)
+
+    if user_profile.exists():
+        user_profile = user_profile.get()
+    else:
+        source = mkt.LOGIN_SOURCE_WEBPAY
+        user_profile = UserProfile.objects.create(
+            email=buyer_email,
+            is_verified=True,
+            source=source)
+
+        log_cef('New Account', 5, request, username=buyer_email,
+                signature='AUTHNOTICE',
+                msg='A new account was created from Webpay (using FxA)')
+        record_action('new-user', request)
+
+    return user_profile
+
+
 @csrf_exempt
 @write
 @require_POST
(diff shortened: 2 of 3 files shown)


mkt/prices/models.py
```diff
@@ -347,7 +347,13 @@ def create_addon_purchase(sender, instance, **kw):
...
         return
 
-    if instance.type == mkt.CONTRIB_PURCHASE:
+    print
+    print dir(instance.price_tier)
+    print instance.price_tier.price
+    print instance.price_tier.name
+
+    if instance.type == mkt.CONTRIB_PURCHASE \
+            and not instance.price_tier.name == '0':
```
kumar303: Hmm, I wonder if we should not save the AddonPurchase record for all in-app purchases.

I'm struggling to find out where we actually store the data that a user purchased an in-app product. We need to store that, especially for free products so they can be restored.

kumar303: actually, I think we rely on Contribution to store the record of a user purchasing an in-app product. Thus, you could skip AddonPurchase for all in-app purchases (maybe). But that's a separate bug.

mkt/prices/models.py
```diff
@@ -374,7 +380,8 @@ def create_addon_purchase(sender, instance, **kw):
                       % (p.pk, instance.addon.pk, instance.user.pk))
             p.update(type=instance.type)
 
-    cache.delete(memoize_key('users:purchase-ids', instance.user.pk))
+    # FIXME: Don't leave this commented.
+    # cache.delete(memoize_key('users:purchase-ids', instance.user.pk))
```
kumar303: ha

mkt/prices/models.py
kumar303: Also, does this break free purchases of *apps* ? We want those to get saved to AddonPurchase

mkt/prices/models.py
```diff
@@ -347,7 +347,13 @@ def create_addon_purchase(sender, instance, **kw):
                      c=instance))
         return
 
-    if instance.type == mkt.CONTRIB_PURCHASE:
+    print
+    print dir(instance.price_tier)
+    print instance.price_tier.price
+    print instance.price_tier.name
+
+    if instance.type == mkt.CONTRIB_PURCHASE \
```
muffinresearch: This bit is wrong at the moment as this should be run. The problem is the user profile bit isn't working as the user object here is not right.

mkt/prices/models.py
muffinresearch: Yeah this is wrong at the moment. It's busted at this point as the user object is missing. I'm debugging that now. I'm expecting this code to run (and the added conditional will be removed).

mkt/webpay/urls.py
```diff
@@ -22,7 +22,8 @@
         name='webpay-prepare'),
     url(r'^webpay/inapp/prepare/', PreparePayInAppView.as_view(),
         name='webpay-prepare-inapp'),
-    url(r'^webpay/failure/(?P<pk>\d+)/', FailureNotificationView.as_view(),
+    url(r'^webpay/failure/(?P<pk>(?:simulate|free|webpay)?:?[a-z0-9-]+)/',
```
muffinresearch: This can be simplified to be more like the other patterns that match on everything that isn't a '/'

mkt/purchase/tests/test_webpay.py
```diff
@@ -274,6 +274,33 @@ def test_simulation(self):
...
 
+    def test_free_inapp(self):
+        response_trans_id = '<free-uuid>'
+        (self.solitude.api.generic.buyer.get_object_or_404
+                                        .return_value) = {
+            'email': self.buyer_email,
+        }
+        inapp = InAppProduct.objects.create(
+            logo_url='logo.png', name=u'Free Inapp Product',
+            price=Price.objects.filter(price=0)[0],
```
kumar303: Could be ``Price.objects.get(price=0)``

mkt/purchase/tests/test_webpay.py
```diff
@@ -274,6 +274,33 @@ def test_simulation(self):
...
+        jwt_dict['request']['pricePoint'] = '0'
+        jwt_encoded = self.jwt(req=jwt_dict)
+        self.decode.return_value = jwt_dict
+        resp = self.post(req=jwt_encoded)
+        (self.solitude.api.generic.buyer
+             .get_object_or_404.assert_called_with)(uuid='<buyer:uuid>')
+        eq_(resp.status_code, 200)
+        eq_(resp.content, response_trans_id)
+        cn = Contribution.objects.get(pk=self.contrib.pk)
+        eq_(cn.type, mkt.CONTRIB_PURCHASE)
```
kumar303: This should also assert that the contribution gets attributed to the right user.

mkt/purchase/webpay.py
```diff
@@ -206,6 +228,20 @@ def simulated_postback(contrib, trans_id):
     return http.HttpResponse(trans_id)
 
 
+def free_postback(request, contrib, trans_id, user_profile):
+    log.info('Got free product postback: fulfilling pruchase for contrib={c}; '
+             'trans={t}; user={u}'.format(c=contrib,
+                                          t=trans_id, u=user_profile))
```
kumar303: Since you have this log message here, you can probably remove the same one from above: https://github.com/mozilla/zamboni/pull/2970/files#diff-bff80a5504e028c0f04952576a512f85R175

mkt/purchase/webpay.py
```diff
@@ -206,6 +228,20 @@ def simulated_postback(contrib, trans_id):
     return http.HttpResponse(trans_id)
 
 
+def free_postback(request, contrib, trans_id, user_profile):
+    log.info('Got free product postback: fulfilling pruchase for contrib={c}; '
```
kumar303: these strings should be unicode literals like ``u'Got free...'`` because the stringified user_profile object might contain non-ascii chars (I think)

kumar303:
Another simple patch! I like it.

r+wc

mkt/purchase/webpay.py
```diff
@@ -206,6 +226,20 @@ def simulated_postback(contrib, trans_id):
     return http.HttpResponse(trans_id)
 
 
+def free_postback(request, contrib, trans_id, user_profile):
+    log.info(u'Got free product postback: fulfilling purchase for '
+             'contrib={c}; trans={t}; user={u}'.format(
```
kumar303: annoying, I know, but all concatenated strings need to be unicode literals. This line needs to start with ``u'contrib...``

mkt/purchase/webpay.py
```diff
@@ -194,7 +214,7 @@ def postback(request):
 
 def simulated_postback(contrib, trans_id):
     simulate = contrib.inapp_product.simulate_data()
-    log.info('Got simulated payment postback; contrib={c}; '
+    log.info(u'Got simulated payment postback; contrib={c}; '
              'trans={t}; simulate={s}'.format(c=contrib, t=trans_id,
```
kumar303: this line need to start with ``u'trans...``

mkt/purchase/webpay.py
```diff
@@ -139,6 +159,20 @@ def postback(request):
                         contrib_uuid=contrib_uuid,
                         contrib_trans=contrib.transaction_id))
 
+    # Special-case free apps.
```
kumar303: s/apps/products/

mkt/purchase/tests/test_webpay.py
kumar303: oh I see you added it

mkt/purchase/tests/test_webpay.py
```diff
@@ -274,6 +274,34 @@ def test_simulation(self):
...
+        user = UserProfile.objects.get(email=self.buyer_email)
+        eq_(cn.user, user)
```
kumar303: could just be ``eq_(cn.user.email, self.buyer_email)``

kumar303:
r+wc

PR #2973, opened 2015-03-11
Fix device filter form in reviewer tools (bug 1107924)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1107924

### DIFF
--- a/mkt/reviewers/forms.py
+++ b/mkt/reviewers/forms.py
@@ -31,6 +31,9 @@ MODERATE_ACTION_FILTERS = (('', ''), ('approved', _lazy(u'Approved reviews')),
 MODERATE_ACTION_DICT = {'approved': mkt.LOG.APPROVE_REVIEW,
                         'deleted': mkt.LOG.DELETE_REVIEW}
 
+COMBINED_DEVICE_CHOICES = [('', _lazy(u'Any Device'))] + [
+    (dev.api_name, dev.name) for dev in mkt.DEVICE_TYPE_LIST]
+
 
 class ModerateLogForm(happyforms.Form):
     start = forms.DateField(required=False,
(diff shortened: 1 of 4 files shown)


mkt/reviewers/forms.py
```diff
@@ -185,6 +188,9 @@ class ApiReviewersSearchForm(ApiSearchForm):
         required=False,
         label=_lazy(u'Tarako-ready'),
         widget=CustomNullBooleanSelect)
+    dev_and_device = forms.ChoiceField(
```
diox: if anyone has a better name, I'm open to suggestions :)

mkt/reviewers/forms.py
```diff
@@ -204,6 +210,14 @@ def clean_status(self):
 
         return mkt.STATUS_CHOICES_API_LOOKUP.get(status, mkt.STATUS_PENDING)
 
+    def clean(self):
+        dev_and_device = self.cleaned_data.pop('dev_and_device', '').split('+')
+        self.cleaned_data['dev'] = dev_and_device[0]
+        if len(dev_and_device) > 1:
+            self.cleaned_data['device'] = dev_and_device[1]
```
robhudson: It looks like this is using the slug, but the search filter is expecting the ID?

robhudson: Ah, nvm, I see where it subclasses `ApiSearchForm` and that's handled there.

diox: Yep, I'll add a short comment explaining this.

mkt/reviewers/forms.py
robhudson: the name works for me.

robhudson:
r+

PR #2974, opened 2015-03-11
Pop auth header to avoid errors in other middleware (bug 1141162)

robhudson:


robhudson:
r? @diox 

I can write a test by setting the header, passing the request through both middlewares and assert no log.error is called. But I wanted to make sure this was what you were thinking?

diox:
That's what I was thinking, but now I'm having seconds thoughts and I'm wondering if that's such a good idea. It feels wrong to start deleting things from `request.META` - that makes debugging harder and more confusing.

Maybe instead, the `RestOAuthMiddleware` could check if the header starts with "mkt-shared-secret" and not use it if it's the case ?

robhudson:
We could also change the `log.error` to `log.warn` since it seems like we don't really care about these. I often grep the logs for "ERROR" looking for things that are obviously broken.

robhudson:
Updated to a different tactic, and added a test.

diox:
I agree about the log. It's nice to log these, but it's not an error.

mkt/api/tests/test_middleware.py
```diff
@@ -333,3 +334,12 @@ def test_settings(self):
         index = settings.MIDDLEWARE_CLASSES.index
 
         ok_(index(auth_middleware) > index(api_middleware))
+
+    @mock.patch('mkt.api.middleware.log')
+    def test_shared_secret_no_break_restoauth(self, mock_log):
+        shared_secret = 'mkt-shared-secret me@email.com,hash'
+        request = RequestFactory(HTTP_AUTHORIZATION=shared_secret).get('/')
+        request.API = True
+        RestSharedSecretMiddleware().process_request(request)
```
diox: Since you are no longer popping the header you don't need to call `RestSharedSecretMiddleware`, do you ?

robhudson: Yep, you're right.

I can update the logs while I'm at it too.

diox:
Needs rebasing to make travis happy

robhudson:
Rebased and tests pass.

diox:
r+

robhudson:
Thanks. Merged https://github.com/mozilla/zamboni/commit/7e03cdb

PR #2976, opened 2015-03-12
Fix translate link in moderation queue (bug 1141711)

eviljeff:


mkt/reviewers/views.py
```diff
@@ -918,7 +918,7 @@ def _retrieve_translation(text, language):
 
 
 @waffle_switch('reviews-translate')
-@reviewer_required
+@permission_required([('Apps', 'ModerateReview')])
```
diox: wouldn't `@reviewer_required(moderator=True)` work and be more consistent with the rest ?

eviljeff: `` @reviewer_required(moderator=True) `` gives access to reviewers *or* moderators.  Translation is used in the moderation queue, which is inaccessible to users with only Apps:Review permission.

diox: ah, ok

diox:
r+

PR #2979, opened 2015-03-13
class fixture loading support (bug 1142943)

eviljeff:
based on the implementation in TestCase and TransactionTestCase in Django 1.8

### DIFF
--- a/mkt/site/tests/__init__.py
+++ b/mkt/site/tests/__init__.py
@@ -9,10 +9,12 @@ from functools import partial
 from urlparse import SplitResult, urlsplit, urlunsplit
 
 from django import forms, test
+from django.db import connections, transaction, DEFAULT_DB_ALIAS
 from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
 from django.core.cache import cache
 from django.core.files.storage import default_storage as storage
+from django.core.management import call_command
 from django.core.urlresolvers import reverse
 from django.test.client import Client, RequestFactory
 from django.utils import translation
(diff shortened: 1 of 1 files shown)


mkt/site/tests/__init__.py
```diff
@@ -344,7 +346,87 @@ def login(self, profile):
 JINJA_INSTRUMENTED = False
 
 
-class TestCase(MockEsMixin, RedisTest, MockBrowserIdMixin, test.TestCase):
+class ClassFixtureTestCase(test.TestCase):
```
diox: Please add a docstring, maybe pointing to django's code.

mkt/site/tests/__init__.py
```diff
@@ -344,7 +346,87 @@ def login(self, profile):
...
+        # including mirrors or not. Otherwise, just on the default DB.
+        if getattr(cls, 'multi_db', False):
+            return [alias for alias in connections
+                    if (include_mirrors or
+                        connections[alias].settings_dict['TEST']['MIRROR'])]
+        else:
+            return [DEFAULT_DB_ALIAS]
+
+    @classmethod
+    def _enter_atomics(cls):
```
diox: Can you re-arrange the methods order to follow django's ? It's hard to look at the diff between your class and django's right now because the methods aren't in the same order.

eviljeff:
Re-arranged to follow the TestCase order in 1.8 (TransactionTestCase direct copy is really just _database_names)

mkt/site/tests/__init__.py
```diff
@@ -344,7 +346,99 @@ def login(self, profile):
...
+    def _fixture_setup(self):
+        assert not self.reset_sequences, (
+            'reset_sequences cannot be used on TestCase instances')
+        self.atomics = self._enter_atomics()
+
+    def _fixture_teardown(self):
+        self._rollback_atomics(self.atomics)
+
+    def _post_teardown(self):
+        # Monkey-patch so connections don't get closed
```
diox: Where does this come from ?

eviljeff: That's an original hack/monkeypatch :)
In 1.6 django's _post_teardown connections are closed and we don't want that to happen after each test anymore. https://github.com/django/django/blob/1.6.10/django/test/testcases.py#L788

diox: Ok, please modify the comment to explain that this is an original change

eviljeff:
comments updated

diox:
r+ let's try it

PR #2980, opened 2015-03-13
Remove unused dependencies (bug 846349)

diox:


diox:
@oremj please confirm that it looks ok.

jasonthomas:
lgtm

diox:
Relevant IRC discussion:
> if you are doing requirements cleanup, we don't use gunicorn anymore
> also, django-pymemcache==0.1.1-fixserial was for twemproxy testing
> also possibly heka-py-raven, since we don't route sentry messages through heka anymore

PR #2983, opened 2015-03-17
Avoid useless XHR when #receipt-check-result is absent (bug 1035376)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1035376

There is still a useless XHR done on install for free apps but at least with this change, we are no longer trying to do an xhr to `''`, which uselessly loads the current HTML page.

### DIFF
--- a/media/js/devreg/reviewers/editors.js
+++ b/media/js/devreg/reviewers/editors.js
@@ -182,7 +182,7 @@ function initReviewActions() {
 
     function check_receipt() {
         var $node = $('#receipt-check-result');
-        if ($node) {
+        if ($node.length) {
             $.getJSON($node.data('url'), function(data) {
                 if (data.status) {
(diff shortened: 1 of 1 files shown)


media/js/devreg/reviewers/editors.js
```diff
@@ -182,7 +182,7 @@ function initReviewActions() {
 
     function check_receipt() {
         var $node = $('#receipt-check-result');
-        if ($node) {
+        if ($node.length) {
```
diox: `$node`, being a jQuery object, was always true, even when there is no match (`'#receipt-check-result'` only exists for paid apps). Since it doesn't exist, `$node.data('url')` returned `undefined` and we ended up doing a useless XHR call to the page we were on.

washort: that's what happens when you let python developers write js

washort:
r+

PR #2984, opened 2015-03-17
Re-organize compiled requirements (bug 1143046)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1143046

### DIFF
--- a/requirements/compiled.txt
+++ b/requirements/compiled.txt
@@ -1,17 +1,16 @@
 # These should all be in our wheelhouse to avoid compiling on production and
 # automated testing machines
-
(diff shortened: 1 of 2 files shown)


eviljeff:
note exactly part of this commit, but are docs, test and load.txt still used?

diox:
Sort of.

- `dev.txt` is what you are supposed to use locally. It requires `test.txt`
- `test.txt` is what travis uses
- `load.txt` AFAIK is unused
- `docs.txt` is necessary to build docs. It's what readthedocs use.

eviljeff:
r+

PR #2987, opened 2015-03-18
 Remove mkt.webapps.views.BaseFilter (bug 1144663)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1144663

BaseFilter class is something we inherited from AMO. It was created 5 years ago, does a lot of stuff we don't really need anymore, and in a weird way. It should be much cleaner and reliable to simply use queryset filters directly.


washort:
Was just looking at this wondering if we could get rid of it. and you added more tests!
r+

PR #2988, opened 2015-03-19
Add pagination to versions list in review page (bug 1002434)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1002434

I originally wanted to do everything in JavaScript, but I prefer for something like that to be tested, and I didn't have the motivation to resurrect front-end testing in zamboni all by myself... so I did it in Django.

Screenshot of a faked (artificially set the limit of commbadge notes to load to 5 instead of 25 to show pagination) situation with every paginator shown :
screen shot 2015-03-19 at 16 11 20

### DIFF
--- a/mkt/reviewers/templates/reviewers/includes/review_history.html
+++ b/mkt/reviewers/templates/reviewers/includes/review_history.html
@@ -62,36 +62,33 @@
                   </td>
                 </tr>
               {% endif %}
-              {% if version.approvalnotes %}
-                <tr>
-                  <th>{{ _('Notes for Reviewers') }}</th>
-                  <td class="activity_approval">
-                    <div class="history-notes">
-                      {{ version.approvalnotes|urlize(100)|nl2br }}
-                    </div>
-                  </td>
-                </tr>
-              {% endif %}
+              {# No need to show approvalnotes, that'll be handled by commbadge, see below. #}
             </thead>
             <tbody>
               {# Results populated here by reviewers_commbadge.js (Commbadge API) #}
-              <tr class="comm-notes-paginator">
-                <th>&nbsp;</th>
-                <td>
-                  <a class="button next" href="#">{{ _('Next Page') }}</a>
-                  <a class="button prev" href="#">{{ _('Previous Page') }}</a>
-                </td>
-              </tr>
               <tr class="loading-msg">
                 <td>{{ _('Loading App History &hellip;')|safe }}</td>
               </tr>
             </tbody>
+            <tfoot>
+              {# Paginator will be displayed if necessary. #}
+              <tr class="comm-notes-paginator">
+                <th>&nbsp;</th>
+                <td>
+                  <a class="button next" href="#">{{ _('Next Comments') }}</a>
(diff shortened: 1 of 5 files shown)


diox:
@ngokevin , @eviljeff **r?**

I changed the position of the paginator for commbadge notes to always be at the bottom instead of the top because I thought it made sense given that the latest entries are at the bottom. I realize now that the paginator for *versions* being at the bottom contradicts this, though... Should I change it ?

mkt/reviewers/templates/reviewers/includes/review_history.html
```diff
@@ -62,36 +62,33 @@
                   </td>
                 </tr>
               {% endif %}
-              {% if version.approvalnotes %}
```
ngokevin: nice. i'm not sure if @eviljeff cares that these are at the top. In that case, you'd have to add some JS logic to render it here.

mkt/reviewers/templates/reviewers/includes/review_history.html
```diff
@@ -62,36 +62,33 @@
...
+              </tr>
+            </tbody>
+            <tfoot>
+              {# Paginator will be displayed if necessary. #}
               <tr class="comm-notes-paginator">
                 <th>&nbsp;</th>
                 <td>
-                  <a class="button next" href="#">{{ _('Next Page') }}</a>
-                  <a class="button prev" href="#">{{ _('Previous Page') }}</a>
+                  <a class="button next" href="#">{{ _('Next Comments') }}</a>
```
ngokevin: @eviljeff comments or notes?

ngokevin:
Position of paginator looks fine.

ngokevin:
Are we going to fetch only the threads we need in a different patch?

ngokevin:
lgtm, i didn't apply locally but it's straightforward

eviljeff:
I hate to throw in ux changes into the mix, but it'd actually be better to
lazy load the older notes and versions into the page while not removing the
existing notes  and versions and rather than a button have something
similar to a [...] bar at the top of the relevant section  (i.e.  a ...
section at the top of the versions list,  which would load older versions
into that gap in the history ). Similarly for older notes for a single
version.
If this is out of scope for this patch I'll file another bug.

diox:
@eviljeff that was my initial plan, so I can look into that. As mentioned in the description I just don't like working without tests :) I'll look into it and let you know if it needs to be splitted into a different bug.

diox:
@ngokevin: I've added the patch to not fetch threads we don't need.

@eviljeff: I've looked into it, but to do it right would require some changes in APIs IMHO, to get all the info we need to build the whole history table in JavaScript, so I think we should merge this and open another bug for your suggestion.

ngokevin:
r+

mkt/reviewers/templates/reviewers/includes/review_history.html
eviljeff: Notes is more accurate as some of them won't be comments (automatic messages) but I don't think it matters too much.

PR #2992, opened 2015-03-20
Force update_cached_manifests() task to be called after request is done (bug 1145160)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1145160

Otherwise it's possible for this task to be called too soon, before the view transaction has been committed. This was probably already a potential issue before, but by simplifying the transaction code, 31c0cc5f made it happen more often.

### DIFF
--- a/mkt/webapps/tasks.py
+++ b/mkt/webapps/tasks.py
@@ -286,6 +286,7 @@ def _update_manifest(id, check_hash, failed_fetches):
 
 
 @post_request_task
+@write
 def update_cached_manifests(id, **kw):
     try:
         webapp = Webapp.objects.get(pk=id)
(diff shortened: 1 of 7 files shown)


diox:
The test failures are because of the way `post_requests_task`s work in tests. The errors aren't new, it's because the file does not exist when creating an app with `app_factory()`, they were just hidden when we were using regular celery `@task`s. I need to find a way around this.

mkt/webapps/tasks.py
```diff
@@ -285,7 +285,8 @@ def _update_manifest(id, check_hash, failed_fetches):
         webapp.set_iarc_storefront_data()
 
 
-@task
+@post_request_task
+@write
 def update_cached_manifests(id, **kw):
```
diox: This part is the real fix, everything else is just adjusting mocks because of the way `@post_request_task` spawns sub-tasks.

washort:
r+

PR #2995, opened 2015-03-23
remove debug page

andymckay:
This was useful for development before desktop payments existed, because you could do a payments flow without a device. Desktop payments exist now, this is less useful.


diox:
You also need to patch `mkt/developers/urls.py` (see test failures)

andymckay:
Aargh, save failure in vim. Thanks.

diox:
r+

PR #2996, opened 2015-03-23
Create basic model for Mobile Optimized Websites (bug 1145377)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1145377

### DIFF
--- a/mkt/websites/models.py
+++ b/mkt/websites/models.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+from django.conf import settings
+from django.db import models
+
+import json_field
+
+from mkt.site.models import ModelBase
+from mkt.translations.fields import TranslatedField
+from mkt.tags.models import Tag
+from mkt.translations.fields import save_signal
+
+
+class Website(ModelBase):
+    default_locale = models.CharField(max_length=10,
+                                      default=settings.LANGUAGE_CODE)
+    url = TranslatedField()
+    title = TranslatedField()
+    short_title = TranslatedField()
+    description = TranslatedField()
+    keywords = models.ManyToManyField(Tag)
+    # FIXME regions
+    categories = json_field.JSONField(default=None)
+    icon_type = models.CharField(max_length=25, blank=True)
+    icon_hash = models.CharField(max_length=8, blank=True)
+    # FIXME devices
+    last_updated = models.DateTimeField(db_index=True, auto_now_add=True)
+    # FIXME status
+
+    @classmethod
+    def get_fallback(cls):
+        return cls._meta.get_field('default_locale')
+
+
+models.signals.pre_save.connect(save_signal, sender=Website,
+                                dispatch_uid='website_translations')
--- a/migrations/898-add-websites.sql
+++ b/migrations/898-add-websites.sql
@@ -1,0 +1,31 @@
+CREATE TABLE `websites_website_keywords` (
+    `id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY,
+    `website_id` integer NOT NULL,
+    `tag_id` integer NOT NULL,
+    UNIQUE (`website_id`, `tag_id`)
+)
+;
+ALTER TABLE `websites_website_keywords` ADD CONSTRAINT `tag_id_refs_id_46aff236` FOREIGN KEY (`tag_id`) REFERENCES `tags` (`id`);
+CREATE TABLE `websites_website` (
+    `id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY,
+    `created` datetime NOT NULL,
+    `modified` datetime NOT NULL,
+    `default_locale` varchar(10) NOT NULL,
+    `url` integer UNIQUE,
+    `title` integer UNIQUE,
+    `short_title` integer UNIQUE,
+    `description` integer UNIQUE,
+    `categories` longtext NOT NULL,
+    `icon_type` varchar(25) NOT NULL,
+    `icon_hash` varchar(8) NOT NULL,
+    `last_updated` datetime NOT NULL
+)
+;
+ALTER TABLE `websites_website` ADD CONSTRAINT `url_refs_id_7f415bd0` FOREIGN KEY (`url`) REFERENCES `translations` (`id`);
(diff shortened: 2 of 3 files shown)


mkt/websites/models.py
```diff
@@ -0,0 +1,37 @@
...
+
+class Website(ModelBase):
+    default_locale = models.CharField(max_length=10,
+                                      default=settings.LANGUAGE_CODE)
+    url = TranslatedField()
+    title = TranslatedField()
+    short_title = TranslatedField()
+    description = TranslatedField()
+    keywords = models.ManyToManyField(Tag)
+    # FIXME regions
```
diox: Copy `Webapp` mechanism (M2M like `AddonExcludedRegion`) ? Or use `json_field` like we do for categories ? It would be nice to re-use `RegionForm`.

mkt/websites/models.py
```diff
@@ -0,0 +1,37 @@
...
+from mkt.site.models import ModelBase
+from mkt.translations.fields import TranslatedField
+from mkt.tags.models import Tag
+from mkt.translations.fields import save_signal
+
+
+class Website(ModelBase):
+    default_locale = models.CharField(max_length=10,
+                                      default=settings.LANGUAGE_CODE)
+    url = TranslatedField()
```
chuckharmston: This is interesting. Are we using l10n to stash different idempotent locale URLs?

diox: Regarding translating the URL:

**PROS:**
* Some Websites might need this

**CONS:**
* Make it harder to check for unicity (almost impossible to add a constraint for it).

By the way, URL checking is going to be a nightmare. People might submit the http version when we already have the https version, but also submit with extra query string parameters, extra subdomains, trailing slash or not, maybe different path... Do we want to restrict the number of websites par domain to one ? Probably not, but we have to do *something*.

mkt/websites/models.py
```diff
@@ -0,0 +1,37 @@
...
+    title = TranslatedField()
+    short_title = TranslatedField()
+    description = TranslatedField()
+    keywords = models.ManyToManyField(Tag)
+    # FIXME regions
+    submitter_description = TranslatedField()
+    categories = json_field.JSONField(default=None)
+    icon_type = models.CharField(max_length=25, blank=True)
+    icon_hash = models.CharField(max_length=8, blank=True)
+    # FIXME devices
```
diox: Copy `Webapp` mechanism (M2M like `AddonDeviceType`) ? Or json_field ? It would be nice to re-use `DeviceTypeForm`.

mkt/websites/models.py
```diff
@@ -0,0 +1,37 @@
...
+    last_updated = models.DateTimeField(db_index=True, auto_now_add=True)
+    # FIXME submitter
```
diox: Copy `AddonUser` mechanism ? Call it `WebsiteSubmitter` ? Or just make a generic M2M ? (PRD says *"We want to track the other Submitters who submitted the same URL and “attach” them to the website."*)

mkt/websites/models.py
```diff
@@ -0,0 +1,37 @@
...
+                                      default=settings.LANGUAGE_CODE)
+    url = TranslatedField()
+    title = TranslatedField()
+    short_title = TranslatedField()
+    description = TranslatedField()
+    keywords = models.ManyToManyField(Tag)
+    # FIXME regions
+    submitter_description = TranslatedField()
+    categories = json_field.JSONField(default=None)
+    icon_type = models.CharField(max_length=25, blank=True)
```
diox: We'll need a `fetch_icon` task similar to the one that exists for webapps, but since the one for webapps parses manifest and what not, it'll have to be different. We should reuse the image resizing stuff though, and copy `get_icon_url()`

diox:
General note: how are we doing to handle updates to websites ? Which fields are we going to update, following what logic, and at which intervals ? I don't want to make the same mistakes we made for webapps, which are a mess atm.

mkt/websites/models.py
diox: We discussed translating the URL with @chuckharmston on IRC. The summary:
```
<chuck> If we just progressively enhance the URL we show to the user, I’m cool with that.
<chuck> No downside really.
<chuck> But I don’t think we can even, say, list the locales a site is available in.
<mat> I agree
<chuck> Let alone filter on it.
```

diox: @chuckharmston pointed out that Googles use `<link rel="alternate" hreflang="xx" href="yy" />` so we could detect this automatically, and if this `link` tag is found, we'd take the canonical url and extract the various locale versions automatically, and never let the submitter to it themselves.

diox:
Regarding updating, the plan that is emerging is to fully separate data entered by submitters from automatically generated data, and never mix the 2. So our nightly updates (ala hosted apps manifest update) can blindly overwrite automatically generated data.

chuckharmston:
Which doesn't have to prevent us from having e.g. an editorially-written description; we'd just store it separately in the model from `<meta name="description" />`

diox:
Yep. We already have `description` and `submitter_description` in that PR.

mkt/websites/models.py
```diff
@@ -0,0 +1,37 @@
...
+class Website(ModelBase):
+    default_locale = models.CharField(max_length=10,
+                                      default=settings.LANGUAGE_CODE)
+    url = TranslatedField()
+    title = TranslatedField()
+    short_title = TranslatedField()
+    description = TranslatedField()
+    keywords = models.ManyToManyField(Tag)
+    # FIXME regions
+    submitter_description = TranslatedField()
```
diox: `submitter_description` is phase 1+, so I'll remove it.

mkt/websites/models.py
diox: `submitter` information is phase 1+, so I'll remove it.

washort:
I wouldn't worry too much about reusing forms -- I expect we're going to revamp developer/reviewer tools at some point anyway, and I doubt we'll want to use django forms long-term.

diox:
We're supposed to be revamping developer/reviewer tools for years now... I like those forms because they are mostly separated from the rest (both the rest of the fields and the UI) so they're easy to re-use, already tested, and match the behaviour we want. Less code to maintain for the same result.

washort:
OK. r+, this matches my understanding of the fields in the PRD for this phase

PR #2997, opened 2015-03-24
Update curling to 0.4.0, getting rid of oauth2 (bug 846349)

diox:

Latest curling switched to oauthlib.


andymckay:
r+

PR #2999, opened 2015-03-24
Add basic ES indexer for websites (bug 1146958)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1146958

### DIFF
--- a/mkt/websites/indexers.py
+++ b/mkt/websites/indexers.py
@@ -1,0 +1,89 @@
+from operator import attrgetter
+
+from mkt.search.indexers import BaseIndexer
+from mkt.translations.models import attach_trans_dict
+from mkt.translations.utils import format_translation_es
+
+
+class WebsiteIndexer(BaseIndexer):
+    translated_fields = ('description', 'short_title', 'title', 'url')
+
+    @classmethod
+    def get_mapping_type_name(cls):
+        return 'mkt_website'
+
+    @classmethod
+    def get_model(cls):
+        """Returns the Django model this MappingType relates to"""
+        from mkt.websites.models import Website
+        return Website
+
+    @classmethod
+    def get_mapping(cls):
+        """Returns an Elasticsearch mapping for this MappingType"""
+        doc_type = cls.get_mapping_type_name()
+
+        mapping = {
+            doc_type: {
+                'properties': {
+                    'id': {'type': 'long'},
+                    'category': cls.string_not_analyzed(),
+                    'created': {'type': 'date', 'format': 'dateOptionalTime'},
+                    'description': {'type': 'string',
+                                    'analyzer': 'default_icu'},
+                    'default_locale': cls.string_not_indexed(),
+                    'icon_hash': cls.string_not_indexed(),
+                    'icon_type': cls.string_not_indexed(),
+                    'last_updated': {'format': 'dateOptionalTime',
+                                     'type': 'date'},
+                    'modified': {'type': 'date', 'format': 'dateOptionalTime'},
+                    'short_title': {'type': 'string',
+                                    'analyzer': 'default_icu'},
+                    'title': {'type': 'string', 'analyzer': 'default_icu'},
+                    # FIXME: Add custom analyzer for url, that strips http,
+                    # https, maybe also www. and any .tld ?
+                    'url': {'type': 'string', 'analyzer': 'simple'},
+
+                    # FIXME: categories, regions, devices, status. Might need
+                    # to refactor with webapps/indexers.py
+                }
+            }
+        }
+
+        # Add fields that we expect to return all translations.
+        cls.attach_translation_mappings(mapping, cls.translated_fields)
+
+        # FIXME: add indexed/analyzed translated fields mapping. Refactor with
+        # webapps/indexers.py.
+        return mapping
+
+    @classmethod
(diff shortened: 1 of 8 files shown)


mkt/websites/indexers.py
```diff
@@ -0,0 +1,88 @@
+from operator import attrgetter
+
+
```
ngokevin: one too many blank lines

mkt/websites/indexers.py
```diff
@@ -0,0 +1,88 @@
...
+                    'url': {'type': 'string', 'analyzer': 'default_icu'},
+
+                    # FIXME: categories, regions, devices, status. Might need
+                    # to refactor with webapps/indexers.py
+                }
+            }
+        }
+
+        # Add fields that we expect to return all translations.
+        cls.attach_translation_mappings(mapping, cls.translated_fields)
```
ngokevin: you could return this since it returns a mapping

mkt/websites/indexers.py
```diff
@@ -0,0 +1,88 @@
...
+
+        # Attach translations for searching and indexing.
+        attach_trans_dict(cls.get_model(), [obj])
+
+        attrs = ('created', 'default_locale', 'id', 'icon_hash', 'icon_type',
+                 'last_updated', 'modified')
+        doc = dict(zip(attrs, attrgetter(*attrs)(obj)))
+
+        doc['id'] = obj.pk
+        doc['description'] = list(
```
ngokevin: could do a loop here

mkt/websites/indexers.py
diox: Yeah, but I still need to do more things, see the FIXME that follows this line

mkt/websites/indexers.py
diox: I'm not sure yet if I'll keep all those, so I didn't want to add a loop just yet

mkt/websites/indexers.py
```diff
@@ -0,0 +1,89 @@
...
+from mkt.translations.models import attach_trans_dict
+from mkt.translations.utils import format_translation_es
+
+
+class WebsiteIndexer(BaseIndexer):
+    translated_fields = ('description', 'short_title', 'title', 'url')
+
+    @classmethod
+    def get_mapping_type_name(cls):
+        return 'mkt_website'
```
robhudson: why not just `'website'`? We call the apps document type `'webapp'.

mkt/websites/indexers.py
```diff
@@ -0,0 +1,89 @@
...
+            doc_type: {
+                'properties': {
+                    'id': {'type': 'long'},
+                    'category': cls.string_not_analyzed(),
+                    'created': {'type': 'date', 'format': 'dateOptionalTime'},
+                    'description': {'type': 'string',
+                                    'analyzer': 'default_icu'},
+                    'default_locale': cls.string_not_indexed(),
+                    'icon_hash': cls.string_not_analyzed(),
+                    'icon_type': cls.string_not_analyzed(),
```
robhudson: Are we going to do filters or queries on these? If not, we can use the `no_index` variant so we don't fill up the index for something we're not using.

mkt/websites/indexers.py
```diff
@@ -0,0 +1,89 @@
...
+                    'last_updated': {'format': 'dateOptionalTime',
+                                     'type': 'date'},
+                    'modified': {'type': 'date', 'format': 'dateOptionalTime'},
+                    'short_title': {'type': 'string',
+                                    'analyzer': 'default_icu'},
+                    'title': {'type': 'string', 'analyzer': 'default_icu'},
+                    # FIXME: we probably want url to be analyzed, but with a
+                    # better suited analyzer.
+                    'url': {'type': 'string', 'analyzer': 'default_icu'},
```
robhudson: Yeah, you're going to get some odd tokens like `'http'` and `'com'` that wouldn't make much sense.

mkt/websites/indexers.py
diox: Oops yeah, good catch, meant `cls.string_not_indexed()` not `cls.string_not_analyzed()`

mkt/websites/indexers.py
diox: Any suggestion of what analyzer to use ? I looked a the docs briefly and didn't find anything. Was hoping not to have to write a custom one.

robhudson: Possibly the simple analyzer? You can see how it will tokenize by hitting the URL: `http://localhost:9200/_analyze?text=http://www.example.com/test.html&analyzer=simple`. If not that maybe a custom analyzer that simply does a lowercase tokenizer. For URLs we might want to keep stopwords and not do stemming and such.

robhudson: IIRC you can also create your own custom stopwords filter and include tokens like `'html'`, `'http'`, etc.

mkt/websites/indexers.py
diox: The feed stuff has the `mkt_` prefix, so I flipped a coin... :)

mkt/websites/indexers.py
diox: Yeah, simple analyzer is not going to be enough, we are going to need a custom one that filters stuff out. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1147915

PR #3003, opened 2015-03-25
add regions/devices for websites

washort:


robhudson:
Should these also work their way into the ES indexing that @diox is working on?

washort:
https://github.com/mozilla/zamboni/commit/088973666fdcc43158e7c001c4c36f706097c68d

PR #3004, opened 2015-03-26
Website view/serializer

washort:


mkt/websites/models.py
```diff
@@ -17,7 +20,7 @@ class Website(ModelBase):
     title = TranslatedField()
     short_title = TranslatedField()
     description = TranslatedField()
-    keywords = models.ManyToManyField(Tag)
+    _keywords = models.ManyToManyField(Tag, db_column='keywords')
```
diox: `db_column` is not needed, since it's a `ManyToManyField` there is no column in the main table.

mkt/websites/tests/test_models.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def create_website(self, **data):
+        if 'keywords' in data:
+            kws = data.pop('keywords')
+        else:
+            kws = None
+        ws = Website.objects.create(**data)
+        if kws:
+            for k in kws:
+                ws._keywords.create(tag_text=k)
```
diox: Please move that to `website_factory`

mkt/websites/models.py
```diff
@@ -30,9 +33,70 @@ class Website(ModelBase):
...
+def delete_search_index(sender, instance, **kw):
+    instance.get_indexer().unindex(instance.id)
+
+
+# Save translations before saving Website instance with translated fields.
 models.signals.pre_save.connect(save_signal, sender=Website,
                                 dispatch_uid='website_translations')
+
+
+class DBStore(object):
```
diox: I don't really see the point of this abstraction. If it's to be able to run tests with `MemoryStore`, it's dangerous because its behaviour won't necessarily match reality. We can already switch from ES to DB easily in a view by just switching the queryset / paginator (not mandatory, just for perf) / serializer (which can inherit from a base one).

mkt/websites/serializers.py
```diff
@@ -0,0 +1,13 @@
+from drf_compound_fields.fields import ListField
+from rest_framework import serializers
+
+
+class WebsiteSerializer(serializers.Serializer):
+    url = serializers.CharField()
+    title = serializers.CharField()
+    short_title = serializers.CharField()
+    description = serializers.CharField()
+    keywords = serializers.CharField()
```
diox: Why not a `ListField` too ?

PR #3005, opened 2015-03-26
flake8: noqa > noqa

andymckay:
Apologies, I led people astray, `# flake8: noqa` ignores the entire file `# noqa` ignores just the line. In the case of a multi line import, the `# noqa` should be placed at the top. I fixed this up and an issues that resulted.

### DIFF
--- a/mkt/__init__.py
+++ b/mkt/__init__.py
@@ -1,18 +1,18 @@
 import threading
 
-from mkt.constants import (categories, comm, platforms, iarc_mappings,
-                           ratingsbodies)  # flake8: noqa
-from mkt.constants.applications import *  # flake8: noqa
-from mkt.constants.base import *  # flake8: noqa
-from mkt.constants.payments import *  # flake8: noqa
-from mkt.constants.platforms import *  # flake8: noqa
-from mkt.constants.search import *  # flake8: noqa
-from mkt.constants.submit import *  # flake8: noqa
+from mkt.constants import (categories, comm, platforms, iarc_mappings,  # noqa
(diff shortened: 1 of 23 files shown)


mkt/__init__.py
```diff
@@ -1,18 +1,19 @@
+# flake8: noqa
```
andymckay: I felt that since this file was essentially dodgy import * anyway, just noqa the entire file was fine.

diox: Maybe add a small comment to explain we want to ignore the entire file ?

diox: Also, is there any hope of removing the `import *` from this file and fix all the imports that use it in zamboni ? If there is, maybe we should file a bug. No idea how big of a task it is though.

mkt/constants/__init__.py
```diff
@@ -1,13 +1,14 @@
-from .apps import MANIFEST_CONTENT_TYPE  # flake8: noqa
+# flake8: noqa
```
diox: Like above, a small comment would be nice.

diox:
There are quite a few failures left to fix according to travis.

mkt/__init__.py
andymckay: I've no idea, I'll let someone else who knows more about zamboni go for that. I hope so, its just wrong.

andymckay:
Removed need for comments and fixed remaining failures.

diox:
Travis hates you :) More failures to fix.

diox:
r+

PR #3006, opened 2015-03-27
Add more fields to WebsiteIndexer and implement base serializer (bug 1146958)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1146958

Also add tests and slightly refactor the way we add translations to the indexer.

### DIFF
--- a/mkt/search/indexers.py
+++ b/mkt/search/indexers.py
@@ -12,9 +12,10 @@ import mkt
 from lib.es.models import Reindexing
 from lib.post_request_task.task import task as post_request_task
 from mkt.site.decorators import write
+from mkt.translations.utils import to_language
 
 
-task_log = logging.getLogger('z.task')
+log = logging.getLogger('z.task')
 
 
 class BaseIndexer(object):
@@ -261,9 +262,9 @@ class BaseIndexer(object):
         if not ids:
             return
 
-        task_log.info('Unindexing %s %s-%s. [%s]' %
-                      (cls.get_model()._meta.model_name, ids[0], ids[-1],
-                       len(ids)))
+        log.info('Unindexing %s %s-%s. [%s]' %
+                 (cls.get_model()._meta.model_name, ids[0], ids[-1],
+                  len(ids)))
 
         index = index or cls.get_index()
         # Note: If reindexing is currently occurring, `get_indices` will return
@@ -277,8 +278,8 @@ class BaseIndexer(object):
                     cls.unindex(id_=id_, es=es, index=idx)
                 except elasticsearch.exceptions.NotFoundError:
                     # Ignore if it's not there.
-                    task_log.info(u'[%s:%s] object not found in index' %
-                                  (cls.get_model()._meta.model_name, id_))
+                    log.info(u'[%s:%s] object not found in index' %
+                             (cls.get_model()._meta.model_name, id_))
 
     @classmethod
     def run_indexing(cls, ids, ES, index=None, **kw):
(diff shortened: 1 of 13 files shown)


mkt/webapps/indexers.py
```diff
@@ -222,18 +210,9 @@ def _locale_field_mapping(field, analyzer):
...
-                         % analyzer)
-                continue
-
-            mapping[doc_type]['properties'].update(
-                _locale_field_mapping('name', analyzer))
-            mapping[doc_type]['properties'].update(
-                _locale_field_mapping('description', analyzer))
+        # Add language-specific analyzers.
+        cls.attach_language_specific_analyzers(
+            mapping, ('name', 'description'))
```
robhudson: very nice.

mkt/websites/tests/test_indexers.py
```diff
@@ -12,14 +16,73 @@ def setUp(self):
...
+            'en-US': u'Site Tîtle',
+            'fr': u'Titrè du sïte',
+        }
+        self.obj.title = title
+        self.obj.save()
+        doc = self._get_doc()
+        eq_(sorted(doc['title']), [title['en-US'], title['fr']])
+        eq_(sorted(doc['title_translations']),
+            [{'lang': 'en-US', 'string': title['en-US']},
+             {'lang': 'fr', 'string': title['fr']}])
```
robhudson: There's also `assertSetEqual`, but this works too :)

mkt/websites/views.py
```diff
@@ -2,13 +2,23 @@
...
+from mkt.websites.serializers import ESWebsiteSerializer, WebsiteSerializer
+
+
+class WebsiteView(CORSMixin, MarketplaceView, ListAPIView):
+    cors_allowed_methods = ['get']
+    authentication_classes = [RestSharedSecretAuthentication,
+                              RestOAuthAuthentication]
+    permission_classes = [AllowAny]
+    serializer_class = WebsiteSerializer
+    model = Website
```
robhudson: Is this not hooked up to a URL yet? Or am I missing it?

mkt/websites/indexers.py
```diff
@@ -32,29 +33,32 @@ def get_mapping(cls):
                     'description': {'type': 'string',
                                     'analyzer': 'default_icu'},
```
robhudson: It's probably a good idea to use `'position_offset_gap': 100` for both `description` and `title` since they are a list of all translations. This helps prevent the last word of one translation and the first word of the next from matching in phrase queries.

robhudson:
r+wc. Nice abstractions.

mkt/websites/views.py
diox: Not hooked up to a URL yet no - for now, it's just to help testing (I'm manually instantiating the view in the tests)

mkt/websites/indexers.py
diox: In webapps, we only have `'position_offset_gap': 100` for `name`. I'll steal it for `title`, should I add it for `description` in both webapps and websites ?

robhudson: It only makes sense if we're doing phrase queries but it doesn't hurt in case. So yeah, good catch, description on webapps should have it also.

mkt/search/indexers.py
```diff
@@ -322,6 +326,87 @@ def attach_translation_mappings(cls, mapping, field_names):
             })
         return mapping
 
+    @classmethod
+    def attach_language_specific_analyzers(cls, mapping, field_names):
```
diox: Note to self: add a docstring before merging.

mkt/websites/tests/test_indexers.py
diox: Would have worked for the first, but not the second one, since it's a dict (and therefore unhashable)

PR #3007, opened 2015-03-27
Promote fxpay over mozpay (bug 1134296)

muffinresearch:
* Moved in-app products to the top
* Added a note about Desktop

r? @andymckay 

in-app_payments___fxpay-inapp-demo___firefox_marketplace


andymckay:
r+

PR #3010, opened 2015-03-30
add new locales; bug 1147788

clouserw:


clouserw:
These are being worked on by L10n, but not ready for production yet so only adding them to dev/stage

chuckharmston:
r+ if build passes

clouserw:
thanks

PR #3012, opened 2015-04-01
Implement basic search functionality for Websites (bug 1149989)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1149989

### DIFF
--- a/mkt/search/filters.py
+++ b/mkt/search/filters.py
@@ -47,15 +47,22 @@ class SearchQueryFilter(BaseFilterBackend):
             rules.append(
                 (query.Fuzzy, {'value': q, 'boost': 2, 'prefix_length': 1}))
 
+        # Apply rules to search on few base fields. Some might not be present
+        # in every document type / indexes.
         for k, v in rules:
-            for field in ('name', 'app_slug', 'author'):
+            for field in ('name', 'title', 'short_title',
+                          'app_slug', 'author', 'url'):
                 should.append(k(**{field: v}))
 
         # Exact matches need to be queried against a non-analyzed field. Let's
-        # do a term query on `name.raw` for an exact match against the app
+        # do a term query on `name.raw` for an exact match against the item
         # name and give it a good boost since this is likely what the user
         # wants.
+        # For websites, we do the same on title.raw. It won't work perfectly
+        # on some websites that use a description of the site in the title but
+        # it's better than nothing.
         should.append(query.Term(**{'name.raw': {'value': q, 'boost': 10}}))
+        should.append(query.Term(**{'title.raw': {'value': q, 'boost': 10}}))
 
         if analyzer:
             should.append(
@@ -106,7 +113,6 @@ class SearchFormFilter(BaseFilterBackend):
     }
 
     def filter_queryset(self, request, queryset, view):
-
         form = view.form_class(request.GET)
         if not form.is_valid():
             raise form_errors(form)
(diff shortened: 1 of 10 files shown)


mkt/search/forms.py
```diff
@@ -73,18 +73,7 @@
 
 
 class SimpleSearchForm(forms.Form):
```
diox: This form was not used at all so I'm hijacking it.

mkt/search/filters.py
```diff
@@ -47,16 +47,23 @@ def filter_queryset(self, request, queryset, view):
...
         # Exact matches need to be queried against a non-analyzed field. Let's
-        # do a term query on `name.raw` for an exact match against the app
+        # do a term query on `name.raw` for an exact match against the item
         # name and give it a good boost since this is likely what the user
         # wants.
         should.append(query.Term(**{'name.raw': {'value': q, 'boost': 10}}))
 
+        # FIXME: for websites, what should we match against ? "title" is going
+        # to be extracted from websites <title> so it won't work. "short_title"
+        # is only intended for display purposes. Or should we re-purpose it ?
```
robhudson: Why won't the extracted `<title>` work? That seems like the logical choice to me.

mkt/search/forms.py
```diff
@@ -95,6 +84,43 @@ class ApiSearchForm(forms.Form):
...
+                                                 [self.cleaned_data['cat']])
+
+    def clean_device_and_dev(self):
+        device = self.cleaned_data.pop('dev', None)
+        device_type = self.cleaned_data.pop('device', None)
+        # For android, we need to know the device type to determine the real
+        # device we are going to filter with, because we distinguish between
+        # mobile and tablets.
+        if device == 'android' and device_type:
+            device = '%s-%s' % (device, device_type)
```
robhudson: For some reason I thought we added a helper for this?

mkt/websites/models.py
```diff
@@ -49,6 +49,12 @@ def device_names(self):
         with no_translation():
             return [DEVICE_TYPES[d].api_name for d in device_ids]
 
+    def get_boost(self):
+        """
+        Returns the boost used in Elasticsearch for this website.
+        """
+        return 0.0
```
robhudson: The default should be 1.0, otherwise if you multiple the boost by the relevance you'll always get zero.

mkt/search/filters.py
diox: For instance, the `title` for nytimes.com would be `"The New York Times - Breaking News, World News & Multimedia"` and not just `"The New York Times"` - Users are going to use the latter, not the former.

(Another example that is inversed: the `title` for theguardian.com would be `"Latest news, sport and comment from the Guardian | The Guardian"` when most users would search for `"Guardian"` or `"The Guardian"` so the exact match will never work if we do it against the full title.

At the same time I feel that the short title would not work either: for nytimes.com it might be `"NYTytimes"` and not "The New York Times".

robhudson:
r+wc

mkt/search/forms.py
diox: Not in zamboni AFAIK. This is just moved from the other form.

mkt/websites/models.py
diox: Good catch. I'll try to write a test for relevancy that would have caught that.

mkt/search/filters.py
diox: For now we decided to keep the `title` and add a `title.raw` even though on some websites we'll never get the exact match. We might revisit this later and add a separate `name` attribute.

PR #3013, opened 2015-04-01
return a logo in the simulated in-app payment (bug 1137505)

andymckay:


mkt/inapp/views.py
```diff
@@ -81,13 +84,16 @@ def get_queryset(self):
         return qs
 
     def _create_stub_products(self):
-        for name, amount in (('Kiwi', '0.99'),
-                             ('Unicorn', '1.99')):
```
kumar303: the magic, ruined

kumar303:
looks good but this needs a database migration since the stub products have already been created in prod (and probably all other servers by now).

andymckay:
Ah rightio.

andymckay:
migration added, docs updated

migrations/900-remove-in-app.sql
```diff
@@ -0,0 +1,2 @@
+/* These will be recreated again automagically with the new data. */
+delete from inapp_products where simulate is not null;
```
kumar303: this should actually be ``... where stub = true`` (or however those messed up booleans in mysql work). This is because theoretically developers could kick their own products into simulation mode but we haven't added that feature yet.

kumar303:
Good call on the docs.

r+wc (migration needs a fix)

migrations/900-remove-in-app.sql
andymckay: Ah that works on prod where there's only two values with simulate in. But i'll change thats cool.

PR #3015, opened 2015-04-02
use the shiny new signing server (bug 1115184)

andymckay:


mkt/settings.py
```diff
@@ -1077,11 +1077,11 @@ def JINJA_CONFIG():
 SIGNED_APPS_OMIT_PER_FILE_SIGS = True
 
 # This is the signing REST server for signing receipts.
-SIGNING_SERVER = ''
+SIGNING_SERVER = (os.environ.get('SIGNING_SERVER', ''))
```
robhudson: Wrapping parens not needed.

robhudson:
I discovered https://pypi.python.org/pypi/python-decouple/2.3 recently and I like it. We should switch all of our `os.environment` calls to it. It would allow for defaults and proper casting, as well as other options for where to store these.

mkt/settings.py
andymckay: doh, thanks

robhudson:
r+. python-decouple is still cool.

andymckay-limited-access:
I am more than happy for you to use python-decouple in zamboni :)

PR #3017, opened 2015-04-03
Implement multi-type search for apps+websites (bug 1149995)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1149995

Sorting results by relevancy works, but most of the rest of the sort options are semi-broken either because the corresponding field does not exist in websites mapping (rating, reviewed, name_sort)
or because it's empty (popularity, trending). A follow-up bug will take care of those.

### DIFF
--- a/mkt/search/indexers.py
+++ b/mkt/search/indexers.py
@@ -375,6 +375,30 @@ class BaseIndexer(object):
             return mapping
 
     @classmethod
+    def attach_trending_and_popularity_mappings(cls, mapping):
+        doc_type = cls.get_mapping_type_name()
+        new_properties = {}
+
+        # Add global popularity and trending fields.
+        new_properties['popularity'] = {
+            'type': 'long', 'doc_values': True
+        }
+        new_properties['trending'] = {
+            'type': 'long', 'doc_values': True
+        }
+        # Add region-specific popularity and trending fields.
+        for region in mkt.regions.ALL_REGION_IDS:
+            new_properties['popularity_%s' % region] = {
+                'type': 'long', 'doc_values': True
+            }
+            new_properties['trending_%s' % region] = {
+                'type': 'long', 'doc_values': True
+            }
+        # Add everything to the mapping.
+        mapping[doc_type]['properties'].update(new_properties)
+        return mapping
+
+    @classmethod
     def extract_field_translations(cls, obj, field, db_field=None,
                                    include_field_for_search=False):
         """
--- a/mkt/search/serializers.py
+++ b/mkt/search/serializers.py
@@ -119,3 +119,21 @@ class BaseESSerializer(serializers.ModelSerializer):
 
     def to_datetime(self, value):
         return es_to_datetime(value)
+
+
+class DynamicSearchSerializer(serializers.Serializer):
+    def __init__(self, **kwargs):
+        super(DynamicSearchSerializer, self).__init__(**kwargs)
+        serializer_classes = self.context.get('serializer_classes', {})
+        self.serializers = {k: v(context=self.context)
+                            for k, v in serializer_classes.items()}
+
+    def to_native(self, obj):
+        """
+        Dynamically serialize obj using serializers passed through the context,
+        depending on the doc_type of the obj.
+        """
+        serializer = self.serializers.get(obj._meta['doc_type'])
+        if serializer is None:
+            return super(DynamicSearchSerializer, self).to_native(obj)
+        return serializer.to_native(obj)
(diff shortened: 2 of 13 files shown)


mkt/search/indexers.py
```diff
@@ -375,6 +375,32 @@ def _locale_field_mapping(field, analyzer):
...
+        new_properties['trending'] = {
+            'type': 'long', 'doc_values': True
+        }
+        for region in mkt.regions.ALL_REGION_IDS:
+            new_properties['trending_%s' % region] = {
+                'type': 'long', 'doc_values': True
+            }
+
+        mapping[doc_type]['properties'].update(new_properties)
+        return mapping
```
robhudson: Since we're normalizing this, these looks like they can be collapsed and put into a loop.

mkt/search/serializers.py
```diff
@@ -119,3 +119,21 @@ def _attach_translations(self, obj, data, field_names):
...
+
+    def to_native(self, obj):
+        """
+        Dynamically serialize obj using serializers passed through the context,
+        depending on the doc_type of the obj.
+        """
+        serializer = self.serializers.get(obj._meta['doc_type'])
+        if serializer is None:
+            return super(DynamicSearchSerializer, self).to_native(obj)
+        return serializer.to_native(obj)
```
robhudson: Nice.

mkt/search/views.py
```diff
@@ -54,6 +59,25 @@ def as_view(cls, **kwargs):
...
 
 
+class MultiSearchView(SearchView):
+    serializer_class = DynamicSearchSerializer
+
+    def get_serializer_context(self):
+        context = super(MultiSearchView, self).get_serializer_context()
+        context['serializer_classes'] = {
+            'webapp': ESAppSerializer,
+            'mkt_website': ESWebsiteSerializer
```
robhudson: The inconsistency of those names bugs me. :)

mkt/search/views.py
```diff
@@ -54,6 +59,25 @@ def as_view(cls, **kwargs):
...
+        }
+        return context
+
+    def get_queryset(self):
+        return Search(
+            using=BaseIndexer.get_es(),
+            doc_type=[WebappIndexer.get_mapping_type_name(),
+                      WebsiteIndexer.get_mapping_type_name()],
+            index=[WebappIndexer.get_index(), WebsiteIndexer.get_index()])
```
robhudson: nit: Can you put `index` before `doc_type` since it comes first in the ES URL?

mkt/search/tests/test_views.py
```diff
@@ -1252,3 +1254,66 @@ def test_suggestions_with_multiple_icons(self):
...
+            w.delete()
+        unindex_webapps(list(Webapp.with_deleted.values_list('id', flat=True)))
+        for w in Website.objects.all():
+            w.delete()
+        Website.get_indexer().unindexer(_all=True)
+        super(TestMultiSearchView, self).tearDown()
+
+    def _reindex(self):
+        self.reindex(Website, 'mkt_website')
+        self.reindex(Webapp, 'webapp')
```
robhudson: It's like fingernails on a chalkboard. :grin:

mkt/search/tests/test_views.py
```diff
@@ -1252,3 +1254,66 @@ def test_suggestions_with_multiple_icons(self):
...
+        self.anon.get(self.url)
+        assert _mock.called
+
+    def test_search(self):
+        res = self.anon.get(self.url)
+        eq_(res.status_code, 200)
+        objs = res.json['objects']
+        eq_(len(objs), 2)
+        eq_(objs[0]['id'], self.webapp.pk)
+        eq_(objs[1]['id'], self.website.pk)
```
robhudson: Is the order here based on something specific? Can it change somehow in the future? If both `pk`s are something like `1` can we also check for a type or something? This seems like a loose test is all I'm saying.

mkt/search/views.py
diox: Yes, as I wrote that I thought about your comment when I chose the name :) It's not too late to change it so I'll do that.

mkt/search/tests/test_views.py
```diff
@@ -1252,3 +1254,66 @@ def test_suggestions_with_multiple_icons(self):
...
+
+    def test_search_q(self):
+        res = self.anon.get(self.url, data={'q': 'something'})
+        eq_(res.status_code, 200)
+        objs = res.json['objects']
+        eq_(len(objs), 2)
+        eq_(objs[0]['id'], self.webapp.pk)
+        eq_(objs[1]['id'], self.website.pk)
```
robhudson: similar questions as above.

mkt/search/tests/test_views.py
diox: The order is specific, it's the default ordering, popularity. You're right, I should definitely check something else than just the pk.

mkt/search/tests/test_views.py
diox: Note to self: once the sorting issues are fixed I should make a test with a sorting order that makes the website come out on top.

robhudson:
r+wc. Nicely done.

PR #3020, opened 2015-04-07
Improve relevancy / score of websites in search (bug 1151814)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1151814

**TODO:**
- [x] Add `status` and `id_disabled` fields
- [x] Add something to track popularity by region (like `Installs` model for apps) and use it
- [x] Extract popularity and trending values
- [x] Adjust scoring, taking into account the fact that websites don't have `author` or `app_slug` to help their score compared to apps

### DIFF
--- a/migrations/900-websites-status-and-disabled.sql
+++ b/migrations/900-websites-status-and-disabled.sql
@@ -1,0 +1,27 @@
+ALTER TABLE `websites_website` ADD COLUMN `status` integer UNSIGNED NOT NULL, ADD COLUMN `is_disabled` bool NOT NULL;
+
+CREATE TABLE `websites_popularity` (
+    `id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY,
+    `created` datetime NOT NULL,
+    `modified` datetime NOT NULL,
+    `website_id` integer NOT NULL,
+    `value` double precision NOT NULL,
+    `region` integer UNSIGNED NOT NULL,
+    UNIQUE (`website_id`, `region`)
+);
+ALTER TABLE `websites_popularity` ADD CONSTRAINT `website_id_ref_popularity` FOREIGN KEY (`website_id`) REFERENCES `websites_website` (`id`);
+CREATE INDEX `websites_popularity_website_id` ON `websites_popularity` (`website_id`);
+CREATE INDEX `websites_popularity_region_id` ON `websites_popularity` (`region`);
+
+CREATE TABLE `websites_trending` (
+    `id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY,
+    `created` datetime NOT NULL,
+    `modified` datetime NOT NULL,
+    `website_id` integer NOT NULL,
+    `value` double precision NOT NULL,
+    `region` integer UNSIGNED NOT NULL,
+    UNIQUE (`website_id`, `region`)
+);
+ALTER TABLE `websites_trending` ADD CONSTRAINT `website_id_ref_trending` FOREIGN KEY (`website_id`) REFERENCES `websites_website` (`id`);
+CREATE INDEX `websites_trending_website_id` ON `websites_trending` (`website_id`);
+CREATE INDEX `websites_trending_region_id` ON `websites_trending` (`region`);
--- a/mkt/search/filters.py
+++ b/mkt/search/filters.py
@@ -50,8 +50,14 @@ class SearchQueryFilter(BaseFilterBackend):
         # Apply rules to search on few base fields. Some might not be present
         # in every document type / indexes.
         for k, v in rules:
-            for field in ('name', 'title', 'short_title',
-                          'app_slug', 'author', 'url'):
+            for field in ('name', 'short_title', 'app_slug', 'author'):
+                should.append(k(**{field: v}))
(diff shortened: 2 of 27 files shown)


diox:
My ultimate goal:

A `Website` with `(title='Something Unique', description='... Something Unique ...')` should be as high as a `Webapp` with `(name='Something Unique', description='Something Unique', app_slug='something-unique', author='something unique')`.

My plan to get there:
- Boost score higher on `title` (which only Websites have)
- Boost score higher on exact matches (`title.raw` and `name.raw`) so that it can still be higher than everything else with the new adjustments.

Does it make sense to you, @robhudson ? Happy to discuss this further.

robhudson:
> Add something to track popularity by region (like Installs model for apps) and use it

I would think we should just use GA and track click-throughs? Then that data can be pulled back in via Monolith into popularity, just like app installs are.

> Adjust scoring, taking into account the fact that websites don't have author or app_slug to help their score compared to apps

An admirable goal but seems tough. If apps have more fields to match against which increases its scores than websites will always be lagging behind. I'll be happy to help explore ideas.

diox:
> > Add something to track popularity by region (like Installs model for apps) and use it

> I would think we should just use GA and track click-throughs? Then that data can be pulled back in
> via Monolith into popularity, just like app installs are.

Yes, that's the plan. I'm just adding the model and using it like we do for apps, we'l take care of pulling the actual data later.

diox:
> An admirable goal but seems tough. If apps have more fields to match against which increases its scores than websites will always be lagging behind.

Let's discuss this on IRC ; my goal is not to make a super complex system, I just want to give websites a fighting chance.

mkt/search/filters.py
```diff
@@ -50,8 +50,14 @@ def filter_queryset(self, request, queryset, view):
...
-            for field in ('name', 'title', 'short_title',
-                          'app_slug', 'author', 'url'):
+            for field in ('name', 'short_title', 'app_slug', 'author'):
+                should.append(k(**{field: v}))
+        # We also add those rules with a bigger boost for website titles and
+        # urls, to compensate for them having less fields than apps.
+        for k, v in rules:
+            for field in ('title', 'url'):
+                v = v.copy()
+                v['boost'] *= 4
```
robhudson: Works for me.

mkt/search/indexers.py
```diff
@@ -399,6 +400,18 @@ def attach_trending_and_popularity_mappings(cls, mapping):
...
+    def extract_popularity_trending_boost(cls, obj):
+        extend = {
+            'boost': get_boost(obj),
+            'popularity': get_popularity(obj),
+            'trending': get_trending(obj)
+        }
+        for region in mkt.regions.ALL_REGIONS:
+            extend['trending_%s' % region.id] = get_trending(obj, region)
+            extend['popularity_%s' % region.id] = get_popularity(obj, region)
+        return extend
```
robhudson: :+1: beautiful

mkt/search/utils.py
```diff
@@ -9,3 +15,61 @@ def execute(self):
...
+        # popular or trending.
+        return 0
+
+    if region and not region.adolescent:
+        by_region = region.id
+    else:
+        by_region = 0
+
+    try:
+        return getattr(obj, property).get(region=by_region).value
```
robhudson: What does `getattr(obj, property)` give you back?

mkt/site/tests/__init__.py
```diff
@@ -850,23 +850,35 @@ def tearDown(self):
...
-            print "Could not refresh index '%s': %s" % (index, e)
+
+        if doctypes:
+            if not isinstance(doctypes, (list, tuple)):
+                doctypes = [doctypes]
+            indexes = [settings.ES_INDEXES[doctype] for doctype in doctypes]
+            try:
+                cls.es.indices.refresh(index=indexes)
+            except elasticsearch.NotFoundError as e:
+                print "Could not refresh indexes '%s': %s" % (indexes, e)
```
robhudson: I like this change.

mkt/search/utils.py
diox: The related manager - see `get_popularity()`/`get_trending()` methods below. I can add a comment to make that clear.

migrations/900-websites-status-and-disabled.sql
```diff
@@ -0,0 +1,27 @@
+ALTER TABLE `websites_website` ADD COLUMN `status` integer UNSIGNED NOT NULL, ADD COLUMN `is_disabled` bool NOT NULL;
+
+CREATE TABLE `websites_websitepopularity` (
```
robhudson: These table names don't seem ideal? Can we do `websites_popularity` and `websites_trending`?

mkt/search/utils.py
robhudson: Ah ok. I don't think a comment is necessary, I was just getting a bit lost in the changes.

robhudson:
There's a lot of great changes here, thanks. r+wc

migrations/900-websites-status-and-disabled.sql
diox: Done

PR #3022, opened 2015-04-07
Change the max file size to 250MB; bug 1151599

clouserw:


mkt/constants/submit.py
```diff
@@ -16,4 +16,4 @@
     (700, 1050, 'full'),  # Because it's proportional, that's why.
 ]
 
-MAX_PACKAGED_APP_SIZE = 100 * 1024 * 1024  # 100MB
+MAX_PACKAGED_APP_SIZE = 250 * 1024 * 1024  # 100MB
```
chuckharmston: Update comment?

mstriemer: Could assign an `MB` name instead, too.

```py
MB = 1024 * 1024
MAX_PACKAGED_APP_SIZE = 250 * MB

mstriemer:
r+wc

mkt/constants/submit.py
clouserw: wow, totally missed the comment

clouserw:
Bah, I wanted to do this via github, but two commits for this is ridiculous.  One sec.

PR #3026, opened 2015-04-08
Add NFC feature requirement and accept old signatures (bug 1150372)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1150372

Bonus: I've added support for parsing old signatures. Surprisingly, this wasn't working, so each time we added a feature, we were filtering incorrectly. Added tests to cover that.

### DIFF
--- a/mkt/constants/features.py
+++ b/mkt/constants/features.py
@@ -1,3 +1,4 @@
+import itertools
 from ordereddict import OrderedDict
 
 from django.conf import settings
@@ -335,6 +336,12 @@ APP_FEATURES = OrderedDict([
                              u'1GB RAM.'),
         'apis': (),
     }),
+    ('NFC', {
+        'name': _lazy(u'NFC'),
+        'description': _lazy(u'The app requires access to the Near Field '
+                             u'Communication (NFC) API.'),
+        'apis': ('navigator.mozNfc',),
+    })
 ])
 
 PRERELEASE_PERMISSIONS = [
@@ -351,7 +358,7 @@ class FeatureProfile(OrderedDict):
     representations.
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, _default=False, **kwargs):
         """
         Creates a FeatureProfile object.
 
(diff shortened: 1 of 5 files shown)


mkt/constants/features.py
```diff
@@ -335,6 +336,12 @@
                              u'1GB RAM.'),
         'apis': (),
     }),
+    ('NFC', {
+        'name': _lazy(u'NFC'),
+        'description': _lazy(u'The app requires access to Near Field '
```
robhudson: to *the* NFC API?

mkt/constants/features.py
```diff
@@ -335,6 +336,12 @@
...
+                             u'Communication (NFC) API.'),
+        'apis': ('navigator.mozNfc',),
```
robhudson: Why is `mozNfc` not `mozNFC`? Esp since all the methods under it capitalize `NFC` (e.g. `getNFCPeer`, `getNFCTag`). Weird.

mkt/constants/features.py
```diff
@@ -390,8 +401,10 @@ def from_signature(cls, signature):
         >>> FeatureProfile.from_signature('40000000.32.1')
         FeatureProfile([('apps', False), ('packaged_apps', True), ...)
         """
-        dehexed = int(signature.split('.')[0], 16)
-        return cls.from_int(dehexed)
+        # If the signature is invalid, let the ValueError be raised, it's up to
+        # the caller to decide what to do with it.
+        number, limit, version = signature.split('.')
+        return cls.from_int(int(number, 16), limit=int(limit))
```
robhudson: You've de-`dehexed` the code.

robhudson:
r+. Nice catch adding `limit`.

mkt/constants/features.py
diox: Yep, I wondered the same thing, had to double check that this wasn't a mistake in the documentation, but no, that's how it's defined, see https://developer.mozilla.org/en-US/docs/Web/API/Navigator/mozNfc and https://developer.mozilla.org/en-US/docs/Web/API/NFC_API/Using_the_NFC_API

PR #3027, opened 2015-04-08
Add new languages: sw, wo, ff, ar. Bug 1151944

clouserw:


clouserw:
Marked as not ready, waiting on L10n to catch up

andymckay:
Do other repos like webpay, marketplace-stats, fireplace etc need these?

clouserw:
fireplace reads it from zamboni (as a data attribute).  I kind of think (hope?) the others do too :)

clouserw:
we're doing these one at a time, so closing this

PR #3030, opened 2015-04-10
review pages for deleted apps bug 891159

eviljeff:
Makes review pages viewable even after the app has been deleted.  To workaround them not having a slug any more all deleted apps now get a numeric slug equal to their id on deletion (all numeric slugs being otherwise prohibited)

### DIFF
--- a/mkt/lookup/templates/lookup/helpers/app_header.html
+++ b/mkt/lookup/templates/lookup/helpers/app_header.html
@@ -40,6 +40,11 @@
             {{ _('Review History') }}
           </a>
         </li>
+        <li>
+          <a href="{{ url('commonplace.commbadge.app_dashboard', app.id) }}">
+            {{ _('Communication Dashboard') }}
+          </a>
+        </li>
         {% endif %}
       {% endif %}
     </ul>
--- a/mkt/webapps/decorators.py
+++ b/mkt/webapps/decorators.py
@@ -1,6 +1,5 @@
 import functools
 
-from django import http
 from django.core.exceptions import PermissionDenied
 from django.shortcuts import get_object_or_404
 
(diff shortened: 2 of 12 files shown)


mkt/webapps/decorators.py
```diff
@@ -39,26 +38,8 @@ def wrapper(request, addon_id, addon, *args, **kw):
...
-            # Don't get in an infinite loop if addon.app_slug.isdigit().
-            if addon.app_slug != addon_id:
-                url = request.path.replace(addon_id, addon.app_slug, 1)
-                if request.GET:
-                    url += '?' + request.GET.urlencode()
-                return http.HttpResponsePermanentRedirect(url)
-        elif addon_id:
-            addon = get(app_slug=addon_id)
-        elif app_slug:
-            addon = get(app_slug=app_slug)
```
eviljeff: All of this was unused - addon_id was only passed for add-on pages, all app views passed app_slug instead.

mkt/lookup/templates/lookup/helpers/app_header.html
```diff
@@ -40,6 +40,11 @@
             {{ _('Review History') }}
           </a>
         </li>
+        <li>
+          <a href="{{ url('commonplace.commbadge.app_dashboard', app.id) }}">
+            {{ _('Comm Messages') }}
```
robhudson: Can we expand what "Comm" means to ease translations?

eviljeff: As a comment or the link text itself?

mkt/webapps/tests/test_models.py
```diff
@@ -155,8 +155,8 @@ def test_soft_deleted(self):
         # When an app is deleted its slugs and domain should get relinquished.
         post_mortem = Webapp.with_deleted.filter(id=app.id)
         eq_(post_mortem.count(), 1)
-        for attr in ('app_slug', 'app_domain'):
-            eq_(getattr(post_mortem[0], attr), None)
+        eq_(getattr(post_mortem[0], 'app_domain'), None)
+        eq_(getattr(post_mortem[0], 'app_slug'), '337141')
```
robhudson: I wonder if there ever would be an app named numerically and its slug would conflict with a deleted app? Just in case should we prefix it with `deleted-` so the app slug becomes `deleted-337141`?

mkt/lookup/templates/lookup/helpers/app_header.html
robhudson: I was thinking the link text itself.

robhudson:
r+wc. Looks good to me.

mkt/webapps/tests/test_models.py
eviljeff: We prevent developers setting all numeric slugs because there are a few methods (mainly API, iirc) that can work with an ID or a slug. 'deleted-337141' is a valid slug though, so could conflict :)

eviljeff: https://github.com/eviljeff/zamboni/blob/view-deleted-app-history-891159/mkt/webapps/models.py#L314

PR #3031, opened 2015-04-13
Remove celery-tasktree

robhudson:
It doesn't seem like we were actually using this. Celery 3.x also has newer options to do similar task chaining so let's get rid of the dependency.


diox:
r+

PR #3033, opened 2015-04-14
fix payments link (bug 1154264)

andymckay:


robhudson:
I trust that this URL is correct. r+ but wait to see if tests pass.

PR #3034, opened 2015-04-15
Remove default_language from region API (bug 1154837)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1154837

It's a backwards-incompatible change, but that property is unreliable, was never really documented, and useless to third parties.


robhudson:
r+ with test fixes

PR #3035, opened 2015-04-16
Second attempt to use class fixtures in TestCase (bug 1142943)

eviljeff:


diox:
r+

mkt/websites/indexers.py
```diff
@@ -29,7 +29,7 @@ def get_mapping(cls):
                 'properties': {
                     'id': {'type': 'long'},
                     'category': cls.string_not_analyzed(),
-                    'created': {'type': 'date', 'format': 'dateOptionalTime'},
+                    'created': {'format': 'dateOptionalTime'},
```
robhudson: Is this supposed to be here?

robhudson:
Looks like some infinite recursion happening in the test suite?

mkt/websites/indexers.py
diox: Yeah, it was added on purpose to reproduce the issue that caused us to revert this patch the previous time. If you look at the travis build, you'll see a failure caused by this particular change, which is fine, and then a bunch of other failures that should not happen. The goal is to find out what causes those errors and then when we do, we can remove this part of the patch.

eviljeff:
for reference: see https://github.com/mozilla/zamboni/pull/2979 and https://github.com/mozilla/zamboni/pull/3001 for previous pull and revert

eviljeff:
The tests are failing as expected now, after moving some of the initialisation in ESTestCase to setUpTestData from setUpClass and putting both fixture and test data setup in a try.  Just need to fix the (purposely) broken code in indexers.py before merging.

diox:
r+

PR #3039, opened 2015-04-17
Bug 1093210 move newrelic to a separate file

graingert:
This is to unblock https://bugzilla.mozilla.org/show_bug.cgi?id=1093210


mkt/commonplace/templates/commonplace/index.html
```diff
@@ -27,9 +27,11 @@
     <link rel="dns-prefetch" href="//www.google-analytics.com">
     <link rel="dns-prefetch" href="//stats.g.doubleclick.net">
 
-    {{ newrelic_header()|safe if repo == 'fireplace' }}
+    {% if repo == 'fireplace' %}
+      <script type="text/javascript" src="{{ media('js/lib/newrelic.js') }}" id='newrelic-footer'></script>
```
washort: is this id on the correct element?

graingert: Yep

diox:
Per discussion on the bug, I don't think we should follow that approach. Instead, we should look into integrating newrelic with our js bundles. See https://bugzilla.mozilla.org/show_bug.cgi?id=1093210 comment 6 to 9.

PR #3045, opened 2015-04-20
add utf-8 food

clouserw:
I thought we could use some utf-8 food

### DIFF
--- a/mkt/webapps/fakedata.py
+++ b/mkt/webapps/fakedata.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 import datetime
 import hashlib
 import itertools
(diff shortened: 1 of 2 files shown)


chuckharmston:
r- doesn't weight burrito sufficiently high.

chuckharmston:
(but r+ really)

washort:
r+ them's good eatin'

clouserw:
@washort it looks like the tests are finding a legitimate failure but I don't know why they are failling on my PR and not already...

mkt/webapps/fakedata.py
```diff
@@ -29,11 +30,11 @@
 from mkt.webapps.models import AddonUser, AppManifest, Preview
 
 adjectives = ['Exquisite', 'Delicious', 'Elegant', 'Swanky', 'Spicy',
-              'Food Truck', 'Artisanal', 'Tasty']
+              'Food Truck', 'Artisanal', 'Tasty', 'Questionable', u'Drôle']
 nouns = ['Sandwich', 'Pizza', 'Curry', 'Pierogi', 'Sushi', 'Salad', 'Stew',
          'Pasta', 'Barbeque', 'Bacon', 'Pancake', 'Waffle', 'Chocolate',
-         'Gyro', 'Cookie', 'Burrito', 'Pie']
-fake_app_names = list(itertools.product(adjectives, nouns))[:-1]
```
washort: this was just my little joke to keep "Tasty Pie" from being generated. :-)

mkt/webapps/tests/test_fakedata.py
```diff
@@ -25,7 +25,7 @@ def test_smallset(self):
         ok_(not any(appname[-1].isdigit() for appname, cat in data))
 
     def test_bigset(self):
-        size = 300
+        size = 1500
```
washort: Good idea, this should work just fine now that I've simplified things a bit; the original version of this code was mainly intended for small sets so I didn't look too close at the bigger cases.

washort:
r+

clouserw:
Thanks!

PR #3046, opened 2015-04-21
Update prod requirements to include compiled.txt.

jasonthomas:
Related to Bug 1156442.

r? @robhudson

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,4 +1,4 @@
-# test.txt also includes prod.txt and compiled.txt
+# test.txt also includes prod.txt
 -r test.txt
 
 translate-toolkit==1.12.0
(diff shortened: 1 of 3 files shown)


diox:
So the goal is to stop using RPMs and simply have wheels for every compiled dependency ?

robhudson:
Just to understand completely... we still need ops to create the wheels for the platform but when finished you put them on pyrepo and they get pip installed like the rest of the packages. Is that correct?

jasonthomas:
> So the goal is to stop using RPMs and simply have wheels for every compiled dependency ?
Yes. I think it would be easier to maintain project specific dependencies going forward.

> Just to understand completely...  we still need ops to create the wheels for the platform but when finished you put them on pyrepo
Anyone should be able to generate the wheel packages within a RHEL6 env, but ops can continue to build them as needed.

> and they get pip installed like the rest of the packages. Is that correct?
Correct.

diox:
cool, thanks!

robhudson:
Yeah, I like this. Thanks @jasonthomas.  r+

PR #3047, opened 2015-04-21
Switch previews/icons tasks to post_request_task (bug 1153669)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1153669

This should ensure the tasks for previews and icons handling are not fired before all transactions have been committed.

As a result, the `transaction.atomic()` call in mkt/submit/views.py is now useless since the task will be called after the view transaction is done anyway.

### DIFF
--- a/mkt/developers/tasks.py
+++ b/mkt/developers/tasks.py
@@ -11,12 +11,10 @@ import traceback
 import urlparse
 import uuid
 import zipfile
-from datetime import date
 
 from django import forms
 from django.conf import settings
 from django.core.files.storage import default_storage as storage
-from django.utils.http import urlencode
 
 import requests
 from appvalidator import validate_app, validate_packaged_app
@@ -138,7 +136,7 @@ def _hash_file(fd):
     return hashlib.md5(fd.read()).hexdigest()[:8]
 
 
-@task
+@post_request_task
 @set_modified_on
 def resize_icon(src, dst, sizes, locally=False, **kw):
     """Resizes addon icons."""
@@ -207,7 +205,7 @@ def pngcrush_image(src, hash_field='image_hash', **kw):
         log.error('Error optimizing image: %s; %s' % (src, e))
 
 
-@task
+@post_request_task
 @set_modified_on
 def resize_preview(src, instance, **kw):
     """Resizes preview images and stores the sizes on the preview."""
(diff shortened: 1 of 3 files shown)


mkt/developers/tasks.py
```diff
@@ -483,35 +458,6 @@ def fetch_manifest(url, upload_pk=None, **kw):
...
-    data['_ri_'] = settings.RESPONSYS_ID
-
-    try:
-        res = requests.get('http://awesomeness.mozilla.org/pub/rf',
-                           data=urlencode(data))
-        return res.status_code == 200
-    except requests.RequestException:
-        return False
-
-
```
robhudson: Yay for dead code removal!

mkt/developers/tests/test_forms.py
```diff
@@ -53,6 +54,9 @@ def test_preview_size(self):
             copyfileobj(open(get_image_path(name)), f)
         assert form.is_valid(), form.errors
         form.save(self.addon)
+        # Since the task is a post-request-task and we are outside the normal
+        # request-response cycle, manually send the tasks.
+        post_request_task._send_tasks()
```
robhudson: Is this the only test affected?

mkt/submit/views.py
```diff
@@ -102,39 +101,38 @@ def manifest(request):
...
-                                                  manifest=True, details=False)
+        AddonUser(addon=addon, user=request.user).save()
+        # Checking it once. Checking it twice.
+        AppSubmissionChecklist.objects.create(addon=addon, terms=True,
+                                              manifest=True, details=False)
 
-            # Create feature profile.
-            addon.latest_version.features.update(**features_form.cleaned_data)
+        # Create feature profile.
+        addon.latest_version.features.update(**features_form.cleaned_data)
```
robhudson: Nice work simplifying this. Are there any other places we need to adjust? I'm curious if there's any management commands that deal with icons or previews that need to trigger the tasks manually now?

robhudson:
r+wc

mkt/submit/views.py
diox: When a `post_request_task` is called from another task the queue is also emptied, so I think we should be fine - most management commands dealing with addons use `process_addons`, which spawns a task.

mkt/developers/tests/test_forms.py
diox: Apparently - any view tests should already be fine, since the `request_finished` signal (which `post_request_tasks` hooks up to) should be sent when we call a view through the test client.

PR #3049, opened 2015-04-22
change generate_apps to use status name not id

eviljeff:


diox:
r+ (test failures are unrelated, already happening on master)

Would be nice to replace the few hardcoded `4` by a constant in `mkt/webapps/fakedata.py` too...

PR #3051, opened 2015-04-22
Add even more regions (bug 1157067)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1157067

screen shot 2015-04-22 at 14 40 25


media/css/devreg/payments.styl
```diff
@@ -697,7 +697,7 @@ label .region {
 }
 
 // India (in) is in quotes since `in` is a reserved word.
-for region in ar bd br bw ci cl cm cn co cr cz de ec eg es fr gr gt gw hu 'in' it jo jp lt me mg ml mm mu mx ne ni pa pe ph pl rs ru sn sv tn tz uk us uy ve vu za restofworld
+for region in ar bd br bw cf ci cl cm cn co cr cz de ec eg es fr gn gr gt gw hu 'in' it jo jp ke lt me mg ml mm mu mx ne ni pa pe ph pl rs ru sn sv tn tz uk us uy ve vu za restofworld
```
mstriemer: Would be nice to pull these from marketplace-constant's regions.styl, but not a big deal.

mstriemer:
r+

media/css/devreg/payments.styl
diox: I wanted to do that, but zamboni only imports marketplace-constants as a python package, it doesn't handle front-end dependencies at all, so that would require some changes in the way it's built. I'll file a bug though.

PR #3052, opened 2015-04-22
enable new locales; bug 1157068

clouserw:


clouserw:
Not ready yet.  This is running into us using pt in the SHORT_LANGUAGES and also I think HIDDEN_LANGUAGES is no longer a thing.  I think this PR will evolve...

washort:
r+

PR #3053, opened 2015-04-22
remove HIDDEN_LANGUAGES; bug 1157775

clouserw:
HIDDEN_LANGUAGES was left over from AMO when we had a dropdown on the list that people could use to choose their language.  Now we preview languages on dev/stage before enabling them in production so this isn't needed anymore.

### DIFF
--- a/migrations/335-perms-locales.py
+++ b/migrations/335-perms-locales.py
@@ -4,7 +4,7 @@ from mkt.access.models import Group, GroupUser
 
 
 LANGS = sorted(list(
-    set(settings.AMO_LANGUAGES + settings.HIDDEN_LANGUAGES) -
+    set(settings.AMO_LANGUAGES) -
     set(['en-US'])))
 
 
--- a/mkt/settings.py
+++ b/mkt/settings.py
@@ -797,17 +797,6 @@ HEKA_CONF = {
 
 HEKA = client_from_dict_config(HEKA_CONF)
 
-# Not shown on the site, but .po files exist and these are available on the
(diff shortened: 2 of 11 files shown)


migrations/335-perms-locales.py
```diff
@@ -4,7 +4,7 @@
 
 
 LANGS = sorted(list(
-    set(settings.AMO_LANGUAGES + settings.HIDDEN_LANGUAGES) -
+    set(settings.AMO_LANGUAGES) -
```
robhudson: You didn't want to rename this to `MKT_LANGUAGES`? :grin:

mkt/settings.py
```diff
@@ -797,17 +797,6 @@ def langs(languages):
...
-# Not shown on the site, but .po files exist and these are available on the
-# L10n dashboard.  Generally languages start here and move into AMO_LANGUAGES.
-# This list also enables translation edits.
-HIDDEN_LANGUAGES = (
-    # List of languages from AMO's settings (excluding mkt's active locales).
-    'af', 'ar', 'fa', 'fi', 'he', 'id', 'mn', 'pt-PT', 'sl', 'sv-SE',
-    'uk', 'vi',
-    # The hidden list from AMO's settings:
-    'cy',
-)
```
robhudson: Some of these aren't in `settings.LANGUAGES` (e.g. `'he'`) so this will actually remove them. Is that ok?

clouserw: Yep, that's why I had to change the unit tests too.  I have another bug about removing some inactive locales and activating some complete locales.  Just trying to clean up right now.

robhudson:
r+

PR #3054, opened 2015-04-23
update file unzip size limit; bug 1151599

clouserw:


mstriemer:
Is this going to be big enough? Looks like the file that is trying to be uploaded is 229MB zipped so I'd imagine it's more than 250MB unzipped.

Maybe `FILE_UNZIP_SIZE_LIMIT = 2 * constants.MAX_PACKAGED_APP_SIZE` might be good?

digitarald:
Agreed. Those assets compress pretty well ZIP compression is usually around 2-7x depending on file type. Why do we need the file size for files inside the zip?

mstriemer:
I thought this was checking total size, if its per file then it will probably work. I'm assuming this is here to prevent someone from uploading a zip that is just `a` repeated a lot so that it compresses a huge amount and then explodes in size on the server. If we're checking each file individually though then you could just make more small files. Maybe that isn't an issue because we'll only ever read one file at a time?

mstriemer:
Looks like the file size check was introduced in https://github.com/mozilla/zamboni/commit/9e28ae25df2071008cd540b7a125844f1352d8d3 (June 1, 2011) for [bug 625801](https://bugzilla.mozilla.org/show_bug.cgi?id=625801) which I don't have access to.

clouserw:
I CC'd you two.  This is for the size of a single file.  Yvan filed the bug, he might have a comment here.  I don't know his github name though.

ygjb:
@mstriemer is correct, the concern was zip bombing.  I am not terribly concerned with the size of the limit (assuming that we aren't seeing these as actual attacks - the initial finding was part of a security review), more that there is one that is enforced.  We can always adjust it if necessary.

clouserw:
@ygjb any advice on the size?  We have 250MB zip files now.  You think it's still valuable to have this limit if it's (for example) 4x that size?

digitarald:
4x max limit sounds sane. We can adapt when we actually see something bigger.

clouserw:
ok, made it 1GB.  @mstriemer r?

mstriemer:
r+

PR #3056, opened 2015-04-24
Work around django-cache-machine redis breakage (bug 881405)

diox:
The latest version of django-cache-machine parses `REDIS_BACKENDS` incorrectly. See https://github.com/django-cache-machine/django-cache-machine/issues/92

This switches us to a fork that has a patch fixing that bug while we wait for that patch to be integrated into upstream master.


diox:
@jasonthomas r?

jasonthomas:
r+

PR #3057, opened 2015-04-24
Set dir and lang on fireplace pages (bug 1155826)

mstriemer:
It makes the spinner go in the right direction.

rtl-from-zamboni mov


mkt/commonplace/templates/commonplace/index.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html{% if appcache %} manifest="{{ url('commonplace.appcache')|urlparams(repo=repo) }}"{% endif %}>
+<html{% if appcache %} manifest="{{ url('commonplace.appcache')|urlparams(repo=repo) }}"{% endif %}{% if repo == 'fireplace' %} lang="{{ LANG }}" dir="{{ DIR }}"{% endif %}>
```
eviljeff: do bad things happen if fireplace isn't special-cased?

mstriemer: I don't think so, I just don't think the other repos have the same list of available locales and I'd rather we not set this and then have it get changed once the page loads.

mstriemer:
Added some tests.

mstriemer:
I updated this to always set lang and dir regardless of the app in https://github.com/mozilla/zamboni/commit/7bfc43d422edc0061484ec924e9727d0a274b55c.

PR #3058, opened 2015-04-24
Fix up redis configuration to use REDIS_BACKEND instead.

jasonthomas:
As far as I can tell we aren't using REDIS_BACKENDS anywhere. I've updated monitor to use REDIS_BACKEND instead and cleaned up sites settings. We may still need to remove references to redisutils but the only place I am seeing it being called is in https://github.com/mozilla/zamboni/blob/660cac59fd09954ca719c1124719ee4c107a1aba/mkt/site/management/commands/clean_redis.py which is not being used.


diox:
https://gist.github.com/diox/b404614d8289d0f16649 should fix some of these test failures.

jasonthomas:
Thanks for the patch!

diox:
r+

PR #3059, opened 2015-04-24
remove _ajaxy stuff, not being used

andymckay:
Just noticed this while I was passing through the code, only one test was using it, may as well remove it.


eviljeff:
r+

robhudson:
I didn't even know that was there. r+

PR #3060, opened 2015-04-24
Update queue to use correct date field (bug 1084654)

robhudson:


mkt/reviewers/tests/test_models.py
```diff
@@ -32,6 +34,15 @@ def setUp(self):
         self.grant_permission(self.admin_user, '*:*', name='Admins')
         user_factory(email='regular')
 
+        # Disconnect indexing signals on rereview and escalation models.
```
diox: why ?

robhudson: Ah, let me remove that now. I had a theory those were causing test problems but let me try again without. I did this early on in this branch.

robhudson:
Updated to remove the signal disconnects.

diox:
r+

PR #3063, opened 2015-04-28
Remove "nuggets" dependency (bug 1156848)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1156848


robhudson:
r+. We can also remove the import names from the `.baked` file.

PR #3066, opened 2015-04-29
Remove dependency on SQLAlchemy (bug 1159220)

diox:

Also stop having a `SERVICES_DATABASE` setting, since it actually points to the default database anyway (discussed with @jasonthomas)

### DIFF
--- a/mkt/receipts/tests/test_verify.py
+++ b/mkt/receipts/tests/test_verify.py
@@ -4,7 +4,6 @@ import time
 import uuid
 from urllib import urlencode
 
-from django.db import connection
 from django.conf import settings
 from django.test.client import RequestFactory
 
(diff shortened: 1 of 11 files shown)


robhudson:
Shouldn't that also be converted to a regular django view with a url instead of the python/wsgi thing it is now? Or is that a later bug?

robhudson:
r+. This looks fine but I'd be interested to see it moved into a regular django app if possible.

diox:
See discussion in the bug. It should, but it's more work, so I've created a follow-up bug to take care of that later: https://bugzilla.mozilla.org/show_bug.cgi?id=1159746

robhudson:
Ok, r+ then.

PR #3069, opened 2015-04-30
Remove default_currency from regions API (bug 1160136)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1160136

### DIFF
--- a/mkt/api/serializers.py
+++ b/mkt/api/serializers.py
@@ -70,7 +70,7 @@ class CarrierSerializer(serializers.Serializer):
 
 
 class RegionSerializer(CarrierSerializer):
-    default_currency = serializers.CharField()
+    pass
 
 
 class CategorySerializer(serializers.Serializer):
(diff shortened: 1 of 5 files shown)


mkt/api/serializers.py
```diff
@@ -70,7 +70,7 @@ class CarrierSerializer(serializers.Serializer):
 
 
 class RegionSerializer(CarrierSerializer):
-    default_currency = serializers.CharField()
+    pass
```
mstriemer: Doesn't really matter but `RegionSerializer = CarrerSerializer`?

mstriemer:
lgtm r+

mkt/api/serializers.py
diox: I thought about doing that, but I preferred leaving them separate for now, I might tweak the region one soon. If we do merge them I'd prefer simply having `RegionOrCarrierSerializer` or something similar.

PR #3070, opened 2015-05-01
reviewers search endpoint for websites

washort:


mkt/reviewers/urls.py
```diff
@@ -71,6 +72,8 @@
     url(r'reviewers/', include(reviewers_router.urls)),
     url('^reviewers/search', views.ReviewersSearchView.as_view(),
         name='reviewers-search-api'),
+    url('^reviewers/sites/search', WebsiteSearchView.as_view(),
```
diox: Should be `ReviewersSearchView` right ?

washort: there's already a `ReviewersSearchView` used in the line previous, want me to rename it to `ReviewersWebsiteSearchView`?

diox: oh, yeah, had not caught the import, that was surprising. Maybe rename to `ReviewersWebsiteSearchView` yeah, including in `views.py` so that you don't need to import as a different name ?

mkt/websites/views.py
```diff
@@ -48,3 +53,11 @@ def as_view(cls, **kwargs):
         # to be wrapped in transactions.
         view = super(WebsiteSearchView, cls).as_view(**kwargs)
         return non_atomic_requests(view)
+
+
+class ReviewersWebsiteSearchView(WebsiteSearchView):
+    permission_classes = [GroupPermission('Apps', 'Review')]
```
diox: We'll probably want a new permission like `Websites:Review`. Thoughts, @eviljeff ?

diox:
r+, we can change the permission later

mkt/websites/views.py
eviljeff: If this was used within a reviewer tool for website, Websites:Review is fine.  Though as they don't exist currently I'm less sure what it should be (Apps:Review is wrong anyway)

PR #3071, opened 2015-05-01
Bug 990000 - input notification image aligned

shiveeg1:
Team member email input notification image is aligned correctly.

### DIFF
--- a/media/css/devreg/authors.styl
+++ b/media/css/devreg/authors.styl
@@ -34,10 +34,10 @@
             border-color: $error-red;
         }
         &.invalid {
-            background: url(../../img/zamboni/notifications.png) no-repeat 229px -157px;
+            background: url(../../img/zamboni/notifications.png) no-repeat 254px -161px;
         }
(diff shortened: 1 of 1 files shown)


eviljeff:
These numbers don't align the icon correctly (I played around with Inspect Element and changed the css).  Its the vertical number which is causing the icon to be clipped by the bottom of the field.
r-

shiveeg1:
Thanks for the explanation eviljeff. So do need only the vertical spacing or the horizontal too to align the notification next to the border of the field ?

eviljeff:
To be perfect you need both - so the numbers in your original patch fix the horizontal issue so the icon is closer to the border; but you need some vertical adjustment to stop the icons being clipped by the bottom border.

shiveeg1:
Got it! made the vertical and horizontal changes and squashed both commits into one. Hope this one works . Thanks eviljeff !

eviljeff:
r+

PR #3072, opened 2015-05-01
Use cursor as a context manager to fix connection errors (bug 1159220)

diox:


diox:
Needed to fix errors (`2013, 'Lost connection to MySQL server during query'`) seen on stage following https://github.com/mozilla/zamboni/commit/e4dcdc9842853fb71e7e05c9ceefa0b4e846289f

PR #3073, opened 2015-05-01
Revert 72fe88e and f76c11a since it's causing issues on dev/stage (bug 1159220)

diox:


ddurst:
r+

PR #3074, opened 2015-05-04
Fix data type incoherencies with production api

itorres:
Not a deep test, found these while playing with the API with Go (strict typing). There are probably more but I won't be able to hunt for them for a week.


washort:
thanks!

PR #3075, opened 2015-05-04
Exclude fields we don't need when querying ES (bug 1127144)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1127144

To keep the `hidden_fields` property manageable, I had to use wildcards and change the name of the analyzer-specific variants of the translated fields, so we'll need a reindex, but I *think* it should still be ok to use this code with an old index, worst case scenario some search results will be missing some specific results for the duration of the reindex.

### DIFF
--- a/mkt/search/indexers.py
+++ b/mkt/search/indexers.py
@@ -30,6 +30,10 @@ class BaseIndexer(object):
     """
     _es = {}
 
+    """Fields we don't need to expose in the results, only used for filtering
+    or sorting."""
+    hidden_fields = ()
+
     @classmethod
     def _key(cls, es_settings):
         """
@@ -120,9 +124,11 @@ class BaseIndexer(object):
         """
         Returns a `Search` object from elasticsearch_dsl.
         """
-        return Search(using=using or cls.get_es(),
-                      index=cls.get_index(),
-                      doc_type=cls.get_mapping_type_name())
+        return (Search(
+            using=using or cls.get_es(),
+            index=cls.get_index(),
+            doc_type=cls.get_mapping_type_name())
+            .extra(_source={'exclude': cls.hidden_fields}))
 
     @classmethod
     def get_index(cls):
@@ -352,7 +358,7 @@ class BaseIndexer(object):
         """
         def _locale_field_mapping(field, analyzer):
             return {
-                '%s_%s' % (field, analyzer): {
+                '%s_l10n_%s' % (field, analyzer): {
                     'type': 'string',
                     'analyzer': (('%s_analyzer' % analyzer)
                                  if analyzer in mkt.STEMMER_MAP else analyzer)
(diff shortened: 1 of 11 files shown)


mkt/search/tests/test_filters.py
```diff
@@ -52,14 +52,14 @@ def test_q(self):
...
-                                               'analyzer': 'english_analyzer',
-                                               'type': 'phrase'}}}
+        ok_({'match': {'name_l_english': {'query': 'search terms',
+                                          'boost': 2.5}}}
             in should)
+        ok_({'match': {'description_l_english':
+            {'query': 'search terms',
+             'boost': 0.6,
+             'analyzer': 'english_analyzer',
+             'type': 'phrase'}}} in should)
```
robhudson: I find the nesting on this confusing but I suppose it's fine if that's the best it can get.

mkt/webapps/indexers.py
```diff
@@ -30,9 +49,10 @@ def search(cls, using=None):
 
         We override this to use our patched version which adds statsd timing.
         """
-        return Search(using=using or cls.get_es(),
-                      index=cls.get_index(),
-                      doc_type=cls.get_mapping_type_name())
+        return (Search(
+            using=using or cls.get_es(), index=cls.get_index(),
+            doc_type=cls.get_mapping_type_name())
+            .extra(_source={'exclude': cls.hidden_fields}))
```
robhudson: Is this the same change as the one in the parent class? Also, I don't see the statsd call the comment refers to?

mkt/search/indexers.py
```diff
@@ -352,7 +358,7 @@ def attach_language_specific_analyzers(cls, mapping, field_names):
         """
         def _locale_field_mapping(field, analyzer):
             return {
-                '%s_%s' % (field, analyzer): {
+                '%s_l_%s' % (field, analyzer): {
```
robhudson: is `l` for l10n? I'm assuming this is to help differentiate the field in some way? Maybe a comment somewhere would help explain the single letter.

robhudson:
r+wc

mkt/webapps/indexers.py
diox: The statsd call is inside the `Search.execute` method. The `Search` class here is not the original one. See `mkt/search/utils.py`.

So, yes, it's the same change. I wanted to replace this by just a `with statsd.timing...()` and `super()` call, but it doesn't work because the `statsd.timing` call needs to be applied to the `execute` method of `Search` class :(

mkt/search/indexers.py
diox: Yeah. I'm open to suggestions - it doesn't have to be one single letter, we just need *something* that can make it easy to exclude all these fields. Maybe `_l10n_` would work, actually.

robhudson: Yeah, I think `_l10n_` would help make it clear what these are for.

diox: done

PR #3065, opened 2015-04-29
API for review actions

washort:
Expose the reviewer actions as API resources.

PR #3076, opened 2015-05-05
API for website review actions

washort:
Just like https://github.com/mozilla/zamboni/pull/3065 but for websites instead of apps.


mkt/reviewers/urls.py
```diff
@@ -74,6 +74,10 @@
         name='reviewers-search-api'),
     url('^reviewers/sites/search', ReviewersWebsiteSearchView.as_view(),
         name='reviewers-website-search-api'),
+    url(r'^reviewers/sites/(?P<pk>[^/<>"\']+)/approve/$',
+        views.WebsiteApprove.as_view(), name='website-approve'),
+    url(r'^reviewers/sites/(?P<pk>[^/<>"\']+)/reject/',
```
diox: Missing '$' at the end

mkt/reviewers/views.py
```diff
@@ -1005,6 +1006,27 @@ def post(self, request, pk, region, *args, **kwargs):
         return Response({'approved': bool(form.cleaned_data['approve'])})
 
 
+class _WebsiteAction(object):
+    permission_classes = [GroupPermission('Apps', 'Review')]
```
diox: Can you add a new `Websites:Review` permission, and change it here and in `ReviewersSearchView` ? 

@eviljeff do you know which groups we should assign the permission to ? Do we need a new group ? (Note that this isn't necessary for MOW Phase 0, only Phase 1, so we can fix things later, there is no rush)

eviljeff: Eventually we will need a `Website Reviewers` group eventually, with this permission added to it.  Also add `Websites:*` to `Staff` group.

mkt/reviewers/tests/test_views_api.py
```diff
@@ -301,6 +302,34 @@ def test_no_premium_filtering(self):
...
+class TestWebsiteReviewerActions(RestOAuth):
+    fixtures = fixture('user_2519')
+
+    def setUp(self):
+        super(TestWebsiteReviewerActions, self).setUp()
+        self.website = website_factory(
+            title='something',
+            categories=json.dumps(['books', 'sports']))
+        self.user = UserProfile.objects.get(pk=2519)
+        self.grant_permission(self.user, 'Apps:Review')
```
diox: Needs to be `Websites:Review`

mkt/reviewers/tests/test_views_api.py
```diff
@@ -301,6 +302,34 @@ def test_no_premium_filtering(self):
         eq_(len(res.json['objects']), 1)
 
 
+class TestWebsiteReviewerActions(RestOAuth):
```
diox: Can you add 2 small permissions tests ? Like one with an anonymous user, one with a logged in user with no permissions

diox:
r+wc and tests passing

mkt/reviewers/tests/test_views_api.py
washort: d'oh

migrations/909-website-reviewers.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO groups (id, name, rules, created, modified, notes)
+VALUES (50080, 'Website Reviewers', 'Websites:Review', NOW(), NOW(), 'Users who can review websites');
```
chuckharmston: Let's not include IDs in our migrations in the future; this caused a deploy error for duplicate IDs.

migrations/909-website-reviewers.sql
```diff
@@ -0,0 +1,3 @@
...
+UPDATE groups SET rules=CONCAT(rules, 'Websites:Review') WHERE name = 'Staff';
```
chuckharmston: There also should have been a prepending comma in front of `Websites:Review`; this will have broken the Staff group for anyone for whom this migration successfully ran, setting `rules` to `Apps:*,Users:Edit,Stats:View,AdminTools:View,AccountLookup:View,AppLookup:View,Lookup:View,Stats:ViewWebsites:Review`.

chuckharmston:
I was told that this wasn't supposed to be merged yet. If that's the case, let's back it out until we're sure it's solid. For now, the migration was fixed in 98b3bd36 and the dev database was manually patched. I'll hold off on reversions until I hear from somebody who was working on this.

cc @washort @diox

diox:
Damn. Thanks for fixing this, I didn't pay enough attention to the migration when reviewing this.

I'm hesitant to revert it now that it has been fixed, because of the migration - I hate having inconsistent states between local/dev/stage/prod. And I am about to merge more (unrelated) code with migrations now, incrementing the migration id.

chuckharmston:
I was thinking more along the lines of backing out all the other code, since the migration has already been fixed. But that was only because someone indicated that this was mistakenly committed. The code itself looks fine.

diox:
Ok - it doesn't hurt to have it so I don't think we should revert it. It's not necessary *now*, but it will be soon (for phase 1)

PR #3079, opened 2015-05-06
Sort regions (and carrier) API list by name (bug 1058257)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1058257

Nice to have to drop the full list of regions from the js.


ngokevin:
r+

PR #3080, opened 2015-05-06
disable stale locales; bug 1157363

clouserw:


clouserw:
This might leave orphan rows in the database if someone happened to created translations in these languages.  They would be un-orphaned when we re-enabled the languages.

eviljeff:
r+

PR #3081, opened 2015-05-08
fake apps with a mix of privileged and unprivileged versions

washort:
also fix various bugs around package/version generation


eviljeff:
r+

PR #3090, opened 2015-05-18
Update categories for HOW Phase 0 (bug 1161869)

robhudson:


robhudson:
I'll clean up the tests, which used old categories.

mkt/webapps/tasks.py
```diff
@@ -744,3 +744,26 @@ def populate_is_offline(ids, **kw):
...
+@task
+@write
+def adjust_categories(ids, **kw):
+    ADJUSTMENTS = {
+        'books': 'books-comics',
+        'news-weather': 'news',
+    }
+    for chunk in chunked(ids, 100):
+        for app in Webapp.objects.filter(pk__in=chunk):
+            cats = app.categories
```
diox: Since it's a `list`, it won't be copied to `cats`, instead `cats` will be a reference to `app.categories`. I'd just remove the intermediary variable, it's not very useful - you don't display the old categories list anywhere so copying would be useless.

diox:
How are we going to handle the push ? Specifically, how are we going to avoid serving empty category pages ?

Assuming we update frontend with those categories as well:

- If we push frontend changes first, `"news"`, `"weather"`, `"books-comics"` and all new categories will be empty
- If we reindex first, we need to push the frontend as soon as reindex is done and not a minute after, otherwise old frontend code will be trying to display categories like "news-weather" and "books" which would now be empty

(Might not be a huge issue, but we need to have a plan :)

robhudson:
@diox I hadn't thought about front-end, etc. I thought we'd run this command right after push, which also updates the search index for each app that changes. Is that good enough in your opinion?

mkt/webapps/tasks.py
```diff
@@ -744,3 +744,61 @@ def populate_is_offline(ids, **kw):
...
+                    app.categories.append(v)
+            if save:
+                task_log.info(u'[app:{0}] Adjusted categories: {1}'
+                              .format(app, app.categories))
+                app.save()
+    # Add apps to new categories.
+    for cat, apps in NEW_CAT_APPS.items():
+        for app in Webapp.objects.filter(pk__in=apps):
+            if cat not in app.categories:
+                app.categories.append(cat)
```
diox: None of those apps already have 2 categories ?

robhudson: Yeah, this is something I will fix. I have the above list but now in `{<app>: [<list of cats]}` format for that case.

settings_test.py
```diff
@@ -73,3 +73,53 @@ def _polite_tmpdir():
 TASK_USER_ID = '4043307'
 TEMPLATE_DEBUG = False
 VIDEO_LIBRARIES = ['lib.video.dummy']
```
robhudson: I'll remove this below but don't want to trigger another travis build just now.

mkt/search/tests/test_views.py
```diff
@@ -200,6 +200,11 @@ def test_right_category(self):
         eq_(res.status_code, 200)
         eq_(res.json['objects'], [])
 
+    def test_old_category(self):
+        res = self.anon.get(self.url, data={'cat': 'books'})
+        eq_(res.status_code, 200)
+        eq_(res.json['objects'], [])
```
diox: We should also add a test to make sure the old category name properly returns content from the new one.

diox:
r+

robhudson:
https://github.com/mozilla/zamboni/commit/1a463a5

PR #3091, opened 2015-05-19
Guess rating language and filter queries on it (bug 875455)

washort:
Based on mstriemer's code in https://github.com/mozilla/zamboni/pull/2818

### DIFF
--- a/docs/api/topics/ratings.rst
+++ b/docs/api/topics/ratings.rst
@@ -21,6 +21,10 @@ _`List`
 
     :query app: the ID or slug of the app whose ratings are to be returned.
     :query user: the ID of the user or `mine` whose ratings are to be returned.
+    :query lang: a language to filter ratings by if `match_lang` is set.
+    :query match_lang: a boolean to specify to match language or not. If 1, only
+                       results matching `lang` are returned. Otherwise all
+                       results are returned.
 
     The value `mine` can be used to filter ratings belonging to the currently
     logged in user.
@@ -41,7 +45,8 @@ _`List`
             },
             "info": {
                 "average": "3.4",
-                "slug": "marble-run"
+                "slug": "marble-run",
+                "total_reviews": 391
             },
             "objects": [
                 {
@@ -49,6 +54,7 @@ _`List`
                     "body": "This app is top notch. Aces in my book!",
                     "created": "2013-04-17T15:25:16",
                     "is_author": true,
+                    "lang": "en",
                     "modified": "2013-04-17T15:34:19",
                     "rating": 5,
                     "resource_uri": "/api/v2/apps/rating/19/",
--- a/mkt/ratings/models.py
+++ b/mkt/ratings/models.py
@@ -7,6 +7,7 @@ from django.db import models
 import bleach
 from caching.base import CachingQuerySet
 from celery import task
+
 from tower import ugettext_lazy as _
 
 from mkt.site.models import ManagerBase, ModelBase
@@ -56,6 +57,8 @@ class Review(ModelBase):
     rating = models.PositiveSmallIntegerField(null=True)
     title = TranslatedField(require_locale=False)
     body = TranslatedField(require_locale=False)
+    lang = models.CharField(max_length=5, null=True, blank=True,
+                            editable=False)
     ip_address = models.CharField(max_length=255, default='0.0.0.0')
 
     editorreview = models.BooleanField(default=False)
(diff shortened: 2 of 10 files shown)


mkt/ratings/models.py
```diff
@@ -120,6 +130,47 @@ def transformer(reviews):
...
+    if confidence < 0.9:
+        depunct = dict.fromkeys([ord(c) for c in string.punctuation], u' ')
+        words = text.lower().translate(depunct).split()
+        for w in words:
+            assumed_lang = COMMON_REVIEW_WORDS.get(w)
+            if assumed_lang:
+                break
+        if assumed_lang is None and len(words) <= 3:
+            # Too little text to reliably guess.
+            return None
```
mstriemer: Could we fallback to the user's language at this point?

migrations/880-lang-on-reviews.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE reviews
```
diox: Need to bump migration number.

mkt/ratings/models.py
```diff
@@ -120,6 +130,47 @@ def transformer(reviews):
         for user in UserProfile.objects.no_cache().filter(id__in=user_ids):
             user_ids[user.id].user = user
 
+    def guess_language(self):
```
mstriemer: The language guessing code feels a bit out of place in `mkt.ratings.models`. Maybe it should go in a utils module somewhere?

mkt/ratings/views.py
```diff
@@ -137,6 +142,7 @@ def post_save(self, obj, created=False):
             mkt.log(mkt.LOG.EDIT_REVIEW, app, obj)
             log.debug('[Review:%s] Edited by %s' %
                       (obj.pk, self.request.user.id))
+        obj.guess_language()
```
mstriemer: Perhaps this should add `lang=guess_language(obj.body) or obj.user.lang` to the params before it saves instead of doing a second save after. It would work well with moving the language guessing code out of `mkt.ratings.models` too.

mkt/ratings/serializers.py
```diff
@@ -82,6 +90,11 @@ def validate(self, attrs):
             user = self.request.user
             attrs['user'] = user
             attrs['ip_address'] = self.request.META.get('REMOTE_ADDR', '')
+            guessed_lang = guess_language(attrs['body'])
```
mstriemer: Would we ever want `guessed_lang` to be falsey? We shouldn't need to explicitly check against `None`, right? I think an `or` would be more clear here.

```py
attrs['lang'] = guess_language(attrs['body']) or user.lang

File: mkt/ratings/tests/test_models.py [hunk 6]
```diff
@@ -1,8 +1,9 @@
+# -*- coding: utf-8 -*-
 from mock import patch
 from nose.tools import eq_
 
 import mkt.site.tests
-from mkt.ratings.models import check_spam, Review, Spam
+from mkt.ratings.models import check_spam, guess_language, Review, Spam
```
mstriemer: Looks like this is going to blow up.

docs/api/topics/ratings.rst
```diff
@@ -21,6 +21,11 @@ _`List`
 
     :query app: the ID or slug of the app whose ratings are to be returned.
     :query user: the ID of the user or `mine` whose ratings are to be returned.
+    :query lang: a language to filter ratings by if `match_lang` is set.
+    :query match_lang: a boolean to specify to match language or not. If unset
+                       all results are returned, if 1 only results matching
+                       `lang` are returned if 0 only results not matching
```
mstriemer: I think this only supports the `match_lang=1` case. I'm not sure if we need the inverse case, my original mockup of the UI had a "my language / not my language" control but we probably really want a "my language / all" control which I think it why this no longer supports `match_lang=0`.

mkt/ratings/tests/test_views.py
```diff
@@ -364,6 +364,45 @@ def test_already_rated_version(self):
...
+
+    def test_lang_filter_inverse(self):
+        Review.objects.create(addon=self.app, user=self.user, body='yes',
+                              lang='en')
+        other_user = UserProfile.objects.exclude(pk=self.user.pk)[0]
+        Review.objects.create(addon=self.app, user=other_user, body='yes',
+                              lang='pt')
+        res, data = self._get_url(self.list_url, app=self.app.app_slug,
+                                  client=self.anon, lang='pt',
+                                  match_lang='0')
```
mstriemer: This test should likely be removed along with the documentation saying that `match_lang=0` gives ratings in other languages.

mstriemer:
ltgm wc

Not sure if you want someone else to review this since I wrote part of it and haven't been doing much Python lately. Looks good to me though.

mkt/ratings/tests/test_models.py
washort: whoops, that's what i get for only running the tests i changed :)

mkt/ratings/serializers.py
washort: In general I prefer to avoid surprises or code that requires extra thinking to read.

PR #3092, opened 2015-05-19
Add all the countries (bug 1058257)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1058257

Also, since the flags don't make much sense with a list of 250+ countries, remove them from compatibility page.


mkt/api/tests/test_views.py
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf8 -*-
```
washort: mind changing this to `utf-8`? python doesn't care but emacs doesn't understand it without the hyphen

mkt/developers/tests/test_commands.py
```diff
@@ -60,8 +60,9 @@ def test_migration_of_regional_content(self):
         # Exclude in everywhere except Brazil.
         regions = list(mkt.regions.REGIONS_CHOICES_ID_DICT)
         regions.remove(mkt.regions.BRA.id)
-        for region in regions:
-            self.webapp.addonexcludedregion.create(region=region)
+        AddonExcludedRegion.objects.bulk_create(
```
washort: nice

washort:
r+

PR #3094, opened 2015-05-21
API for "Report an issue" with a Website (bug 1167171)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1167171


mstriemer:
You didn't change it but the comment on https://github.com/diox/zamboni/blob/add-abuse-website-api/mkt/abuse/models.py#L21-L22 looks out of date.

```
# An abuse report can be for an addon, a user, or a website. Only one of
# these should be null.
```
Should be...
```
# An abuse report can be for an addon, a user, or a website. Only one of
# these should be set.
```

diox:
Fixed, good catch.

mkt/abuse/serializers.py
```diff
@@ -3,28 +3,37 @@
...
 
 
 class UserAbuseSerializer(BaseAbuseSerializer):
     user = SplitField(serializers.PrimaryKeyRelatedField(), UserSerializer())
 
-    class Meta:
-        model = AbuseReport
-        fields = ('text', 'ip_address', 'reporter', 'user')
+    class Meta(BaseAbuseSerializer.Meta):
+        fields = BaseAbuseSerializer.Meta.fields + ('user',)
```
mstriemer: :+1:

mkt/abuse/tests/test_views.py
```diff
@@ -144,3 +151,22 @@ def test_invalid_app(self):
...
+
+class TestWebsiteAbuseResource(AbuseResourceTests, BaseTestAbuseResource,
+                               RestOAuth):
+    resource_name = 'website'
+
+    def setUp(self):
+        super(TestWebsiteAbuseResource, self).setUp()
+        self.website = website_factory()
+        self.default_data = {
+            'text': "This website is weird.",
```
mstriemer: Mixing `'` and `"`

mkt/commonplace/urls.py
```diff
@@ -33,6 +33,11 @@ def fireplace_route(path, name=None):
     url('^(?P<review_id>\d+)/', include(fireplace_reviews_patterns)),
 )
 
+fireplace_website_patterns = patterns(
+    '',
+    fireplace_route('', 'website.detail'),
```
mstriemer: Ah, I probably should have done that. Thanks!

mkt/settings.py
```diff
@@ -298,7 +298,8 @@ def path(*args):
 APP_DELETION_EMAIL = 'marketplace-staff+deletions@mozilla.org'
 SUPPORT_GROUP = 'https://groups.google.com/forum/#!forum/mozilla.appreview'
 MARKETPLACE_EMAIL = 'marketplace-staff@mozilla.org'
-MKT_FEEDBACK_EMAIL = 'apps-feedback@mozilla.com'
+MKT_APPS_FEEDBACK_EMAIL = 'apps-feedback@mozilla.com'
```
mstriemer: Should this just go away?

mstriemer:
Just tried it out with `curl` locally, looks good! Thanks for doing this so quickly.

r+

mkt/settings.py
diox: It's certainly messy , but it needs to stay since it's the email used for general Marketplace feedback (the email is misleading :/).

PR #3095, opened 2015-05-21
fix url regex for translate API endpoint to allow non alphanumeric slugs bug 1166429

eviljeff:


diox:
You should use the `mkt.APP_SLUG` constant everywhere instead of copy/pasting

mkt/reviewers/tests/test_views.py
```diff
@@ -3986,7 +3986,7 @@ def setUp(self):
         self.grant_permission(self.profile, 'Apps:ModerateReview')
         self.create_switch('reviews-translate')
         user = user_factory(email='diego')
-        app = app_factory(app_slug='myapp')
+        app = app_factory(app_slug='myapp~@[]{},.#;:')
```
diox: Actually surprised to see this is a valid slug but eh :)

diox:
r+wc

mkt/reviewers/tests/test_views.py
eviljeff: I was going for what the regex allowed but actually looks like https://github.com/mozilla/zamboni/blob/master/mkt/site/utils.py#L172 limits it to just ~-_

eviljeff:
I thought maybe there was a reason the API urls didn't use APP_SLUG like the non-API ones, but I see they're the same.

diox:
I think it's just because we are not very consistent (because the code wasn't DRY in the first place). APP_SLUG is more permissive than necessary, but it should not matter much, at least using it is a step in the right direction.

PR #3097, opened 2015-05-21
 Schedule smaller chunks to reindex if time limit hit (bug 1162062)

washort:
Previously we caught the soft time limit exception and kept going. This makes smaller chunks and starts new tasks.


mkt/webapps/indexers.py
```diff
@@ -423,11 +424,17 @@ def run_indexing(cls, ids, ES=None, index=None, **kw):
 
         qs = Webapp.with_deleted.no_cache().filter(id__in=ids)
         ES = ES or cls.get_es()
-
         docs = []
-        for obj in list(qs):
```
washort: calling `list()` was a guess at not leaving a queryset open when the limit's hit, but obviously more db access is done in `extract_document` etc.

mkt/webapps/indexers.py
```diff
@@ -423,11 +424,17 @@ def run_indexing(cls, ids, ES=None, index=None, **kw):
...
+        for i, obj in enumerate(qs):
             try:
                 docs.append(cls.extract_document(obj.id, obj=obj))
+            except SoftTimeLimitExceeded:
+                from lib.es.management.commands import reindex
+                # obviously this chunk was too big, maybe try two smaller ones.
+                mid = len(ids) / 2
+                reindex.run_indexing.apply_async(args=[index, cls, ids[:mid]])
```
robhudson: Could this index the docs it has built up currently, and pass the rest off to a new task? No sense throwing all this work away right?

diox:
Shouldn't we also bump the limit for reindex ?

mkt/webapps/indexers.py
washort: The time limit's been hit, so there's no time left to do that. Gotta go out and come back in again.

robhudson: My understanding was the soft time limit was a chance to finish/clean up, but the hard time limit force kills the task.

washort: Right.

PR #3101, opened 2015-05-22
Change 'tested with' fields to a formset tested-on for multiple devices Bug 1156759

eviljeff:


diox:
Typo in your title "fileds" (and you probably want to copy that title to the commit message)

media/css/devreg/reviewers.styl
```diff
@@ -1092,6 +1092,19 @@ table.activity .error {
     }
 }
 
+.review-actions-testedon {
+    .formset {
+        margin-top: 1.5em;
+    }
+    a.add-row {
+        background: url("../../img/icons/news-feed-sprite.png") no-repeat 0 -1444px;
```
diox: Wow, this is a huge sprite. We should try to get rid of it instead of relying on it more, or at least optimize its contents, we only need like 2 images from it, that's a bit ridiculous :)

Can you try to either a) use a separate image b) regenerate the `img/mkt/icons/mkt-reviewer-icons.png` to include what you need, and use that ? Bonus if you also fix `.log-attachments a` to stop using this sprite too

media/js/devreg/init.js
```diff
@@ -74,31 +74,40 @@ $(document).ready(function() {
...
             // Add a new formset row when the button is clicked
             $addRowBtn.on('click', function(evt) {
                 evt.preventDefault();
-                $attachment.formset('addRow');
-                var $added = $attachment.formset('options')['latestRow'];
-                $added.find('[data-fileinput]').fakeFileField();
+                $row.formset('addRow');
+                var $added = $row.formset('options')['latestRow'];
+                var field = $added.find('[data-fileinput]')
+                if (field) field.fakeFileField();
```
diox: Add braces and newline

media/js/devreg/init.js
```diff
@@ -74,31 +74,40 @@ $(document).ready(function() {
...
 
             // Add a new formset row when the button is clicked
             $addRowBtn.on('click', function(evt) {
                 evt.preventDefault();
-                $attachment.formset('addRow');
-                var $added = $attachment.formset('options')['latestRow'];
-                $added.find('[data-fileinput]').fakeFileField();
+                $row.formset('addRow');
+                var $added = $row.formset('options')['latestRow'];
+                var field = $added.find('[data-fileinput]')
```
diox: Add `;`

media/js/devreg/init.js
```diff
@@ -74,31 +74,40 @@ $(document).ready(function() {
     // Better <input type="file" />s
     $('[data-fileinput]').fakeFileField();
 
-    // Create additional forms in the attachment formset
-    var $attachments = $('.review-actions-attachments');
+    var setup_add_row = function($node) {
```
diox: Add small comment explaining what this does

diox:
So you can dynamically add "tested on" rows, but not remove them ? Is that going to be a problem ?

media/css/devreg/reviewers.styl
chuckharmston: Yeah, one of the problems about sprites are that the browser needs to hold the entire decompressed image in memory while the page is loaded, which is height x width bytes, regardless of how well it's compressed. Reducing requests are good, but you have to be pretty careful about how you do it. The best medium will often be in several smaller sprites.

eviljeff: I reduced it down to the 3 (I decided on a different one for adding a new line) icons and deleted the original.  We probably have a *lot* of unused image files in media (different bug)

eviljeff:
re: adding more but no remove -the only problem was the defaults really:  the select field defaulted to a device type which would be used in the message.  I added a 'choose' item at the top with an empty value that will be ignored in the form logic later if selected.  Now it functions like the add attachment function.

media/js/devreg/init.js
```diff
@@ -74,31 +74,42 @@ $(document).ready(function() {
...
             // Add a new formset row when the button is clicked
             $addRowBtn.on('click', function(evt) {
                 evt.preventDefault();
-                $attachment.formset('addRow');
-                var $added = $attachment.formset('options')['latestRow'];
-                $added.find('[data-fileinput]').fakeFileField();
+                $row.formset('addRow');
+                var $added = $row.formset('options')['latestRow'];
+                var field = $added.find('[data-fileinput]');
+                if (field) {
```
diox: That will never be false. The proper way to test it would be to test `field.length`

media/js/devreg/init.js
```diff
@@ -74,31 +74,42 @@ $(document).ready(function() {
...
     }
+
+    var $testedon = $('.review-actions-testedon');
+    if($testedon.length) {
+        setup_add_row($testedon);
+    }
+
+    var $attachments = $('.review-actions-attachments');
+    // Hide attachments from Gaia; it doesn't like <file type="input" />.
+    if($attachments.length && !z.capabilities.gaia) {
```
diox: nit: not your code, but please ad a space between `if` and `(` on this line and the one above.

mkt/reviewers/utils.py
```diff
@@ -122,19 +124,33 @@ def create_note(self, action):
...
+                elif not device and version:
+                    text = '%s with version %s' % (dtype, version)
+                else:
+                    text = dtype
+                if text:
+                    tested_on_text.append(text)
+        if not len(tested_on_text):
+            return ''
+        else:
+            return 'Tested on ' + '; '.join(tested_on_text)
```
diox: It was already the case, but it's a bit annoying that this is not translated

mkt/reviewers/tests/test_views.py
```diff
@@ -90,6 +90,32 @@ def _attachments(self, num):
...
+        """
+        Generate and return data for a management form for `num` tested on
+        platforms.
+        """
+        return {'testedon-TOTAL_FORMS': max(1, num),
+                'testedon-INITIAL_FORMS': 0,
+                'testedon-MAX_NUM_FORMS': 1000}
+
+    def _platforms(self, num, device_types=['Desktop', 'FirefoxOS'],
+                   devices=['PC', 'ZTE Open'], versions=['34', '1.3']):
```
diox: Add some unicode strings to devices, to catch some potential unicode errors. `ZTE Opén` for instance :)

mkt/reviewers/utils.py
eviljeff: it gets added to the end of the review message that gets stored in commbadge as text so it needs to be in English really.

diox:
r+wc and tests passing

PR #3102, opened 2015-05-22
Simple URL parsing for websites (bug 1147915)

robhudson:
@diox Based on the test results I'm thinking we need 2 fields for the URL? One that returns the url unchanged (`raw_url`?) and one that uses the URL for searching? And the serialize should probably return the raw URL. If so, I'll update this next week.

### DIFF
--- a/mkt/websites/indexers.py
+++ b/mkt/websites/indexers.py
@@ -1,8 +1,16 @@
+import re
 from operator import attrgetter
+from urlparse import urlparse
 
 from mkt.search.indexers import BaseIndexer
 from mkt.tags.models import attach_tags
 from mkt.translations.models import attach_trans_dict
+
+
+URL_RE = re.compile(
+    r'^www\.|^m\.|^mobile\.|'  # Remove common subdomains.
+    r'\.com$|\.net$|\.org$|\.\w{2}$'  # Remove common TLDs incl. ccTLDs.
+)
 
 
 class WebsiteIndexer(BaseIndexer):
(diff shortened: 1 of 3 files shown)


diox:
2 fields sounds good, yes

mkt/websites/indexers.py
```diff
@@ -138,3 +145,26 @@ def extract_document(cls, pk=None, obj=None):
...
+
+    @classmethod
+    def strip_url(cls, url):
+        """
+        Remove all unwanted sections of the URL and return a string that will
+        be passed to Elasticsearch.
+
+        E.g. 'https://m.domain.com/topic/' will become 'domain/topic/', which
+        will further get tokenized by Elasticsearch into 'domain' and 'topic'.
+
```
diox: Mention somewhere that it can never be perfect (and it's ok) because we don't use a database of tlds

robhudson:
re-r? @diox

mkt/websites/indexers.py
```diff
@@ -121,7 +128,7 @@ def extract_document(cls, pk=None, obj=None):
         attach_tags([obj], m2m_name='keywords')
 
         attrs = ('created', 'default_locale', 'id', 'icon_hash', 'icon_type',
-                 'is_disabled', 'last_updated', 'modified', 'status', 'url')
```
diox: You could have kept url in there

diox:
r+

mkt/websites/indexers.py
robhudson: Ah, you're right. I'll move it. Thanks.

PR #3107, opened 2015-05-27
don't html-escape collection names (bug 1142301)

washort:
names for apps etc are all `TranslatedField`; since this needs to be used in `<title>` elements etc then we should avoid the cleaned version here too.


ngokevin:
r+

PR #3108, opened 2015-05-27
Convert app lookup to DRF (bug 1161865)

robhudson:


mkt/lookup/serializers.py
```diff
@@ -0,0 +1,21 @@
...
+class LookupPaginationSerializer(CustomPaginationSerializer):
+    results_field = 'results'
+
+
+class AppLookupSerializer(ESAppSerializer):
+    url = serializers.SerializerMethodField('get_app_summary_url')
+    is_escalated = serializers.BooleanField()
+
+    class Meta(ESAppSerializer.Meta):
+        fields = ['id', 'url', 'app_slug', 'name']
```
robhudson: TODO: Add status here and use it in the result list.

mkt/lookup/templates/lookup/includes/app_search.html
```diff
@@ -5,6 +5,7 @@
       {{ _('App Lookup') }}
     </label>
     <input type="search" id="app-search" name="q" autocomplete="off">
+    <input type="hidden" id="app-lang" name="lang" autocomplete="off" value="en-US">
```
robhudson: @diox If you know, can you advise me on how to use the user's locale here (instead of this hard-coded `'en-US'`)?

chuckharmston: It's accessible at `window.navigator.l10n.language`, if the lookup tool uses the same setup as commonplace.

diox: It doesn't. But {{ request.LANG }} should work.

robhudson:
@diox I'm thinking making this PR about the DRF conversion, then a follow up PR will add website lookup in a similar fashion. If you agree please review.

mkt/lookup/serializers.py
```diff
@@ -0,0 +1,21 @@
+from rest_framework import serializers
+
+from django.core.urlresolvers import reverse
+
+from mkt.api.paginator import CustomPaginationSerializer
+from mkt.webapps.serializers import ESAppSerializer
+
+
+class LookupPaginationSerializer(CustomPaginationSerializer):
```
diox: You don't need that anymore, do you ?

mkt/lookup/urls.py
```diff
@@ -47,7 +47,7 @@
         name='lookup.user_search'),
     url(r'^transaction_search$', views.transaction_search,
         name='lookup.transaction_search'),
-    url(r'^app_search\.json$', views.app_search,
+    url(r'^app_search\.json$', views.AppLookupSearchView.as_view(),
```
diox: nit: we could use the occasion to rename this? the `.json` part is a bit useless/misleading, the first time I had to do something in the lookup tool I thought it was a static file and was very confused for a couple minutes :)

mkt/lookup/views.py
```diff
@@ -426,44 +428,19 @@ def transaction_search(request):
...
-            app['url'] = reverse('lookup.app_summary', args=[app['id']])
-            app['name'] = app['name__localized_string']
-            results.append(app)
+    def get_paginate_by(self, *args, **kwargs):
+        if self.request.GET.get(self.paginate_by_param) == 'max':
+            return self.max_paginate_by
         else:
-            # This is a result from elasticsearch which returns `Result`
-            # objects and "name_translations" as a list, one for each locale.
-            for trans in app.name_translations:
```
diox: So happy we get to remove this mess (and the hacks to use/non use ES)

mkt/webapps/indexers.py
```diff
@@ -258,9 +259,9 @@ def extract_document(cls, pk=None, obj=None):
             status = None
 
         attrs = ('app_slug', 'bayesian_rating', 'created', 'default_locale',
-                 'icon_hash', 'id', 'is_disabled', 'is_offline', 'file_size',
-                 'last_updated', 'modified', 'premium_type', 'status',
-                 'uses_flash')
+                 'guid', 'icon_hash', 'id', 'is_disabled', 'is_offline',
```
diox: Can you also add `guid` to `ESAppSerializer.fake_object(), in the first `self._attach_fields()` call ` ? Otherwise it won't be available to the serializer ; Not a big deal since we don't use it, but we might ?

diox:
Works great for me locally. r+wc.

mkt/lookup/serializers.py
robhudson: Nope, thanks.

mkt/lookup/urls.py
robhudson: Yep, I didn't like it when I wrote it. I'll strip the `.json` part off.

robhudson:
Updated to address comments and merged:
https://github.com/mozilla/zamboni/commit/1ed1e3b3bbfdea0dcff600cfd814ef2583b85ad2

PR #3109, opened 2015-05-28
Implement website keywords in ES and the API (bug 1168817)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1168817

Refactor some code along the way that we need to share between webapps and websites.

### DIFF
--- a/mkt/tags/models.py
+++ b/mkt/tags/models.py
@@ -3,6 +3,7 @@ from django.core.urlresolvers import NoReverseMatch, reverse
 
 import mkt
 from mkt.site.models import ManagerBase, ModelBase
+from mkt.site.utils import sorted_groupby
 
 
 class TagManager(ManagerBase):
@@ -16,8 +17,6 @@ class Tag(ModelBase):
     tag_text = models.CharField(max_length=128)
     blocked = models.BooleanField(default=False)
     restricted = models.BooleanField(default=False)
-    addons = models.ManyToManyField('webapps.Webapp', through='AddonTag',
-                                    related_name='tags')
 
     objects = TagManager()
 
@@ -57,3 +56,21 @@ class AddonTag(ModelBase):
 
     class Meta:
         db_table = 'users_tags_addons'
+
+
+def attach_tags(objs, m2m_name):
+    """
+    Fetch tags from `objs` in one query and then attach them to a property on
+    each instance. The name of the property will be `m2m_name` + '_list'.
+
+    Assumes every instance in `objs` uses the same model, and needs `m2m_name`
+    to be set to the name of the m2m field between the model and Tag.
+    """
+    if objs:
+        obj_dict = {obj.id: obj for obj in objs}
+        field_name = getattr(objs[0], m2m_name).query_field_name
+        qs = (Tag.objects.not_blocked()
+              .filter(**{'%s__in' % field_name: obj_dict.keys()})
+              .values_list('%s__id' % field_name, 'tag_text'))
+        for obj, tags in sorted_groupby(qs, lambda x: x[0]):
+            setattr(obj_dict[obj], '%s_list' % m2m_name, [t[1] for t in tags])
--- a/mkt/tags/utils.py
+++ b/mkt/tags/utils.py
@@ -1,0 +1,58 @@
+from django import forms
+
+from tower import ugettext as _, ungettext as ngettext
+
+import mkt
+from mkt.access import acl
+from mkt.site.utils import slugify
+from mkt.tags.models import Tag
+
+
+def clean_tags(request, tags):
+    target = [slugify(t, spaces=True, lower=True) for t in tags.split(',')]
+    target = set(filter(None, target))
+
+    min_len = mkt.MIN_TAG_LENGTH
+    max_len = Tag._meta.get_field('tag_text').max_length
+    max_tags = mkt.MAX_TAGS
+    total = len(target)
+
+    blocked = (Tag.objects.values_list('tag_text', flat=True)
+               .filter(tag_text__in=target, blocked=True))
+    if blocked:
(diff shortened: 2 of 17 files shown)


mkt/tags/utils.py
```diff
@@ -0,0 +1,58 @@
...
+                       len(blocked)).format(', '.join(blocked))
+        raise forms.ValidationError(msg)
+
+    restricted = (Tag.objects.values_list('tag_text', flat=True)
+                     .filter(tag_text__in=target, restricted=True))
+    if not acl.action_allowed(request, 'Apps', 'Edit'):
+        if restricted:
+            # L10n: {0} is a single tag or a comma-separated list of tags.
+            msg = ngettext('"{0}" is a reserved tag and cannot be used.',
+                           '"{0}" are reserved tags and cannot be used.',
```
robhudson: These should probably be `u''` strings.

mkt/tags/utils.py
```diff
@@ -0,0 +1,58 @@
...
+    min_len = mkt.MIN_TAG_LENGTH
+    max_len = Tag._meta.get_field('tag_text').max_length
+    max_tags = mkt.MAX_TAGS
+    total = len(target)
+
+    blocked = (Tag.objects.values_list('tag_text', flat=True)
+               .filter(tag_text__in=target, blocked=True))
+    if blocked:
+        # L10n: {0} is a single tag or a comma-separated list of tags.
+        msg = ngettext('Invalid tag: {0}', 'Invalid tags: {0}',
```
robhudson: `u''` here?

robhudson:
This is awesome, nice work. r+wc

mkt/tags/models.py
```diff
@@ -57,3 +56,23 @@ class AddonTag(ModelBase):
...
+
+def attach_tags(objs, m2m_name=None):
+    """
+    Fetch tags from `objs` in one query and then attach them to a property on
+    each instance. The name of the property will be `m2m_name` + '_list'.
+
+    Assumes every instance in `objs` uses the same model, and needs `m2m_name`
+    to be set to the name of the m2m field between the model and Tag.
+    """
+    if m2m_name is None:
```
mstriemer: Do you still need this check if you remove the default value from `m2m_name`?

mkt/tags/models.py
```diff
@@ -57,3 +56,23 @@ class AddonTag(ModelBase):
...
+        raise ValueError('m2m_name needs to be set.')
+    if objs:
+        obj_dict = {obj.id: obj for obj in objs}
+        field_name = getattr(objs[0], m2m_name).query_field_name
+        qs = (Tag.objects.not_blocked()
+              .filter(**{'%s__in' % field_name: obj_dict})
```
mstriemer: This string formatting is a bit hairy but I'm not sure I would change it. Would be be better to pass `obj_dict.keys()` here? I'm assuming it ends up getting iterated over and it works as a `dict` but it seems clearer.

diox: Good idea, I'll try that.

mkt/tags/models.py
diox: No, but I hate functions with more than one non-keyword arguments :)

mstriemer: It's still non-keyword though. It can be called with two positional or keyword arguments.

diox: Yes, but without hacks there is nothing I can do about it (would need https://www.python.org/dev/peps/pep-3102/ )

mstriemer: This code won't catch the `attach_tags(webapps, 'tags')` call so I don't really see the point. It just makes me think `m2m_name` is going to default so something common like `'tags'` by looking at the signature. Is it worth adding an `arity` decorator?

```python
def arity(n):
    def wrapper(fn):
        def inner(*args, **kwargs):
            if len(args) != n:
                raise ValueError(
                    '{fn} takes exactly {n} positional arguments'.format(
                        fn=fn.__name__, n=n))
            else:
                return fn(*args, **kwargs)
        return inner
    return wrapper

@arity(2)
def foo(arg1, arg2, kwarg1, kwarg2):
    return arg1, arg2, kwarg1, kwarg2
```
```bash
>>> from arity import foo
>>> print(foo(1, 2, kwarg1=3, kwarg2=4))
(1, 2, 3, 4)
>>> print(foo(1, 2, 3, 4))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "arity.py", line 7, in inner
    fn=fn.__name__, n=n))
ValueError: foo takes exactly 2 positional arguments
>>> foo(1, 2, kwarg1=3, kwarg2=4, kwarg3=5)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "arity.py", line 9, in inner
    return fn(*args, **kwargs)
TypeError: foo() got an unexpected keyword argument 'kwarg3'
```

This won't handle a `*args` case but if we just want to restrict to one position arg it does the trick.

diox: But why ? It's annoying that we accept it as a positional argument but ultimately it's not such a big deal, since it works anyway. I made it a keyword argument because I preferred to stay as explicit as I could, but there is no need to enforce that the caller is always using it like I intended, IMHO.

mstriemer: Setting a default doesn't make something a keyword argument. So I don't see the point in setting it to default `None` just to raise an error if it isn't set.

It isn't a big deal, I just figured if we want to stick to the one positional argument thing a decorator to enforce it is easy enough to write.

diox: I don't think we want to really enforce that (too much old code that would be inconsistent with that to make it a rule and not enough code using the decorator to really justify it, IMHO).

But you convinced me, I'll remove the check and the default value :)

PR #3110, opened 2015-05-28
Abuse report viewer fixes (bug 1165881 & more)

eviljeff:
fixes bug 1165881 and makes it work with website abuse reports (hadn't landed for the first patch)

### DIFF
--- a/mkt/reviewers/tests/test_views.py
+++ b/mkt/reviewers/tests/test_views.py
@@ -55,6 +55,7 @@ from mkt.versions.models import Version
 from mkt.webapps.models import AddonDeviceType, Webapp
 from mkt.webapps.tasks import unindex_webapps
 from mkt.webapps.tests.test_models import PackagedFilesMixin
+from mkt.websites.utils import website_factory
 from mkt.zadmin.models import get_config, set_config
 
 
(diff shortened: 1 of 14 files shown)


eviljeff:
no idea why the test is failing - the failure is on the delete() itself

diox:
We need to adjust the ES mock in our base `TestCase`. Currently it patches:
```
 ES_patchers = [mock.patch('elasticsearch.Elasticsearch'),
                mock.patch('mkt.webapps.indexers.WebappIndexer', spec=True),
                mock.patch('mkt.search.indexers.index', spec=True),
                mock.patch('mkt.search.indexers.Reindexing', spec=True,
                           side_effect=lambda i: [i])]
```

We should either add `'mkt.websites.indexers.WebsiteIndexer'` to that list (easiest) or change the first line to patch `'mkt.search.indexers.BaseIndexer'` instead (cleanest, but not sure it'd work - might break other tests, needs to be tested)

mkt/reviewers/tests/test_views.py
```diff
@@ -3392,6 +3393,101 @@ def test_deleted_app(self):
         eq_(doc('.tabnav li a')[0].text, u'Abuse Reports (1)')
 
 
+class TestWebsiteAbuseQueue(mkt.site.tests.TestCase, AccessMixin):
```
diox: I didn't think about that when you did the app abuse viewer, but can you add XSS tests ? To make sure we are properly escaping the data displayed in the queue etc.

diox:
r+wc

PR #3115, opened 2015-06-03
Do fewer queries to fetch popularity/trending values when indexing (bug 1170625)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1170625

r? @robhudson

### DIFF
--- a/mkt/search/indexers.py
+++ b/mkt/search/indexers.py
@@ -11,7 +11,8 @@ from elasticsearch_dsl import Search
 import mkt
 from lib.es.models import Reindexing
 from lib.post_request_task.task import task as post_request_task
-from mkt.search.utils import get_boost, get_popularity, get_trending
+from mkt.constants.regions import MATURE_REGION_IDS
+from mkt.search.utils import get_boost
 from mkt.site.decorators import write
 from mkt.translations.utils import to_language
 
@@ -394,8 +395,11 @@ class BaseIndexer(object):
         new_properties['trending'] = {
             'type': 'long', 'doc_values': True
         }
-        # Add region-specific popularity and trending fields.
-        for region in mkt.regions.ALL_REGION_IDS:
+        # Add region-specific popularity / trending fields for mature regions.
+        # We don't have to store anything for adolescent regions, as the
+        # sorting code in SortingFilter will not try to sort using the regional
+        # value if the region in the request is adolescent.
+        for region in MATURE_REGION_IDS:
             new_properties['popularity_%s' % region] = {
                 'type': 'long', 'doc_values': True
             }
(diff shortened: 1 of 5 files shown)


mkt/search/indexers.py
```diff
@@ -394,8 +395,11 @@ def attach_trending_and_popularity_mappings(cls, mapping):
         new_properties['trending'] = {
             'type': 'long', 'doc_values': True
         }
-        # Add region-specific popularity and trending fields.
-        for region in mkt.regions.ALL_REGION_IDS:
+        # Add region-specific popularity / trending fields for mature regions.
+        # We don't have to store anything for adolescent regions, as the
+        # sorting code in SortingFilter will not try to sort using the regional
+        # value if the region in the request is adolescent.
```
robhudson: ... *if* the request is ...

robhudson: Also, excellent thinking. This will help keep these region fields to a minimum.

mkt/search/indexers.py
```diff
@@ -408,14 +412,29 @@ def attach_trending_and_popularity_mappings(cls, mapping):
...
         }
-        for region in mkt.regions.ALL_REGIONS:
-            extend['trending_%s' % region.id] = get_trending(obj, region)
-            extend['popularity_%s' % region.id] = get_popularity(obj, region)
+        trending = get_dict(obj, 'trending')
+        popularity = get_dict(obj, 'popularity')
+
+        # Global popularity.
+        extend['trending'] = popularity[RESTOFWORLD.id]
+        extend['popularity'] = trending[RESTOFWORLD.id]
```
robhudson: Are these backwards?

diox: wooops

robhudson:
So you're not changing how we call `get_popularity`? But with these changes we only query 5 or so regions rather than 250. I'm ok with that.

diox:
I'm doing both, actually. Not relying on `get_popularity()` anymore, just fetching popularity in one query for all mature regions, and only caring about mature regions.

mkt/search/indexers.py
```diff
@@ -408,14 +412,29 @@ def attach_trending_and_popularity_mappings(cls, mapping):
 
     @classmethod
     def extract_popularity_trending_boost(cls, obj):
+        def get_dict(obj, prop):
+            qs = getattr(obj, prop).filter(
+                region__in=MATURE_REGION_IDS + (RESTOFWORLD.id))
+            return dict(qs.values_list('region', 'value'))
```
robhudson: Ah, here's the query code I overlooked. Is it written like this so it will work for both `Webapp` and `Website`?

mkt/search/indexers.py
```diff
@@ -408,14 +412,29 @@ def attach_trending_and_popularity_mappings(cls, mapping):
 
     @classmethod
     def extract_popularity_trending_boost(cls, obj):
+        def get_dict(obj, prop):
+            qs = getattr(obj, prop).filter(
+                region__in=MATURE_REGION_IDS + (RESTOFWORLD.id))
```
robhudson: missing comma on that tuple?

robhudson:
r+wc

mkt/search/indexers.py
diox: yeah, and also to work for both popularity and trending

PR #3117, opened 2015-06-03
Include app state in lookup tool (bug 1166528)

robhudson:


mkt/lookup/serializers.py
```diff
@@ -2,14 +2,22 @@
...
     class Meta(ESAppSerializer.Meta):
-        fields = ['id', 'url', 'app_slug', 'name']
+        fields = ['app_slug', 'id', 'name', 'status', 'url']
 
     def get_app_summary_url(self, obj):
         return reverse('lookup.app_summary', args=[obj.id])
+
+    def get_app_status(self, obj):
+        status = mkt.STATUS_CHOICES_API_v2[obj.status]
+        if obj.status != mkt.STATUS_DISABLED and obj._is_disabled:
```
eviljeff: I'm pretty sure developers can't disable their app after we've Banned it (or make any changes), so this couldn't happen.  Harmless though.

mkt/lookup/tests/test_views.py
```diff
@@ -660,6 +660,8 @@ def verify_result(self, data):
         eq_(data['objects'][0]['url'], reverse('lookup.app_summary',
                                                args=[self.app.pk]))
         eq_(data['objects'][0]['app_slug'], self.app.app_slug)
+        eq_(data['objects'][0]['status'],
+            mkt.STATUS_CHOICES_API[self.app.status])
```
eviljeff: should this be API_V2 ?

eviljeff:
r+wc

mkt/lookup/tests/test_views.py
robhudson: ah yeah, thanks. Probably ok here b/c it's public but better to have it right.

mkt/lookup/serializers.py
robhudson: This is there b/c we only store `is_disabled` in elasticsearch, which comes from a method on the app object and is true when it is either `STATUS_DISABLED` or `disabled_by_user`. So this conditional is needed to pull it back apart and only return `'disabled'` if it's `disabled_by_user`.

eviljeff: Can we add a comment explaining?  (Took me some thought to get my head around it)
Also, nit - the lookup on #20 is redundant if the if.. is true so should be in an else instead I guess.

PR #3119, opened 2015-06-04
Implement script to import e.me data (bug 1161871)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1161871

Goes with https://docs.google.com/spreadsheets/d/1CPrPQ4xVsbll_vp_Bca0iWkTgg2nx7WrhCprVHwS-6Y/edit#gid=304044273

Depends on #3090 (some rows in that spreadsheet reference the new categories, the code expect the new names to be present)

### DIFF
--- a/migrations/912-websites-add-moz-id.sql
+++ b/migrations/912-websites-add-moz-id.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `websites_website` ADD COLUMN `moz_id` integer UNSIGNED UNIQUE;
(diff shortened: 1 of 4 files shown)


mkt/constants/categories.py
```diff
@@ -2,24 +2,31 @@
 
 
 CATEGORY_CHOICES = (
```
diox: These changes will go away as soon as #3090 is merged.

robhudson:
r+ I couldn't find anything after reading closely. Nice work. :+1:

PR #3120, opened 2015-06-04
Add option to reindex every website (bug 1161871)

diox:


lib/es/management/commands/reindex.py
```diff
@@ -194,7 +195,11 @@ def handle(self, *args, **kwargs):
 
         if index_choice:
             # If we only want to reindex a subset of indexes.
-            INDEXES = INDEX_DICT.get(index_choice, INDEXES)
+            INDEXES = INDEX_DICT.get(index_choice, None)
```
mstriemer: Assigning this to `INDEXES` worries me but I guess it works.

mstriemer: What if you want to index everything? This removes support for that?

mstriemer: Nevermind, only if you provide an index. Makes sense to me!

mstriemer:
Not crazy about reassigning `INDEXES` would rather see `indexes = INDEXES` instead of `global INDEXES` but since this is a management command I guess it doesn't matter and it's clearly been working.

r+

diox:
Not crazy about the code either, but I didn't want to start a refactor just to add this.

robhudson:
I added this also in a branch I have so :+1:

PR #3121, opened 2015-06-04
Add lookup tool for websites (bug 1161865)

robhudson:
screenshot 2015-06-04 14 05 09

### DIFF
--- a/migrations/913-websites-lookup-perms.sql
+++ b/migrations/913-websites-lookup-perms.sql
@@ -1,0 +1,3 @@
+-- Changing Websites:Review to Websites:* and removing duplicate Stats:View.
+UPDATE groups SET rules='Apps:*,Websites:*,Users:Edit,Stats:View,AdminTools:View,AccountLookup:View,AppLookup:View,Lookup:View' WHERE name='Staff';
+UPDATE groups SET rules=CONCAT(rules, ',WebsiteLookup:View') WHERE name IN ('Staff', 'Support Staff', 'Carriers and Operators');
--- a/mkt/lookup/helpers.py
+++ b/mkt/lookup/helpers.py
@@ -15,7 +15,6 @@ def format_currencies(context, currencies):
     return jinja2.Markup(cs)
 
 
-# page_type is used for setting the link 'sel' class (activity/purchases)
 @register.function
 def user_header(account, title, is_admin=False, page_type=''):
     t = env.get_template('lookup/helpers/user_header.html')
@@ -23,7 +22,6 @@ def user_header(account, title, is_admin=False, page_type=''):
                                    'title': title, 'page_type': page_type}))
 
 
-# page_type is used for setting the link 'sel' class
 @register.function
 @jinja2.contextfunction
 def app_header(context, app, page_type=''):
@@ -39,6 +37,15 @@ def app_header(context, app, page_type=''):
                                    'is_reviewer': is_reviewer,
                                    'is_author': is_author,
                                    'is_operator': is_operator}))
+
+
+@register.function
+@jinja2.contextfunction
+def website_header(context, website, page_type=''):
+    t = env.get_template('lookup/helpers/website_header.html')
+
+    return jinja2.Markup(t.render({'website': website,
+                                   'page_type': page_type}))
 
 
 @register.function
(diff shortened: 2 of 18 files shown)


migrations/912-websites-lookup-perms.sql
```diff
@@ -0,0 +1 @@
+UPDATE groups SET rules=CONCAT(rules, ',WebsiteLookup:View') WHERE name IN ('Staff', 'Support Staff', 'Carriers and Operators');
```
diox: I just stole this migration number

mkt/lookup/helpers.py
```diff
@@ -41,6 +41,20 @@ def app_header(context, app, page_type=''):
                                    'is_operator': is_operator}))
 
 
+# page_type is used for setting the link 'sel' class
```
diox: docstring instead?

mkt/lookup/views.py
```diff
@@ -315,6 +318,18 @@ def app_summary(request, addon_id):
 
 
 @login_required
+@permission_required([('WebsiteLookup', 'View')])
+def website_summary(request, addon_id):
+    website = get_object_or_404(Website, pk=addon_id)
+    if not hasattr(website, 'keywords_list'):
```
diox: mmm, it's unfortunate that you have to do this, since we do this already in the serializer.

We should use attach_tags as a transformer in a custom manager method like `Website.objects.with_tags()` and use that method in this view and the API view... Maybe worth filing a code quality bug for that

migrations/912-websites-lookup-perms.sql
robhudson: :rage:

mkt/lookup/views.py
robhudson: I could just do that in this PR. I don't think it would take long.

robhudson: It's Fri afternoon and my brain is fried. I'll file a bug.

mkt/lookup/helpers.py
```diff
@@ -41,6 +41,20 @@ def app_header(context, app, page_type=''):
...
+@register.function
+@jinja2.contextfunction
+def website_header(context, website, page_type=''):
+    t = env.get_template('lookup/helpers/website_header.html')
+
+    is_admin = acl.action_allowed(context['request'], 'Users', 'Edit')
+    is_staff = acl.action_allowed(context['request'], 'Apps', 'Configure')
```
eviljeff: depends what is_staff will be used for, but testing for 'Apps:Configure' seems wrong for websites.

robhudson: Yeah, I agree. And I could remove this from the patch for now, it's not being used. In bug 1161867 it'll be used to show the link to edit the data.

migrations/913-websites-lookup-perms.sql
```diff
@@ -0,0 +1 @@
+UPDATE groups SET rules=CONCAT(rules, ',WebsiteLookup:View') WHERE name IN ('Staff', 'Support Staff', 'Carriers and Operators');
```
eviljeff: can you change Staff to: `'Apps:*,Users:Edit,Stats:View,AdminTools:View,AccountLookup:View,AppLookup:View,Lookup:View,Websites:*'`
so we don't have to keep adding new ones.

robhudson: @eviljeff ^

eviljeff: :+1: 
we probably should have started out doing `Lookup:App` , etc in the beginning imo.

robhudson: Except the pattern is `<Noun>`:`<Verb>`. I don't have a problem with a few extra rules.

mkt/lookup/views.py
robhudson: https://bugzilla.mozilla.org/show_bug.cgi?id=1172575

migrations/913-websites-lookup-perms.sql
eviljeff: `Lookup:ViewApp` :smiley:

mkt/websites/models.py
```diff
@@ -116,6 +118,17 @@ def get_icon_url(self, size):
     def get_url_path(self):
         return reverse('website.detail', kwargs={'pk': self.pk})
 
+    def get_regions(self, sort_by='slug'):
```
diox: I wouldn't call this `get_regions`: this is confusing since the method already exists on webapps, but unlike webapps websites are available in all regions anyway

diox: `get_preferred_regions()` would work fine I guess

diox:
r+wc

PR #3123, opened 2015-06-05
Fix log for apps with unicode urls

robhudson:
I saw this in the logs and thought I'd fix it real quick.


mkt/webapps/tests/test_tasks.py
```diff
@@ -673,7 +673,7 @@ class TestPreGenAPKs(mkt.site.tests.WebappTestCase):
 
     def setUp(self):
         super(TestPreGenAPKs, self).setUp()
-        self.manifest_url = 'http://some-app.net/manifest.webapp'
+        self.manifest_url = u'http://some-âpp.net/manifest.webapp'
```
andymckay: nice

andymckay:
r+

PR #3124, opened 2015-06-09
Work around bug preventing the reindex task to work (bug 1172904)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1172904


jasonthomas:
r+

PR #3125, opened 2015-06-09
Add a "reviewed" field in ES for Websites (bug 1161858)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1161858

This is needed to make the "New" page work.


robhudson:
If there's a websites test that checks the extracted document add that new field to it. Otherwise, r+

diox:
There is, you're right, I should add the field to it.

PR #3126, opened 2015-06-09
Update reindexing tasks to accept strings for the indexer (bug 1172904)

robhudson:


lib/es/management/commands/reindex.py
```diff
@@ -31,22 +31,34 @@
 
 # The subset of settings.ES_INDEXES we are concerned with.
 ES_INDEXES = settings.ES_INDEXES
+
+# We have this mapping so we can pass which indexer to use to celery.
+INDEXER_MAP = {
+    'webapp': WebappIndexer,
```
diox: Can we not hardcode these keys somehow ? I feel like we have too much stuff to modify when adding an index, between settings, this dict, the `INDEXES` tuple below, the indexer itself...

robhudson: Ok. I could use the indexer object's `get_mapping_type_name`?

diox: Yeah, something like that. Maybe even get rid of `INDEXES` and just set the chunk size in each indexer? Almost all of them default to 500 anyway, so you could set a default value in the base indexer and just override it when necessary

robhudson: Ooh yeah, I like that.

diox:
Too late for me to test this locally but it looks good, and since it fixes a complete breakage anyway, r+wc

robhudson:
Updated. re-r?

diox:
r+

PR #3131, opened 2015-06-12
Use pickle for celery for now (bug 1171052)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1171052

Should probably add a test, going to let Travis run a full test though.

### DIFF
--- a/mkt/submit/views.py
+++ b/mkt/submit/views.py
@@ -103,6 +103,7 @@ def manifest(request):
 
         upload = form.cleaned_data['upload']
         addon = Webapp.from_upload(upload, is_packaged=form.is_packaged())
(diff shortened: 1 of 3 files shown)


mkt/submit/views.py
```diff
@@ -136,7 +137,7 @@ def manifest(request):
 
         # Call task outside of `atomic` to avoid it running before
         # the transaction is committed and not finding the app.
```
diox: Can you also remove this comment ? It's obsolete, we are no longer manually using `atomic` here, instead it's a `post_request_task` automatically delayed till the end of the request-response cycle (which is wrapped in a transaction with `ATOMIC_REQUESTS`)

diox:
r+wc

PR #3132, opened 2015-06-12
Change addon tag relation to implicit through model (bug 1174253)

robhudson:


mkt/tags/models.py
```diff
@@ -37,25 +37,23 @@ def can_reverse(self):
     def get_url_path(self):
         return reverse('tags.detail', args=[self.tag_text])
 
-    def save_tag(self, addon):
+    def _get_m2m_name(self, obj):
+        model = obj._meta.model_name
+        if model == 'webapp':
+            return 'tags'
+        elif model == 'website':
+            return 'keywords'
```
robhudson: Curious if there's a smarter way to do this?

amuntner: You
Can Can trotting by jbbbnhi Bingham B in nbg

by nbgg

diox: Maybe. You can access the m2m related to the `Tag` model with `Tag._meta.get_all_related_m2m_objects_with_model()`. From the other side, you can use `<Website|Webapp>._meta.many_to_many` and find the one linking to `Tag`.

diox: also, if you figure this out, you can use that to remove the `m2m_name` parameter that I added to `attach_tags()` below?

robhudson: Curious if it's worth it?

diox: I don't think it's that important. Would be more DRY, but it's not *that* bad. If you don't do it, maybe just add a `else` throwing an exception ? That will help us when we introduce tags to other models in the future.

robhudson: Updated. Thanks for the pointers.

migrations/915-rename-addon-tags.sql
```diff
@@ -0,0 +1 @@
+RENAME TABLE `users_tags_addons` TO `addons_tags`;
```
diox: In theory, this does not require a full copy and is supposed to be almost instantaneous, so we should be ok. Maybe we should make a quick test though ? I assume we have lots of data in there.

mkt/tags/models.py
```diff
@@ -37,25 +37,22 @@ def can_reverse(self):
...
         return reverse('tags.detail', args=[self.tag_text])
 
-    def save_tag(self, addon):
-        tag, created = Tag.objects.get_or_create(tag_text=self.tag_text)
-        AddonTag.objects.get_or_create(addon=addon, tag=tag)
-        mkt.log(mkt.LOG.ADD_TAG, tag, addon)
-        return tag
+    def _get_m2m_name(self, obj):
+        related_models = Tag._meta.get_all_related_m2m_objects_with_model()
+        field_map = dict((rm[0].model, rm[0].field.name)
```
diox: Could use a dict comprehension instead of calling `dict()` here since we are on python 2.7

Also, do you think you could also use this method (making it a `classmethod`) in `attach_tags()` below, removing the need to pass the `m2m_name` and determining it automatically instead ?

migrations/915-rename-addon-tags.sql
robhudson: I was trying to find docs on whether this is a simple rename vs copying data to a new table and found that all FK references also need to be updated. Ugh.

robhudson: Table renames don't require a temporary table.

Also, the constraints are on the table itself, not on the related tables so we should be ok there. There are no constraints on either `tags` or `addons` pointing to `addons_tags`.

robhudson: WRT testing... I'm comfortable with the dev/stage roll-out to test any thing that may break.

mkt/tags/models.py
```diff
@@ -37,28 +37,26 @@ def can_reverse(self):
...
+    def remove_tag(self, obj):
+        for tag in obj.tags.filter(tag_text=self.tag_text):
+            getattr(obj, self._get_m2m_name(obj)).remove(tag)
+        mkt.log(mkt.LOG.REMOVE_TAG, self.tag_text, obj)
 
 
-def attach_tags(objs, m2m_name):
+def attach_tags(objs):
     """
     Fetch tags from `objs` in one query and then attach them to a property on
```
diox: Forgot to tell you, the docstring needs to be updated too

mkt/tags/models.py
```diff
@@ -68,6 +66,7 @@ def attach_tags(objs, m2m_name):
     """
     if objs:
         obj_dict = {obj.id: obj for obj in objs}
+        m2m_name = Tag._get_m2m_name(objs[0])
```
diox: :+1:

diox:
r+wc

migrations/915-rename-addon-tags.sql
robhudson: On -dev this took 0.17 seconds

PR #3135, opened 2015-06-16
Adds helper to guess string language, transform to dict suitable for assignment to a TranslatedField.

chuckharmston:
r? @ngokevin


ngokevin:
r+

PR #3136, opened 2015-06-16
move flake8 to after

andymckay:
If we do the check after, it will still fail the test suite... but you'll get to see all the tests that fail as well, instead of just a flake8 failure.


diox:
I'm not against changing it, but it was done by design, precisely to avoid an useless full test run.

If you do the check after, you need to wait for a full test run, find out there were flake8 errors, fix them, and re-run the full test, that's wasteful and takes a lot more time.

PR #3137, opened 2015-06-16
Update tasks to accept only JSON serializable data (bug 1174219)

robhudson:


lib/video/tasks.py
```diff
@@ -18,8 +20,9 @@
 # Video decoding can take a while, so let's increase these limits.
 @task(time_limit=time_limits['hard'], soft_time_limit=time_limits['soft'])
 @set_modified_on
-def resize_video(src, instance, user=None, **kw):
+def resize_video(src, pk, user=None, **kw):
```
diox: that `user` parameter is probably a not going to work either with JSON serialization

mkt/site/tasks.py
```diff
@@ -40,15 +42,16 @@ def send_email(recipient, subject, message, real_email, from_email=None,
 
 
 @task
-def set_modified_on_object(obj, **kw):
+def set_modified_on_object(app_label, model_name, pk, **kw):
```
diox: pretty sure this doesn't actually need to be a task. Maybe that'd make things simpler if it wasn't?

lib/video/tasks.py
robhudson: Ah, good call out. I'll fix that.

mkt/site/tasks.py
robhudson: I don't think this is bad as it is right now, do you?

diox: No, it's fine. We should have a bug about it, it's worth fixing (would make things simpler and reduce the amount of tasks in the queue)

diox:
r+

PR #3140, opened 2015-06-17
Adds WebsiteSubmission model.

chuckharmston:
Creating PR to get Travis on the case, but if you want to r? @ngokevin, I wouldn't be opposed.

### DIFF
--- a/migrations/917-adds-websitesubmission-model.sql
+++ b/migrations/917-adds-websitesubmission-model.sql
@@ -1,0 +1,35 @@
+CREATE TABLE `websites_websitesubmission_keywords` (
+    `id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY,
+    `website_id` integer UNIQUE,
+    `tag_id` integer unsigned NOT NULL,
+    UNIQUE (`website_id`, `tag_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+ALTER TABLE `websites_websitesubmission_keywords` ADD CONSTRAINT `websites_websitesubmission_keywords_tag_id` FOREIGN KEY (`tag_id`) REFERENCES `tags` (`id`);
+
+CREATE TABLE `websites_websitesubmission` (
+    `id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY,
+    `created` datetime NOT NULL,
+    `modified` datetime NOT NULL,
+    `date_approved` datetime NULL,
+    `name` integer unsigned UNIQUE,
+    `description` integer unsigned UNIQUE,
+    `categories` longtext NOT NULL,
+    `detected_icon` varchar(255) NOT NULL,
+    `icon_type` varchar(25) NOT NULL,
+    `icon_hash` varchar(8) NOT NULL,
+    `url` varchar(255) NOT NULL,
+    `canonical_url` varchar(255),
+    `works_well` integer unsigned UNIQUE,
+    `submitter_id` integer unsigned UNIQUE,
+    `public_credit` bool NOT NULL DEFAULT FALSE,
+    `why_relevant` longtext NOT NULL,
+    `preferred_regions` longtext NOT NULL,
+    `approved` bool NOT NULL DEFAULT FALSE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+ALTER TABLE `websites_websitesubmission` ADD CONSTRAINT `websites_websitesubmission_submitter_id` FOREIGN KEY (`submitter_id`) REFERENCES `users` (`id`);
+ALTER TABLE `websites_websitesubmission` ADD CONSTRAINT `websites_websitesubmission_name` FOREIGN KEY (`name`) REFERENCES `translations` (`id`);
+ALTER TABLE `websites_websitesubmission` ADD CONSTRAINT `websites_websitesubmission_description` FOREIGN KEY (`description`) REFERENCES `translations` (`id`);
+
+ALTER TABLE `websites_websitesubmission_keywords` ADD CONSTRAINT `websites_websitesubmission_websitesubmission_id` FOREIGN KEY (`websitesubmission_id`) REFERENCES `websites_websitesubmission` (`id`);
--- a/mkt/websites/models.py
+++ b/mkt/websites/models.py
@@ -11,13 +11,88 @@ from django_extensions.db.fields.json import JSONField
 
 import mkt
 from lib.utils import static_url
+from mkt.api.fields import IntegerRangeField
 from mkt.constants.applications import DEVICE_TYPE_LIST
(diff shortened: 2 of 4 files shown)


mkt/websites/models.py
```diff
@@ -11,15 +11,98 @@
...
+    """
+    Model representing a website submission.
+
+    When accepted, the data from this will be copied to a new instance of the
+    Website model, and accepted=True will be set on the submission instance.
+    """
+    name = TranslatedField()
+    keywords = models.ManyToManyField(Tag)
+    description = TranslatedField()
+    categories = JSONField(default=None)
```
ngokevin: how will this work? i'm currently expecting 2 categories

mkt/websites/models.py
```diff
@@ -11,15 +11,98 @@
...
+
+    # Why does the user think the website is relevant to Marketplace?
+    why_relevant = models.TextField()
+
+    # If the submitter says the website is relevant worldwide,
+    # preferred_regions should be set to [].
+    preferred_regions = JSONField(default=None)
+
+    # Turn true when a reviewer has accepted and published a submission.
+    accepted = models.BooleanField(default=False)
```
ngokevin: maybe ```approved```?

ngokevin:
looks good on skim

mkt/websites/models.py
chuckharmston: This mirrors the `Website` model's `categories` field.

mkt/websites/models.py
mstriemer: There's also a field called `accepted` already, L52 it's a `DateTimeField`.

mkt/websites/models.py
```diff
@@ -11,15 +11,98 @@
...
+    # A 1-5 Likert scale indicating how well the submitter says the website
+    # works. This is primarily used to make the submitter question the
+    # relevance of their submission, though submissions should be prevented if
+    # the user says it works poorly.
+    works_well = IntegerRangeField(min_value=2, max_value=5)
+
+    # Who is submitting the website? Do they want public credit for their
+    # submission?
+    submitter = models.ForeignKey(UserProfile,
+                                  related_name='website_submitter')
```
mstriemer: `website_submitter` doesn't seem like a good `related_name`. Wouldn't that be: `some_user.website_submitter.all()  # => all websites this user submitted`

chuckharmston: Agreed. Good call.

mkt/websites/models.py
```diff
@@ -11,15 +11,90 @@
...
+    # works. This is primarily used to make the submitter question the
+    # relevance of their submission, though submissions should be prevented if
+    # the user says it works poorly.
+    works_well = IntegerRangeField(min_value=2, max_value=5)
+
+    # Who is submitting the website? Do they want public credit for their
+    # submission?
+    submitter = models.ForeignKey(UserProfile,
+                                  related_name='websites_submitted')
+    public_credit = models.BooleanField(default=False)
```
mstriemer: I'm not sure how this is supposed to work with the same website being submitted by multiple users. I guess at some point there be a many-to-one from `WebsiteSubmission` to `Website`. Not expecting a change about it and this could be addressed in the PRD but thought I'd mention it.

chuckharmston: The `approved` field will likely be converted to a `status` field, TBD. For now, we're accepting all submissions, as they might contain metadata in other languages that we might want to crib.

migrations/917-adds-websitesubmission-model.sql
```diff
@@ -0,0 +1,35 @@
...
+    `why_relevant` longtext NOT NULL,
+    `preferred_regions` longtext NOT NULL,
+    `approved` bool NOT NULL DEFAULT FALSE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+ALTER TABLE `websites_websitesubmission` ADD CONSTRAINT `websites_websitesubmission_submitter_id` FOREIGN KEY (`submitter_id`) REFERENCES `users` (`id`);
+ALTER TABLE `websites_websitesubmission` ADD CONSTRAINT `websites_websitesubmission_name` FOREIGN KEY (`name`) REFERENCES `translations` (`id`);
+ALTER TABLE `websites_websitesubmission` ADD CONSTRAINT `websites_websitesubmission_description` FOREIGN KEY (`description`) REFERENCES `translations` (`id`);
+
+ALTER TABLE `websites_websitesubmission_keywords` ADD CONSTRAINT `websites_websitesubmission_websitesubmission_id` FOREIGN KEY (`websitesubmission_id`) REFERENCES `websites_websitesubmission` (`id`);
```
robhudson: The `websites_websitesubmission_keywords`.`website_id` is unsigned but `websites_websitesubmission`.`id` is not. Won't that produce an error in mysql?

mkt/websites/models.py
```diff
@@ -11,15 +11,90 @@
...
+
+    objects = WebsiteSubmissionManager()
+
+    class Meta:
+        ordering = (('-modified'), )
+
+    def __unicode__(self):
+        return unicode(self.url)
+
+
```
robhudson: I'm curious why we don't follow the `Webapp` approach to this and use the `status` fields as needed to denote a new submission and approval? I think this is going to lead to a headache as we add fields to the `Website` model and we have to add them here also. Or we forget and they get out of sync.

chuckharmston: Because the process is very different.

For apps, we're gathering *all* necessary data from the app developers, and the reviewers' job is to review that information and publish. For sites, we're asking the submitters to submit a URL and some basic metadata. The reviewers are reviewing that data, and supplementing it with additional fields before publishing. In future phases, this process will likely become more complex (the original vision involved 4 different roles, each responsible for a different set of operations).

We _could_ use statuses and complex validation rules for each, but that would be a lot more complex.

robhudson: Would it be as complex if we stored the website info in the website table and reviewers operated on that while in a non-approved state. And all the meta data would be in this submission table? Going forward that seems clearer and easier to maintain IMO.

chuckharmston: We'd need to `NOT NULL` most fields on the `Website` model then, and maintain a complex set of validation rules based on the state of the review. Do you think that's easier to maintain?

chuckharmston: That'd also be much more brittle to process changes (which are inevitable; submission/review is an MVP for phase 0).

ngokevin: I had the same question as @robhudson at first. Having a WebsiteSubmission model extending from Website made some sense. But given that the process for websites will involve multiple reviewer steps and them supplementing additional metadata, it would be a lot simpler to have a simple, flexible website submission model (at least for an MVP). There will be some data duplication between the models, but it may make things a bit simpler.

ngokevin:
r+ for me

PR #3141, opened 2015-06-18
Drop date fields from addons_tags (bug 1174253)

robhudson:


mstriemer:
Hit an issue on migration 916 locally, but after manually fixing that this one works.

r+

PR #3142, opened 2015-06-19
Add website edit form to lookup tool (bug 1161867)

robhudson:


robhudson:
Ignore the black bar... the built-in developer toolbar screenshot keeps adding that for some reason.
screen shot 2015-06-19 at 12 51 47

mkt/lookup/views.py
```diff
@@ -39,19 +39,19 @@
...
 from mkt.webapps.models import Webapp
 from mkt.websites.models import Website
+from mkt.websites.forms import WebsiteForm
 from mkt.websites.views import WebsiteSearchView
 
 
 log = commonware.log.getLogger('z.lookup')
 
 
-@login_required
```
eviljeff: not needed when permissions are asked for?

robhudson: Right, `@permission_required` does `@login_required` already.

mkt/lookup/views.py
```diff
@@ -329,7 +323,23 @@ def website_summary(request, addon_id):
     })
 
 
-@login_required
+@permission_required([('WebsiteLookup', 'View')])
+def website_edit(request, addon_id):
```
eviljeff: addon_id :cry:

robhudson: heh, yeah. I guess I could change it. But everything else passes around `addon_id`s.

mkt/websites/forms.py
```diff
@@ -0,0 +1,85 @@
...
+            regions = map(int, self.cleaned_data.get('preferred_regions'))
+        except (TypeError, ValueError):
+            # Data is not a list or data contains non-integers.
+            raise forms.ValidationError(_('Invalid region(s) selected.'))
+
+        return list(regions)
+
+    def save(self, commit=False):
+        form = super(WebsiteForm, self).save(commit=False)
+        form.save()
```
eviljeff: this looks... weird.

mkt/lookup/tests/test_views.py
```diff
@@ -1211,3 +1212,22 @@ def test_by_name(self):
     def test_by_id(self):
         data = self.search(q=self.website.pk)
         self.verify_result(data)
+
+
+class TestWebsiteEdit(mkt.site.tests.TestCase):
+    fixtures = fixture('user_support_staff', 'user_999')
```
eviljeff: user_999 seems unused.

eviljeff:
r+wc but I'm fairly unfamiliar with the MOW implementation so far.

mkt/websites/forms.py
robhudson: Hmm, yes it does. I'll double check this, thanks.

mkt/lookup/tests/test_views.py
```diff
@@ -1211,3 +1212,22 @@ def test_by_name(self):
     def test_by_id(self):
         data = self.search(q=self.website.pk)
         self.verify_result(data)
+
+
+class TestWebsiteEdit(mkt.site.tests.TestCase):
```
diox: Could/should we have more view tests ?

robhudson: The view is actually pretty empty, most of the logic is in the form, which is also where the tests are. But if you have something specific I'd be happy to add it. Perhaps just a simple test that a form submission redirects and updates the data as a view test?

diox: Yeah, simple test that loads and then posts the form should be enough

diox:
r+wc once the view tests pass

robhudson:
Ha, yep. I'm rebuilding my docker to run the tests locally. I'll merge once the test is good. Thanks for the review.

mkt/websites/forms.py
```diff
@@ -0,0 +1,88 @@
...
+                               widget=forms.Textarea(attrs={'rows': 2}))
+    name = TransField(label=_lazy(u'Name'), widget=TransInput())
+    preferred_regions = forms.MultipleChoiceField(
+        label=_lazy(u'Preferred Regions'), choices=REGIONS_CHOICES_NAME,
+        required=False, widget=forms.SelectMultiple(attrs={'size': 10}))
+    short_name = TransField(label=_lazy(u'Short Name'), widget=TransInput(),
+                            required=False)
+    title = TransField(label=_lazy(u'Title'), widget=TransInput(),
+                       required=False)
+    url = forms.URLField(label=_lazy(u'URL'))
```
diox: Are you having this field just because you want it to be required ? Maybe I should not have allowed blanks in the model...

robhudson: If it's definitely required then yeah, we probably want to not allow it to be empty at the model level too. Should we do that here or in a follow up commit?

robhudson:
https://github.com/mozilla/zamboni/commit/e3818e9

PR #3143, opened 2015-06-19
Updates installation documentation for virtualenvburrito.

chuckharmston:
r?


diox:
r+ (but I wish we didn't use the `curl | shell` pattern in our docs...)

chuckharmston:
I do not disagree. :disappointed:

PR #3148, opened 2015-06-20
Adds an API endpoint to scrape a URL and return metadata.

chuckharmston:
WIP, please don't merge.

Would love some feedback from smart Python folks on a few items I'll specifically comment on. I wouldn't turn down any other feedback, as well.

r? @robhudson @ngokevin @diox

### DIFF
--- a/mkt/websites/tests/test_helpers.py
+++ b/mkt/websites/tests/test_helpers.py
@@ -1,0 +1,253 @@
+import random
+
+import mock
+import requests
+import responses
+from nose.tools import eq_, ok_
+
+from mkt.site.tests import TestCase
+from mkt.websites.helpers import WebsiteMetadata
+
+
+class TestWebsiteMetadataMixin(object):
+    url = 'https://mobile.piedpiper.com'
+
+    @responses.activate
+    def get_obj(self, response_body):
+        responses.add(responses.GET, self.url, body=response_body, status=200)
+        return WebsiteMetadata(self.url)
+
+    def get_obj_with(self, response_body):
+        return self.get_obj('<html>%s</html>' % response_body)
+
+    def get_empty_obj(self):
+        return self.get_obj_with('')
+
+
+class TestWebsiteMetadata(TestWebsiteMetadataMixin, TestCase):
+    def test_check_ms_icon_size_nomatch(self):
+        eq_(self.get_empty_obj()._check_ms_icon_size(310), None)
+
+    def test_check_ms_icon_size_match(self):
+        VALUE = 'http://i.imgur.com/Ng0I5UA.gif'
+        test = ('<meta name="msapplication-square310x310logo" '
+                'content="%s">') % VALUE
+        eq_(self.get_obj_with(test)._check_ms_icon_size(310), VALUE)
+
+    def test_check_apple_icon_size_nomatch(self):
+        eq_(self.get_empty_obj()._check_apple_icon_size(152), None)
+
+    def test_check_apple_icon_size_match(self):
+        VALUE = 'http://i.imgur.com/Ng0I5UA.gif'
+        test = '<link rel="apple-touch-icon-precomposed" href="%s">' % VALUE
+        eq_(self.get_obj_with(test)._check_apple_icon_size(None), VALUE)
+
+    def test_check_apple_icon_size_with_sizes_match(self):
+        VALUE = 'http://i.imgur.com/Ng0I5UA.gif'
+        test = ('<link rel="apple-touch-icon-precomposed" sizes="152x152" '
+                'href="%s">') % VALUE
+        eq_(self.get_obj_with(test)._check_apple_icon_size(152), VALUE)
+
+    def test_check_opengraph_nomatch(self):
+        eq_(self.get_empty_obj()._check_opengraph('url'), None)
+
+    def test_check_opengraph_match(self):
+        VALUE = 'http://www.piedpiper.com'
+        test = '<meta property="og:url" content="%s">' % VALUE
+        eq_(self.get_obj_with(test)._check_opengraph('url'), VALUE)
+
+    def test_check_link_nomatch(self):
+        eq_(self.get_empty_obj()._check_link('canonical'), None)
+
+    def test_check_link_match(self):
+        VALUE = 'http://www.piedpiper.com'
+        test = '<link rel="canonical" href="%s">' % VALUE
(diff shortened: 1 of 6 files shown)


mkt/websites/tests/test_views.py
```diff
@@ -244,3 +246,114 @@ def test_status_filtering(self):
...
+        eq_(response.status_code, 200)
+
+    def test_verbs(self):
+        self._allowed_verbs(self.url, ['get'])
+
+    def test_has_cors(self):
+        self.assertCORS(self.client.get(self.url), 'get')
+
+
+class TestWebsiteScrapeName(TestCase):
```
chuckharmston: Though pretty concise, this test case is very complex. Is the docstring sufficient? Is the code too opaque?

mkt/websites/views.py
```diff
@@ -61,3 +70,200 @@ class ReviewersWebsiteSearchView(WebsiteSearchView):
...
+                              RestOAuthAuthentication]
+    permission_classes = [AllowAny]
+
+    errors = {
+        'no_url': '`url` querystring parameter required',
+        'network': 'Unable to fetch website metadata',
+        'malformed_data': 'Unable to parse website HTML.'
+    }
+
+    def get_name(self, doc):
```
chuckharmston: These methods are really ugly, but they get the job done. Any suggestions for a more elegant way to get that data?

diox:
IMHO this should be a task fired asynchronously, not a view. Something working like `FileUpload` does: the API would return an `id` for a model instance that the task would then fill in, and clients can poll the API with that id till it returns something

mkt/websites/tests/test_views.py
```diff
@@ -244,3 +246,114 @@ def test_status_filtering(self):
...
+
+    def get_value(self, iteration):
+        return self.value % str(iteration)
+
+    def get_fragment(self, iteration):
+        return self.fragments[iteration - 1] % self.get_value(iteration)
+
+    def get_document(self, iteration):
+        doc = '<html>%s</html>'
+        inner = [self.get_fragment(i + 1) for i in xrange(0, iteration)]
```
mstriemer: The `+ 1` and `- 1` look like they should be in `get_value`.

mkt/websites/tests/test_views.py
```diff
@@ -244,3 +246,114 @@ def test_status_filtering(self):
...
+        return pq(doc % ''.join(inner))
+
+    def test_values(self):
+        for n in xrange(0, len(self.fragments)):
```
mstriemer: I find the adding and subtracting one pretty confusing. I understand what's supposed to be happening but it is quite distracting. It would be nice to pull the complexity into a generator or update the loop to start at 1 (or both).

```py
for document, result in self.get_test_data():
    eq_(result, self.view.get_name(document))

File: mkt/websites/tests/test_views.py (hunk 3 above)
Comment (mstriemer) [inline mkt/websites/tests/test_views.py]: Might be a good idea to shuffle `inner` before dropping it into the HTML so that you can't pass the tests based on the ordering of the `<meta>` elements.

File: mkt/websites/tests/test_views.py (hunk 1 above)
Comment (mstriemer) [inline mkt/websites/tests/test_views.py]: The docstring was pretty helpful. The code is split up well and with the docstring I knew what was going on right away.

File: mkt/websites/views.py (hunk 2 above)
Comment (mstriemer) [inline mkt/websites/views.py]: I'd suggest creating some helper methods like `get_title_text(doc)`, `get_meta_content(doc, selector)` and `get_link_href(doc, selector)` (these could take `doc(selector)` instead, too). If you can remove the `try`/`except` that would be better too, I'd say.

```py
def get_meta_content(el):
    el = doc(selector)
    if el and 'content' in el.attrib:
        return el.attrib['content']

def get_description(self, doc):
    return (get_meta_content('meta[name="description"]') or
            get_meta_content('meta[property="og:description"]') or
            get_meta_content('meta[name="msapplication-tooltip"]'))

Comment (mstriemer) [inline mkt/websites/views.py]: I'd rather see these methods pulled out of the view also since they don't have anything to do with a request and it seems like something we might want to use elsewhere.

File: mkt/websites/tests/test_views.py (hunk 1 above)
Comment (chuckharmston) [inline mkt/websites/tests/test_views.py]: :+1: Thanks!

**Comment by chuckharmston** [comment]
@diox You're not wrong, that's the right way to approach it. Submission tools are supposed to be an MVP, so I went with the simpler, "good enough for a few users" approach.

I could be convinced otherwise if you have strong feelings about it.

File: mkt/websites/views.py (hunk 2 above)
Comment (chuckharmston) [inline mkt/websites/views.py]: This is a fantastic suggestion. I refactored in the latest version (c2c8512), would love to hear your thoughts.

File: mkt/websites/tests/test_views.py (hunk 3 above)
Comment (chuckharmston) [inline mkt/websites/tests/test_views.py]: Refactored to only need the `+1` in one place, the second param of `xrange`.

Comment (chuckharmston) [inline mkt/websites/tests/test_views.py]: Also shuffled the array, good call.

File: mkt/websites/tests/test_helpers.py [hunk 5]
```diff
@@ -0,0 +1,150 @@
...
+       `value` string.
+    2) Creates a WebsiteMetadata object from that string.
+    3) Tests that the value of the `prop` key in the resulting dict matches
+       the unique value set in step 1.
+
+    This test is then repeated, instead creating a document with two elements,
+    the first two items in `fragments`, testing that `prop` is set to the
+    unique value of the second item. This is repeated by adding an additional
+    item from `fragments` until the list is exhausted.
+    """
```
chuckharmston: Longer now, since it's not specific to tests for `_get_name`. Still clear-ish?

chuckharmston:
Would appreciate another look, with a major refactor. @mstriemer?

chuckharmston:
Ready for a full review, pending Travis passing.

ngokevin:
r+ from me, let's try it out.

ngokevin:
where's this at

ngokevin:
bump

chuckharmston:
@ngokevin I'm working with Jason right now to get the DB schema on dev worked out for bug 1175691, then I'm going to try to get #3140 re-merged. If you want to merge this and run with it, feel free.

ngokevin:
That's not related so I'll merge to try it out since there's little else to do on the frontend.

PR #3149, opened 2015-06-22
Use the coloured globe icons as default for websites (bug 1174847)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1174847


mstriemer:
This picks a region and colour combination for each website. I'm now thinking we wanted to region to be determined by the user's region. That might be a bit weird to have different icons for the same website.

@pwalm how should this work? If we want to pick a certain globe region that will be more work. We can either pick from the full set or pin it to just americas, asia-australia or europe-africa.

pwalm:
My plan for the different colours was just to provide visual variance on things like list/grid views, so it isn't just a bunch of navy globes on a page. And I think I left it up to Krupa and Durst as to which version of the globe we display (instead of geo-targeting the appropriate icon (for now)), doesn't matter to me. Where do we have more users? That would be the ideal icon.

PR #3151, opened 2015-06-24
Remove uses of request.REQUEST (bug 1087034)

robhudson:


mkt/site/middleware.py
```diff
@@ -111,7 +111,7 @@ def process_request(self, request):
             if stored_ov_lang.lower() in settings.LANGUAGE_URL_MAP:
                 ov_lang = stored_ov_lang
 
-        if 'lang' in request.REQUEST:
+        if 'lang' in request.GET:
```
diox: Maybe we need `POST` here too. I'm thinking if you are doing a `POST` with a `lang` value it should set it like it does with a `GET`. Not sure if we have code depending on this behaviour though

robhudson: I can convert this easily enough but can we find where we might be doing this? Would it be fireplace?

diox: Maybe fireplace reviews use this, I don't remember if we care about the lang of the user reviews that are posted ?

robhudson: I looked in fireplace and couldn't fine any uses of passing lang as anything but a GET parameter in the route args.

PR #3152, opened 2015-06-24
Convert SortedDict to OrderedDict (bug 1087029)

robhudson:


requirements/prod.txt
```diff
@@ -76,8 +76,6 @@ ndg-httpsclient==0.3.3
 newrelic==2.20.1.18
 nose==1.3.4
 oauthlib==0.7.2
-# ordereddict is required by kombu
```
diox: Is this no longer true then ?

robhudson: No. I couldn't find the exact version but kombu now requires python 2.7.

robhudson: Here's one example:
https://github.com/celery/kombu/commit/d3a36c6750fafb6806fd027f429fceca556b0263

diox:
r+

PR #3156, opened 2015-06-26
autoescape app description (bug 1177390)

spasovski:
My local zamboni is wonked but this idea should work.


diox:
We should add a test, not sure of `PurifiedField` escaping works. But also, this would unbreak the page, but for apps with super long descriptions with lots of HTML it's going to look a bit weird... No idea how to fix it though :(

diox:
Thought about this more: since it's for opengraph description, just escaping is not enough, we should strip the HTML (don't want tags to appear at all). Probably truncate, too.

robhudson:
Any progress on this? It is a blocker for push tomorrow.

spasovski:
This approach is untested (and I don't have zamboni running locally). Would appreciate if someone else wants to take over.

PR #3159, opened 2015-06-29
Fix bug 1177718

robhudson:


mkt/search/filters.py
```diff
@@ -102,7 +105,9 @@ def filter_queryset(self, request, queryset, view):
 
         return queryset.query(
             'function_score',
-            query=query.Bool(should=should),
+            query=query.Bool(
+                should=[query.Bool(should=should, minimum_should_match=2),
```
diox: Though moving the guid/id makes sense, after spending some time to think about it more, I think the `preferred_regions` should not be done this way and we should not have the `minimum_should_match=2` here.

We only want the `preferred_region` to affect the scoring of documents already considered relevant. It seems we should be tweaking the `function_score` part instead, though I'm not sure how.

robhudson: I agree. This has too much impact on our search queries. It's way too late but using something related to `function_score` could be an interesting solution here, I agree.

robhudson:
This does more harm than good, closing for another solution.

PR #3162, opened 2015-06-30
ship the source

andymckay:
* fixes mozilla/marketplace-env#92

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,17 +19,7 @@ RUN yum install -y redis \
     supervisor && yum clean all
 
 COPY requirements /srv/zamboni/requirements
-
-# Remove some compiled deps so we just use the packaged versions already installed.
-#
-# If you install the M2Crypto version inside compiled.txt, you will hit an
-# error on libssl. Unfortunately this changes the layer, so the pip install
-# won't be cached by docker and makes builds way slower.
-#
-# Fixing this and removing this would speed things up a lot.
-RUN sed -i 's/M2Crypto.*$/# Removed in favour of packaged version/' /pip/requirements/compiled.txt
-
-RUN pip install --no-deps -r /srv/zamboni/requirements/prod.txt --find-links https://pyrepo.addons.mozilla.org/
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -18,15 +18,21 @@ RUN yum install -y redis \
...
+COPY requirements /srv/zamboni/requirements
 
 # Remove some compiled deps so we just use the packaged versions already installed.
+#
+# If you install the M2Crypto version inside compiled.txt, you will hit an
+# error on libssl. Unfortunately this changes the layer, so the pip install
+# won't be cached by docker and makes builds way slower.
+#
+# Fixing this and removing this would speed things up a lot.
 RUN sed -i 's/M2Crypto.*$/# Removed in favour of packaged version/' /pip/requirements/compiled.txt
```
robhudson: Does payments have this problem too?

Dockerfile
```diff
@@ -18,15 +18,21 @@ RUN yum install -y redis \
...
 
-# This cd into /pip ensures egg-links for git installed deps are created in /pip/src
-RUN cd /pip && pip install -b /pip/build --no-deps --download-cache /pip/cache -r /pip/requirements/dev.txt --find-links https://pyrepo.addons.mozilla.org/
+RUN pip install --no-deps -r /srv/zamboni/requirements/prod.txt --find-links https://pyrepo.addons.mozilla.org/
```
robhudson: Since we often shell into this should we install the requirements/test.txt instead? So we have the nose and flake8 libs.

Dockerfile
andymckay: No, because it doesn't automatically install `compiled.txt`

Dockerfile
andymckay: Probably a good idea.

Dockerfile
andymckay: fwiw, when @jasonthomas fixed the M2Crypto package he fixed this.

robhudson:
r+

PR #3163, opened 2015-07-02
Add devices to Websites model/api/lookup tool (bug 1179747)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1179747

### DIFF
--- a/migrations/919-add-websites-devices.sql
+++ b/migrations/919-add-websites-devices.sql
@@ -1,0 +1,5 @@
+-- Add devices column, which is a JSON array of supported device ids.
+ALTER TABLE `websites_website` ADD COLUMN `devices` longtext NOT NULL;
+-- Start with all existing websites compatible with
+-- all mobile (2, 4) / tablet (3), but not desktop (1).
+UPDATE `websites_website` SET devices='[2, 3, 4]';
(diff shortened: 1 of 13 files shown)


migrations/919-add-websites-devices.sql
```diff
@@ -0,0 +1,5 @@
+-- Add devices column, which is a JSON array of supported device ids.
+ALTER TABLE `websites_website` ADD COLUMN `devices` longtext NOT NULL;
+-- Start with all existing websites compatible with
+-- all mobile (2, 4) / tablet (3), but not desktop (1).
+UPDATE `websites_website` SET devices='[2, 3, 4]';
```
diox: This should ensure backwards-compatibility with existing websites. We'll need to reindex, though.

mstriemer:
Looks good to me. If someone visits on their iPhone I don't think they'll get any results after this, though. Should we start treating mobile but not Firefox OS as `android-mobile` and the same for tablet?

diox:
Mmmm, yeah. The way I see it, we have 4 choices:
 0. Do nothing
 1. Consider treat non Firefox browsers as `android-mobile`, `android-tablet`, `desktop` depending on what's the best match.
 2. New device types, used only for Websites, and find a way to make it work with multi-search
 3. The form factor stuff + new device types

Only your suggestion, 1., sounds really doable to me. We just need to make sure that we still prevent app installation / display incompatibility message on clients without `navigator.mozApps`. I could wait till this is done in fireplace before merging this.

mstriemer:
We get the incompatibility banner in Chrome for Android so I think treating `android-mobile` as `mobile` and `android-tablet` as `tablet` should be fine. Seems like we'll get an incompatibility warning on the detail page if there is no `navigator.mozApps`.

diox:
Ok yeah, I double-checked, even the buttons should stay disabled if `mozApps` is not present, so it should be good.

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1180296 is now fixed so I think we're ready to use this.

PR #3164, opened 2015-07-03
Fix mobile_url for Websites in the search API (bug 1179841)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1179841


mstriemer:
Looks good, tried it out locally also and it works for me. r+

PR #3167, opened 2015-07-08
update marketplace home screenshot (bug 1180654)

spasovski:
Updated to reflect Marketplace's new design.


mstriemer:
This looks like a new file. Shouldn't it replace https://github.com/mozilla/zamboni/blob/master/media/img/ecosystem/homescreen-partners.jpg?

diox:
Yeah, the path looks wrong. Also, can you run `jpegoptim` or similar tool to gain a couple kilobytes ?

ngokevin:
Don't mean to pile on, but could you set up Zamboni if you haven't already? We shouldn't be doing blind patches. Docker should make it simpler to set up.

spasovski:
Good catch. Needed to crop the image locally and saved it in the wrong place. In either case setting up zamboni now.

PR #3168, opened 2015-07-08
Adds WebsiteSubmission model, part 2.

chuckharmston:
This is an update of #3140, now that [bug 1175691](https://bugzilla.mozilla.org/show_bug.cgi?id=1175691) has been resolved. Migration has been updated accordingly.

r? @robhudson @ngokevin

### DIFF
--- a/migrations/920-adds-websitesubmission-model-back.sql
+++ b/migrations/920-adds-websitesubmission-model-back.sql
@@ -1,0 +1,39 @@
+CREATE TABLE `websites_websitesubmission_keywords` (
+    `id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY,
+    `website_id` integer UNIQUE,
+    `tag_id` integer unsigned NOT NULL,
+    UNIQUE (`website_id`, `tag_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+ALTER TABLE `websites_websitesubmission_keywords` ADD CONSTRAINT `websites_websitesubmission_keywords_tag_id` FOREIGN KEY (`tag_id`) REFERENCES `tags` (`id`);
+
+CREATE TABLE `websites_websitesubmission` (
+    `id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY,
+    `created` datetime NOT NULL,
+    `modified` datetime NOT NULL,
+    `date_approved` datetime NULL,
+    `name` integer unsigned UNIQUE,
+    `description` integer unsigned UNIQUE,
+    `categories` longtext NOT NULL,
+    `detected_icon` varchar(255) NOT NULL,
+    `icon_type` varchar(25) NOT NULL,
+    `icon_hash` varchar(8) NOT NULL,
+    `url` varchar(255) NOT NULL,
+    `canonical_url` varchar(255),
+    `works_well` integer unsigned UNIQUE,
+    `submitter_id` integer unsigned UNIQUE,
+    `public_credit` bool NOT NULL DEFAULT FALSE,
+    `why_relevant` longtext NOT NULL,
+    `preferred_regions` longtext NOT NULL,
+    `approved` bool NOT NULL DEFAULT FALSE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE INDEX websites_websitesubmission_created_idx ON websites_websitesubmission (created);
+CREATE INDEX websites_websitesubmission_modified_idx ON websites_websitesubmission (modified);
+CREATE INDEX websites_websitesubmission_approved_idx ON websites_websitesubmission (approved);
+
+ALTER TABLE `websites_websitesubmission` ADD CONSTRAINT `websites_websitesubmission_submitter_id` FOREIGN KEY (`submitter_id`) REFERENCES `users` (`id`);
+ALTER TABLE `websites_websitesubmission` ADD CONSTRAINT `websites_websitesubmission_name` FOREIGN KEY (`name`) REFERENCES `translations` (`id`);
+ALTER TABLE `websites_websitesubmission` ADD CONSTRAINT `websites_websitesubmission_description` FOREIGN KEY (`description`) REFERENCES `translations` (`id`);
+
+ALTER TABLE `websites_websitesubmission_keywords` ADD CONSTRAINT `websites_websitesubmission_websitesubmission_id` FOREIGN KEY (`websitesubmission_id`) REFERENCES `websites_websitesubmission` (`id`);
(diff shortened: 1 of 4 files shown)


chuckharmston:
Travis passes \o/

migrations/917-adds-websitesubmission-model.sql
```diff
@@ -0,0 +1,35 @@
...
+    `why_relevant` longtext NOT NULL,
+    `preferred_regions` longtext NOT NULL,
+    `approved` bool NOT NULL DEFAULT FALSE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+ALTER TABLE `websites_websitesubmission` ADD CONSTRAINT `websites_websitesubmission_submitter_id` FOREIGN KEY (`submitter_id`) REFERENCES `users` (`id`);
+ALTER TABLE `websites_websitesubmission` ADD CONSTRAINT `websites_websitesubmission_name` FOREIGN KEY (`name`) REFERENCES `translations` (`id`);
+ALTER TABLE `websites_websitesubmission` ADD CONSTRAINT `websites_websitesubmission_description` FOREIGN KEY (`description`) REFERENCES `translations` (`id`);
+
+ALTER TABLE `websites_websitesubmission_keywords` ADD CONSTRAINT `websites_websitesubmission_websitesubmission_id` FOREIGN KEY (`websitesubmission_id`) REFERENCES `websites_websitesubmission` (`id`);
```
robhudson: 917 is taken

migrations/917-adds-websitesubmission-model.sql
```diff
@@ -0,0 +1,35 @@
+CREATE TABLE `websites_websitesubmission_keywords` (
+    `id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY,
+    `website_id` integer UNIQUE,
+    `tag_id` integer unsigned NOT NULL,
+    UNIQUE (`website_id`, `tag_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+ALTER TABLE `websites_websitesubmission_keywords` ADD CONSTRAINT `websites_websitesubmission_keywords_tag_id` FOREIGN KEY (`tag_id`) REFERENCES `tags` (`id`);
```
robhudson: No constraint to `websites_website.id`?

robhudson: Oh, it's down there on L35.

migrations/917-adds-websitesubmission-model.sql
robhudson: I'm confused. 917 is here but the same stuff is in 920?

chuckharmston: Merge botch. Will fix, good catch.

mkt/websites/models.py
```diff
@@ -28,6 +30,79 @@
...
+    # preferred_regions should be set to [].
+    preferred_regions = JSONField(default=None)
+
+    # Turn true when a reviewer has approved and published a submission.
+    approved = models.BooleanField(default=False)
+
+    objects = WebsiteSubmissionManager()
+
+    class Meta:
+        ordering = (('-modified'), )
```
robhudson: Might be a good idea to put an index on `modified`.

migrations/920-adds-websitesubmission-model-back.sql
```diff
@@ -0,0 +1,38 @@
...
+    `works_well` integer unsigned UNIQUE,
+    `submitter_id` integer unsigned UNIQUE,
+    `public_credit` bool NOT NULL DEFAULT FALSE,
+    `why_relevant` longtext NOT NULL,
+    `preferred_regions` longtext NOT NULL,
+    `approved` bool NOT NULL DEFAULT FALSE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE INDEX websites_websitesubmission_created_idx ON websites_websitesubmission (created);
+CREATE INDEX websites_websitesubmission_modified_idx ON websites_websitesubmission (modified);
```
robhudson: `approved` also, since it will almost always be filtered by the approved flag one way or the other.

migrations/920-adds-websitesubmission-model-back.sql
```diff
@@ -0,0 +1,38 @@
...
+    `canonical_url` varchar(255),
+    `works_well` integer unsigned UNIQUE,
+    `submitter_id` integer unsigned UNIQUE,
+    `public_credit` bool NOT NULL DEFAULT FALSE,
+    `why_relevant` longtext NOT NULL,
+    `preferred_regions` longtext NOT NULL,
+    `approved` bool NOT NULL DEFAULT FALSE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE INDEX websites_websitesubmission_created_idx ON websites_websitesubmission (created);
```
robhudson: I was thinking only `modified` since it's part of the `ordering` in the model. If we will query it with a `created` order often this is fine.

mkt/websites/models.py
```diff
@@ -28,6 +30,79 @@
...
+
+    # Why does the user think the website is relevant to Marketplace?
+    why_relevant = models.TextField()
+
+    # If the submitter says the website is relevant worldwide,
+    # preferred_regions should be set to [].
+    preferred_regions = JSONField(default=None)
+
+    # Turn true when a reviewer has approved and published a submission.
+    approved = models.BooleanField(default=False)
```
robhudson: If you add an index on this field in the migration also add it here, `db_index=True`.

robhudson:
r+wc

migrations/920-adds-websitesubmission-model-back.sql
chuckharmston: I did this consciously because I think both will happen regularly.

mkt/websites/models.py
```diff
@@ -28,6 +30,79 @@
...
+
+    objects = WebsiteSubmissionManager()
+
+    class Meta:
+        ordering = (('-modified'), )
+
+    def __unicode__(self):
+        return unicode(self.url)
+
+
```
robhudson: You could help keep the migration and python code in sync by adding:
```
WebsiteSubmission._meta.get_field('created').db_index = True
WebsiteSubmission._meta.get_field('modified').db_index = True
```

chuckharmston: Kind of shocked that this isn't here: https://github.com/mozilla/zamboni/blob/be1aa96d5e24fe398c18929732f4cac4f75769a8/mkt/site/models.py#L294-L295

robhudson: And you can't simply override the model by adding your own `created` field with it. Seems like Django should allow that when the base class is abstract. But oh well.

I think it's probably better to not have an index in the base class b/c it could add fs size when not needed.

chuckharmston: Maybe. It is pretty low cost to do so if it's just an index on a few date fields though.

PR #3169, opened 2015-07-08
Fix search when explicitly passed region=None

robhudson:


robhudson:
This fixes the issue found here:
http://sentry.mktmon.services.phx1.mozilla.com/mkt/marketplacefirefoxcom/group/16335/

chuckharmston:
Is there a bug number for this? Commit message doesn't really describe what was broken.

robhudson:
No, just a fix for a sentry error I saw. I could file one and attach it tho which is probably a better idea.

mkt/search/tests/test_views.py
```diff
@@ -1389,6 +1389,12 @@ def test_search_preferred_region_match(self):
         objs = res.json['objects']
         eq_(len(objs), 0)
 
+    def test_region_explicit_none(self):
+        res = self.anon.get(self.url, data={'region': 'None'})
```
chuckharmston: It's really weird that the querystring parameter `None` is turned into the Python `None` value. Where does that happen?

robhudson: https://github.com/mozilla/zamboni/blob/be1aa96d5e24fe398c18929732f4cac4f75769a8/mkt/api/base.py#L51-L52

chuckharmston: I guess that works. I'd have used 0, but ¯\_(ツ)_/¯

chuckharmston:
r+ if tests pass, looks good.

PR #3171, opened 2015-07-09
De-activate cache-machine in production (bug 881405)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=881405

r? @jasonthomas


jasonthomas:
r+ thanks!

PR #3172, opened 2015-07-09
Set devices to everything but desktop when importing websites (bug 1182039)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1182039

### DIFF
--- a/mkt/websites/utils.py
+++ b/mkt/websites/utils.py
@@ -1,5 +1,6 @@
 import random
 
+from mkt.constants.applications import DEVICE_TYPES
 from mkt.constants.base import STATUS_PUBLIC
 from mkt.websites.models import Website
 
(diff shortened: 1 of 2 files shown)


mkt/websites/utils.py
```diff
@@ -27,6 +28,7 @@ def website_factory(**kwargs):
         'title': 'Title for %s' % text.capitalize(),
         'url': 'http://%s.example.com' % text,
         'mobile_url': 'http://mobile.%s.example.com' % text,
+        'devices': DEVICE_TYPES.keys(),
```
mstriemer: Fake data websites get all device types?

mstriemer:
Changes look good to me. I feel like fake data websites shouldn't have desktop set or some subset of the fake data websites shouldn't have desktop set. That probably doesn't really matter though.

r+

mkt/websites/utils.py
diox: Yeah, could not decide, and it's just a default - the caller can override any of those parameters, so it doesn't matter much really.

PR #3174, opened 2015-07-10
Remove some obsolete code that manipulates files (bug 1167271)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1167271

Trying to simplify all code that deals with files before switching everything to using storage.


eviljeff:
(I should have removed this during the apps->mkt move really)
r+

PR #3177, opened 2015-07-10
remove receipt verifier

andymckay:
* gunicorn is needed for this but has been removed


ddurst:
r+

kumar303:
fwiw, it would be really nice to run the receipt verifier in docker. I had to do it a couple times for fxpay development.

PR #3182, opened 2015-07-14
Adjust Django models to match schema (bug 1175691)

robhudson:


washort:
r+ nice finds. Let's go ahead and add these indices now, and make a note to consider which ones help us most (or if we need new ones) soon.

PR #3183, opened 2015-07-15
Convert FileViewer to use storage API (bug 1167271)

washort:


mkt/files/tests/test_helpers.py
```diff
@@ -73,7 +74,7 @@ def test_get_files_not_extracted(self):
     def test_get_files_size(self):
         self.viewer.extract()
         files = self.viewer.get_files()
-        eq_(len(files), 14)
+        eq_(len(files), 15)
```
robhudson: how'd we get an extra one?

mkt/files/tests/test_helpers.py
```diff
@@ -118,10 +119,12 @@ def test_syntax(self):
...
         dest = self.viewer.dest
-        open(os.path.join(dest, 'chrome.manifest'), 'w')
+        storage.open(os.path.join(dest, 'manifest.webapp'), 'w').write(".")
         subdir = os.path.join(dest, 'chrome')
-        os.mkdir(subdir)
-        open(os.path.join(subdir, 'foo'), 'w')
+        storage.open(os.path.join(subdir, 'foo'), 'w').write(".")
+        if not storage.exists(subdir):
+            # might be on S3, which doesn't have directories
+            storage.open(subdir, 'w').write(".")
```
robhudson: What's the `write(".")` stuff about? (Also, move to single quotes?)

mkt/files/tests/test_helpers.py
washort: Added a `manifest.webapp` to the zip file so that `is_extracted` can detect it.

mkt/files/tests/test_helpers.py
washort: didn't remember if s3 supported empty files. looks like it  actually does so i'll fix this

robhudson:
r+

PR #3184, opened 2015-07-16
Add support for generating fake inapp payments apps (payments/#92)

washort:
See https://github.com/mozilla/payments/issues/92

### DIFF
--- a/mkt/developers/models.py
+++ b/mkt/developers/models.py
@@ -170,11 +170,13 @@ class UserInappKey(ModelBase):
         self._product().patch(data={'secret': generate_key(48)})
 
     @classmethod
-    def create(cls, user):
+    def create(cls, user, secret=None):
+        if secret is None:
+            secret = generate_key(48)
         sel = SolitudeSeller.create(user)
         # Create a product key that can only be used for simulated purchases.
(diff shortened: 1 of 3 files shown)


mkt/developers/models.py
```diff
@@ -170,11 +170,13 @@ def reset(self):
...
-    def create(cls, user):
+    def create(cls, user, secret=None):
+        if secret is None:
+            secret = generate_key(48)
         sel = SolitudeSeller.create(user)
         # Create a product key that can only be used for simulated purchases.
         prod = client.api.generic.product.post(data={
-            'seller': sel.resource_uri, 'secret': generate_key(48),
+            'seller': sel.resource_uri, 'secret': secret,
             'external_id': str(uuid.uuid4()), 'public_id': str(uuid.uuid4()),
```
kumar303: The `public_id` here also needs to be set by the fake data script like:
````
UserInappKey.create(acct.user, secret='fake data secret key', public_id='fake data public ID')
````
That way the functional tests can configure their in-app tester [key and secret](https://github.com/kumar303/inapp-pay-test/blob/master/config-dist.js#L30-L31) with known values.

PR #3185, opened 2015-07-16
Add logging for setting priority review status via edit listing

eviljeff:


mkt/developers/forms.py
```diff
@@ -361,7 +362,10 @@ def save(self, addon, commit=True):
         contact = self.cleaned_data.get('mozilla_contact')
         if contact is not None:
             updates['mozilla_contact'] = contact
+        priority_before = addon.priority_review
         addon.update(**updates)
+        if self.cleaned_data.get('priority_review') and not priority_before:
```
mstriemer: Can you inline `addon.priority_review` here? Looks like `priority_before` is only used once.

mkt/developers/utils.py
```diff
@@ -154,3 +154,14 @@ def escalate_prerelease_permissions(app, validation, version):
         escalate_app(
             app, version, nobody, 'App uses prerelease permissions',
             mkt.LOG.ESCALATION_PRERELEASE_APP)
+
+
+def prioritize_app(app, user):
+    if not app.priority_review:
```
mstriemer: Do you still want to send the comms and log it if the app is already flagged for priority review? Seems like it would make more sense to return here with `if app.priority_review: return`. It would simplify the form also because you could skip that extra check and call this whenever requested.

mstriemer: Looks like this came from `mkt.lookup.views` and it would only do this if the app wasn't already priority review.

mkt/developers/forms.py
eviljeff: This is because .priority_review is being updated in the `addon.update(**updates)` line above.

mkt/developers/utils.py
eviljeff: I was wanting to avoid multiple updates to the record in the database - `priority_review` is set with all the updates in that edit listing section in forms.py.  But the message and so on needs to go out if it was changed still.  Aware the code is a little weird though - I'm open to suggestions.

mstriemer: I'm tempted to say take the update out of here and rename this `notify_app_prioritization` or something. I do like having a function you can call and have it do everything you need to prioritize an app though. You could also remove `priority_review` from the `updates` dict above and just live with the fact that we're going to do two updates. I'm not sure if there are any real downsides to doing two updates.

mstriemer: I guess you could split this function in two as well. Have a `notify_app_prioritization` function and then this can be:

```py
def prioritize_app(app, user):
    if not app.priority_review:
        app.update(priority_review=True)
        notify_app_prioritization(app, user)
```

But then we're writing a function that only gets called once, so that might not be better.

eviljeff:
I made prioritize_app always update priority_review and moved around the updates in forms.py so I didn't need the intermediate variable and wouldn't bother trying to update priority_review twice if it was being done in prioritize_app.

mstriemer:
Looks good. Test failure appears unrelated. r+

PR #3186, opened 2015-07-17
Allow fake data to create real purchase keys

kumar303:


kumar303:
It's working, just waiting for a review. I'll bug some people again.

kumar303:
from IRC

> ashort: kumar: there aren't any unit tests that would cover this

That was my only question. I'm just going to merge this so I don't hold you up anymore ;)

PR #3187, opened 2015-07-20
in-app products for web packages without origins

kumar303:
Fixes: https://github.com/mozilla/fxpay/issues/116

### DIFF
--- a/mkt/developers/views_payments.py
+++ b/mkt/developers/views_payments.py
@@ -391,15 +391,27 @@ def in_app_products(request, addon_id, addon, webapp=True, account=None):
     products = addon.inappproduct_set.all()
     new_product = InAppProduct(webapp=addon)
     form = InAppProductForm()
-    list_url = None
-    detail_url = None
+
     if addon.origin:
-        list_url = _fix_origin_link(reverse('in-app-products-list',
-                                            kwargs={'origin': addon.origin}))
-        detail_url = _fix_origin_link(reverse('in-app-products-detail',
-                                              # {guid} is replaced in JS.
-                                              kwargs={'origin': addon.origin,
-                                                      'guid': "{guid}"}))
+        inapp_origin = addon.origin
+    elif addon.guid:
+        # Derive a marketplace specific origin out of the GUID.
+        # This is for apps that do not specify a custom origin.
+        inapp_origin = 'marketplace:{}'.format(addon.guid)
+    else:
+        # Theoretically this is highly unlikely. A hosted app will
+        # always have a domain and a packaged app will always have
+        # a generated GUID.
+        raise TypeError(
+            'Cannot derive origin: no declared origin, no GUID')
+
+    list_url = _fix_origin_link(reverse('in-app-products-list',
+                                        kwargs={'origin': inapp_origin}))
+    detail_url = _fix_origin_link(reverse('in-app-products-detail',
+                                          # {guid} is replaced in JS.
+                                          kwargs={'origin': inapp_origin,
+                                                  'guid': "{guid}"}))
+
     return render(request, 'developers/payments/in-app-products.html',
                   {'addon': addon, 'form': form, 'new_product': new_product,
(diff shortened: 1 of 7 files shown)


kumar303:
Client changes: https://github.com/mozilla/fxpay/pull/117

mkt/developers/views_payments.py
```diff
@@ -391,15 +391,23 @@ def in_app_products(request, addon_id, addon, webapp=True, account=None):
...
-                                                      'guid': "{guid}"}))
+
+    inapp_origin = addon.origin
+    if not inapp_origin:
+        if not addon.guid:
+            raise TypeError(
+                'Cannot derive origin: no declared origin, no GUID')
+        # Derive a marketplace specific origin out of the GUID.
+        # This is for apps that do not specify a custom origin.
+        inapp_origin = 'marketplace:{}'.format(addon.guid)
```
mstriemer: I'd try to avoid some of this nesting by stating the 3 cases:

```py
if addon.origin
    inapp_origin = addon.origin
elif addon.guid:
    # Derive a marketplace specific origin out of the GUID.
    # This is for apps that do not specify a custom origin.
    inapp_origin = 'marketplace:{}'.format(addon.guid)
else:
    raise TypeError('Cannot derive origin: no declared origin, no GUID')

File: mkt/inapp/tests/test_views.py [hunk 2]
```diff
@@ -201,6 +204,37 @@ def test_filter(self):
...
+class TestInAppProductsWithPackagedWebApp(AuthenticatedInAppProductTest):
+
+    def setUp(self):
+        super(TestInAppProductsWithPackagedWebApp, self).setUp()
+        # Set up a packaged web app, i.e. one without a declared domain.
+        self.webapp.update(app_domain=None, is_packaged=True)
+        self.product = self.create_product()
+
+    def marketplace_origin(self):
+        return 'marketplace:{}'.format(self.webapp.guid)
```
mstriemer: Is it worth throwing an `assert self.webapp.guid, 'webapp needs a guid'` before this line as a sanity check?

mkt/inapp/views.py
```diff
@@ -52,9 +52,13 @@ def get_queryset(self):
         return InAppProduct.objects.filter(webapp=self.get_app())
 
     def get_app(self):
+        origin = self.kwargs['origin']
         if not hasattr(self, 'app'):
-            self.app = get_object_or_404(Webapp,
-                                         app_domain=self.kwargs['origin'])
+            if origin.startswith('marketplace:'):
+                lookup = dict(guid=origin.replace('marketplace:', ''))
```
mstriemer: I feel like these should use `{'guid': origin.replace('marketplace:', '')}` instead of calling `dict` but I don't know if that's explicitly in our style guide.

kumar303: I just went with dict because I could copy/paste the previous keyword arg definition

mkt/inapp/tests/test_views.py
kumar303: this is just a test case helper but I realize I forgot to generate one in the setup, whoops

mkt/inapp/views.py
mstriemer: Also I'm probably being a crazy person but it might be a good idea to set the count: `origin.replace('marketplace:', '', 1)`.

mkt/developers/views_payments.py
kumar303: ah, that reads cleaner, thanks

mkt/inapp/tests/test_views.py
mstriemer: I know it's a helpers but we don't want all these tests to run with a webapp that has no guid since that might mess things up. Settings one explicitly in `setUp` might work too.

mstriemer:
r+wc

kumar303:
Thanks for the review!

PR #3188, opened 2015-07-21
Django migrations

washort:
Initial Django migrations and support for migrating translation fields.


mkt/site/fields.py
```diff
@@ -21,10 +21,15 @@ class SeparatedValuesField(fields.Field):
     """
 
     def __init__(self, base_field, separator=None, *args, **kwargs):
+        self.args, self.kwargs = self.args, self.kwargs
```
robhudson: RHS shouldn't have the `self`.

robhudson:
Add the migrations to the exclude line in `.travis.yml` so flake8 will pass. Nice work so far. What's left?

mkt/webapps/migrations/0001_initial.py
```diff
@@ -0,0 +1,578 @@
...
+                ('popular_region', models.CharField(max_length=10, null=True)),
+                ('banner_regions', mkt.webapps.models.RegionListField(null=True)),
+                ('region_de_usk_exclude', models.BooleanField(default=False)),
+                ('region_cn_status', models.PositiveIntegerField(default=2, help_text=b'<django.utils.functional.__proxy__ object at 0x2b32d90> approval status', db_index=True, choices=[(0, 'Incomplete'), (16, 'Unlisted'), (2, 'Pending approval'), (4, 'Published'), (5, 'Banned from Marketplace'), (11, 'Deleted'), (12, 'Rejected'), (13, 'Approved but private'), (15, 'Blocked')])),
+                ('region_cn_nominated', models.DateTimeField(help_text=b'<django.utils.functional.__proxy__ object at 0x2b32d90> nomination date', null=True)),
+                ('region_br_iarc_exclude', models.BooleanField(default=False)),
+                ('region_de_iarc_exclude', models.BooleanField(default=False)),
+                (b'TranslatedField', mkt.translations.fields.TranslatedField(null=True, on_delete=django.db.models.deletion.SET_NULL, to_field=b'id', to=mkt.translations.models.Translation, blank=True, unique=True)),
```
robhudson: curious why this is a byte string

washort: Huh. Maybe because I didn't return a unicode string from `reconstruct()` on that field. Probably some python3-compat nonsense.

.travis.yml
```diff
@@ -13,7 +13,7 @@ install:
   - npm install
 before_script:
   - mysql -e 'create database zamboni;'
-  - flake8 . --exclude='node_modules,commonplace_projects,docs,.git'
+  - flake8 . --exclude='node_modules,commonplace_projects,docs,.git,lib/es/migrations,mkt/abuse/migrations,mkt/access/migrations,mkt/api/migrations,mkt/comm/migrations,mkt/commonplace/migrations,mkt/developers/migrations,mkt/feed/migrations,mkt/files/migrations,mkt/inapp/migrations,mkt/langpacks/migrations,mkt/monolith/migrations,mkt/operators/migrations,mkt/prices/migrations,mkt/purchase/migrations,mkt/ratings/migrations,mkt/reviewers/migrations,mkt/site/migrations,mkt/submit/migrations,mkt/tags/migrations,mkt/users/migrations,mkt/versions/migrations,mkt/webapps/migrations,mkt/webpay/migrations,mkt/websites/migrations,mkt/zadmin/migrations'
```
robhudson: I'm pretty sure this is a pattern, so you don't need to list each in full. It's worth trying just `migrations`.

mkt/translations/fields.py
```diff
@@ -31,8 +32,19 @@ def __init__(self, **kwargs):
         kwargs.update(options)
         self.short = kwargs.pop('short', True)
         self.require_locale = kwargs.pop('require_locale', True)
+
+        # "to" is passed here from the migration framework; we ignore it
+        # since it's the same for every instance.
+        kwargs.pop("to", None)
```
robhudson: single quotes

.travis.yml
diox: Yeah, we should not need to hardcode all those dirs, just `migrations` should work (and it will not exclude files named `migrations.py` so we should be fine)

mkt/abuse/migrations/0003_abusereport_website.py
```diff
@@ -0,0 +1,21 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='abusereport',
+            name='website',
+            field=models.ForeignKey(related_name='abuse_reports', to='websites.Website', null=True),
+            preserve_default=True,
+        ),
+    ]
```
robhudson: Can we squash these 3 into one? Might as well start fresh.

mkt/comm/migrations/0002_auto_20150722_1517.py
```diff
@@ -0,0 +1,52 @@
...
+            field=models.ForeignKey(related_name='notes', to='comm.CommunicationThread'),
+            preserve_default=True,
+        ),
+        migrations.AddField(
+            model_name='commattachment',
+            name='note',
+            field=models.ForeignKey(related_name='attachments', to='comm.CommunicationNote'),
+            preserve_default=True,
+        ),
+    ]
```
robhudson: Same here? Squash?

mkt/site/tests/__init__.py
```diff
@@ -430,6 +430,12 @@ def shortDescription(self):
     def _pre_setup(self):
         super(TestCase, self)._pre_setup()
 
+        # XXX See if we can remove this when we switch to Django 1.8.
+        # Some crud from the migration system gets stuck here and we have to
+        # kick it loose manually.
+        from mkt.webapps.models import AddonUser
+        UserProfile.addonuser_set.related.model = AddonUser
```
robhudson: Did you ever find out the underlying cause? I'd feel better knowing what the actual problem is and possibly fixing it or linking to a Django bug.

washort: When loading models, the migrations system creates fake versions of models that have all the same database schema as the regular ones; this is sufficient for its purposes but doesn't match the `isinstance` check done for foreign-key assignment. Some of this code was rewritten for 1.8 so I have hopes that it'll work better there (though our dependencies are far from ready to support 1.8 it seems).

mkt/abuse/migrations/0003_abusereport_website.py
washort: I'll give it a try. It generated all 3 at once, so I don't know if there's a load-order problem it's trying to avoid.

robhudson: Yeah, I was wondering about that when I see the `dependencies`. If not no big deal.

mkt/site/tests/__init__.py
diox: But why does this bleed in our tests, and why is it only a problem for this particular FK ? This should not happen, so I feel we should figure out whether there is an actual bug on django's side or ours before proceeding with this hack

washort: Bleeds into our tests because the migrations are run in the same process. If I run the tests with `REUSE_DB=1` they pass just fine. Why this FK? Not sure. I thought it was due to `UserForeignKey` but I replaced it with a regular `ForeignKey` and the problem persisted.

washort:
The only way to remove that `AddonUser` hack for the tests is to replace all uses of `self.addonuser_set` with direct references to `AddonUser`; since this is only an issue in the tests I didn't feel like it was warranted. I tried squashing the migrations down to a single one per app but got circular dependency problems. So I think this is as good as it gets for today.

robhudson:
What's left here? Anything I can help do or test before we merge this?

washort:
I think this is ready to go, but we should probably wait til we get new schema on prod before merging and dropping schematic.

washort:
merged https://github.com/mozilla/zamboni/commit/5a6366ed591bfd58ed3f19347eb58c518459e725

PR #3193, opened 2015-07-22
[WIP] Updated filesystem calls to use django-storages (bug 1167271)

robhudson:


mkt/developers/tasks.py
```diff
@@ -61,8 +62,18 @@ def validator(upload_id, **kw):
         log.info(u'[FileUpload:%s] Does not exist.' % upload_id)
         return
 
+    temp_path = None
+    # Make a copy of the file if it's a packaged app since we can't assume the
+    # uploaded file is on the local filesystem.
+    if upload.name.endswith('.zip'):
```
diox: Do we have something more reliable than that ? Or are we already doing that everywhere anyway ? It seems fragile. I'd rather do something like `if not upload.name.startswith(('https://', 'http://'))`

mkt/site/utils.py
```diff
@@ -265,11 +265,11 @@ def __iter__(self):
             return iter([])
 
         chunk = 4096
-        fp = open(self.path, 'rb')
+        fp = storage.open(self.path, 'rb')
```
diox: The entire `HttpResponseSendFile` class won't work :( There are legitimate uses for it though - things that do status/permission checks before serving a package for instance - so I wonder what we'll replace it with. Not sure a redirect will be enough (I worry about security and compatibility with older FxOS devices)

PR #3196, opened 2015-07-23
Don't load all Price currencies everytime we load a Webapp (bug 1089848)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1089848

This already happens once per worker thread, since `Price.get_price_currency()` calls `Price.transformer([])` if necessary. It could be useful to get rid of that too, but let's start slow first.

### DIFF
--- a/mkt/webapps/serializers.py
+++ b/mkt/webapps/serializers.py
@@ -394,18 +394,6 @@ class ESAppSerializer(BaseESSerializer, AppSerializer):
         # Remove fields that we don't have in ES at the moment.
         self.fields.pop('upsold', None)
 
-    def _attach_payments_info(self, obj):
-        try:
-            # When fetching Webapp from the database, optimizations are made
-            # with attach_prices() etc. But we can't do that with ES, so we
-            # have to manually select_related() the price when fetching the
-            # AddonPremium object to do fewer queries. The `premium` property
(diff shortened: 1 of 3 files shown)


mkt/webapps/serializers.py
```diff
@@ -468,9 +456,10 @@ def fake_object(self, data):
         # regions stored in ES instead of making SQL queries.
         obj.get_excluded_region_ids = lambda: data['region_exclusions']
 
-        # Set up payments stuff to avoid extra queries.
+        # Set up payments stuff to avoid extra queries later (we'll still make
+        # some, because price info is not in ES).
         if obj.is_premium():
-            self._attach_payments_info(obj)
+            Webapp.attach_premiums([obj])
```
diox: Doing this higher up, in `field_to_native()`/data`, would spare us a few queries when there are multiple premium apps returned, I might do that later once all tests pass.

washort:
r+

robhudson:
Is this mostly just removing that big select all on `Price`? B/c the rest seems to be the same just moved to the transformer?

diox:
Yes. This is unnecessary, when you get a price for the first time, it actually does this behind the scenes, except it caches it as a property on the `Price` class (see description/full commit message).

PR #3197, opened 2015-07-24
Implement new Feature Profile signature mechanism in the API (bug 1172487)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1172487

Instead of encoding all features into a single huge bitfield, use smaller ones that are then encoded in base64. This allows to have more than 53 features in JavaScript, where integers are not safe to use past 2^53.

Code for mozilla/fireplace to follow later.

### DIFF
--- a/mkt/constants/features.py
+++ b/mkt/constants/features.py
@@ -1,4 +1,6 @@
+import base64
 import itertools
+import math
 from collections import OrderedDict
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


mkt/constants/features.py
```diff
@@ -361,6 +363,56 @@
...
+        store the values.
+        """
+        self.size = size
+        if values is not None:
+            self.values = values
+        else:
+            self.values = [0] * int(math.ceil(self.size / 8.0))
+
+    def get(self, i):
+        index = (i / 8) | 0
```
washort: `| 0` seems useless here

diox: Right, copied over from my JavaScript version, but we are in python :)

washort:
Looks great, r+

PR #3198, opened 2015-07-27
Do fewer queries in mkt.downloads.views.download_file (bug 1186901)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1186901

There were a few issues previously:
- We were loading all transforms on `File`, `Version` and `Webapp`, which is costly and useless here
- We were calling `sign_if_packaged()` to find out the file path, but this should already have been done before by the reviewer utils before changing the `Webapp` status, so we can avoid it and the additional queries it makes.


robhudson:
r+ nice

PR #3199, opened 2015-07-27
Add an action to reviewer tools to flag for rereview bug 1040929

eviljeff:


mkt/reviewers/utils.py
```diff
@@ -419,6 +429,14 @@ def get_actions(self):
                              u'The message won\'t be visible to the '
                              u'author(s), and no notification will be sent '
                              u'them.')}
+        manual_rereview = {
+            'method': self.handler.process_manual_rereview,
+            'label': _lazy(u'Request Re-review'),
+            'minimal': True,
+            'details': _lazy(u'Add this app to the re-review queue. Any '
+                             u'comments here won''t be visible to the '
```
diox: you want double quotes here to make the `'` work (or escaping, but we prefer double quotes)

eviljeff: above uses \' - I guess that works too?

diox: Yeah it works too, it's just more annoying in translations IMHO, that's why we usually use double quotes when there are `'` in the string

eviljeff: I used escaping for consistency with the rest of the strings.

diox:
r+

PR #3200, opened 2015-07-27
update homescreen image on devhub (bug 1180654)

spasovski:
Didn't compress much but I used "save for web" so not very surprising:

homescreen-partners.jpg 320x480 24bit N Exif XMP Adobe  [OK] 26317 --> 26301 bytes (0.06%), optimized.


diox:
I don't think it matters much here but you can run that through `jpegoptim` or similar tool to gain a few more bytes:
`homescreen-partners.jpg 320x480 24bit N Exif JFIF  [OK] 26301 --> 25418 bytes (3.36%), optimized.`

r+wc

spasovski:
That's exactly what I ran and got the 0.06% optimization. Did you use any switches?

diox:
Ah, misread the description. I didn't use any switches, might be the version ? I'm using jpegoptim 1.3.0. I don't think it matters that much for this particular image though.

spasovski:
1.4.2 here. Thanks for the quick r.

PR #3201, opened 2015-07-28
[WIP] Make dump_apps work using a storage class (bug 1167271)

diox:


mkt/site/utils.py
```diff
@@ -382,10 +382,16 @@ class LocalFileStorage(FileSystemStorage):
     Unlike Django's default FileSystemStorage, this class behaves more like a
     "cloud" storage system. Specifically, you never have to write defensive
     code that prepares for leading directory paths to exist.
+
+    It will also overwrite existing files by default - it's the caller
+    responsability to check for file existence first.
```
robhudson: "responsibility"

robhudson:
Lots of "FIXME"s here but I like the idea of using subclasses.

diox:
This is trickier than I thought:
- The tarball is built in `compress_export()`, which calls the `tar` command. So all files must exist on the local filesystem (or at least something POSIX-compatible like NFS)
- There are different commands: `process_addons --task=dump_apps` and `export_data` relying on the same code. The former uses task chaining in `process_addons.py`, the latter does it in `tasks.py` - but in any case, that means that part of the code needs to access files in a way that is shared across workers... 

Those 2 are difficult to reconcile. I initially started modifying the code to always use the local filesystem, but that means I should nuke the task chaining or make sure it all happens on the exact same worker somehow.

diox:
Closing for now because of previous comment. We might try to get rid of dump_apps, or we'll look at this later.

PR #3202, opened 2015-07-29
Push apps requiring external-app permission to escalated queue (bug 1178810)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1178810


mkt/comm/templates/comm/emails/escalation_reserved_perms_app.html
```diff
@@ -3,8 +3,9 @@
 {% block content %}
 Dear Superheroes,
 
-A developer has uploaded/resubmitted an app that uses prerelease permissions
-(the ones that used to be certified but are now privileged).
+A developer has uploaded/resubmitted an app that uses reserved permissions
+(the ones that used to be certified but are now privileged, or that are only
+ intended for specific partners).
```
eviljeff: I wonder can we rephrase to avoid saying 'partners' - to amateurs it may sound like we're excluding them; and to commercial developers that they can pay us to use these permissions.  "specific projects" ?

eviljeff:
Are there validator -side changes needed?  (I forget)

mkt/comm/templates/comm/emails/escalation_reserved_perms_app.html
diox: Yeah, didn't like "partners" very much either. "specific projects" works for me.

diox:
The `external-app` has been added to app-validator already. It's missing the warning we add in the validator itself for pre-release perms, but I don't think it's worth adding the warning - that permission is not really documented anyway.

eviljeff:
r+wc

PR #3204, opened 2015-07-29
Cleaner local-file handling for validator

washort:


mkt/developers/tasks.py
```diff
@@ -63,17 +64,15 @@ def validator(upload_id, **kw):
         return
 
     temp_path = None
-    # Make a copy of the file if it's a packaged app since we can't assume the
+    # Make a copy of the file since we can't assume the
     # uploaded file is on the local filesystem.
-    if upload.name.endswith('.zip'):
```
diox: I didn't like the `endswith('.zip')` check, but I'm not entirely sure it's a good idea to even copy hosted apps manifests. The validator does support hosted apps URLs just fine, I'm not sure how it works when you feed it a local hosted app manifest

diox: Nevermind, it does require the contents of the manifest. How did tests pass with that change then ?

washort: Good question. But we have to copy it locally for the `is_zipfile` check at the very least.

diox:
r+

PR #3206, opened 2015-07-30
Add s3dev site settings

jasonthomas:


diox:
Please set `AWS_QUERYSTRING_AUTH` to `False` for now, it's going to help us while we figure out if we can make expiring links work

PR #3208, opened 2015-07-31
Test fixes for S3 storage (bug 1167271)

washort:
Some more test fixes.


mkt/reviewers/tests/test_views.py
```diff
@@ -1994,7 +1992,7 @@ def test_no_attachments(self, save_mock):
         eq_(save_mock.called, False, save_mock.call_args_list)
 
     @override_settings(REVIEWER_ATTACHMENTS_PATH=ATTACHMENTS_DIR)
-    @mock.patch('mkt.site.utils.LocalFileStorage.save')
+    @mock.patch(settings.DEFAULT_FILE_STORAGE + '.save')
```
diox: Clever Girl

mkt/webapps/cron.py
```diff
@@ -118,12 +119,15 @@ def clean_old_signed(seconds=60 * 60):
...
-    for path in os.listdir(root):
-        full = os.path.join(root, path)
-        age = time.time() - os.stat(full)[stat.ST_ATIME]
-        if age > seconds:
-            log.debug('Removing signed app: %s, %dsecs old.' % (full, age))
-            shutil.rmtree(full)
+    for nextroot, dirs, files in walk_storage(root):
+        for fn in files:
+            full = os.path.join(nextroot, fn)
+            now = datetime.utcnow() if storage_is_remote() else datetime.now()
```
diox: Alternatively, before the loop:
`now_func = datetime.utcnow if storage_is_remote() else datetime.now`
then in the loop, just:
`now = nowfunc()`

diox: Also, a comment would be nice as it's not obvious

diox:
r+wc+squash, nice work!

lib/crypto/packaged.py
```diff
@@ -116,10 +116,8 @@ def _no_sign(src, dest):
     # If this is a local development instance, just copy the file around
     # so that everything seems to work locally.
     log.info('Not signing the app, no signing server is active.')
-    dest_dir = os.path.dirname(dest)
-    if not os.path.exists(dest_dir):
-        os.makedirs(dest_dir)
-    shutil.copy(src, dest)
+    with storage.open(dest, 'w') as destf:
+        shutil.copyfileobj(open(src), destf)
```
diox: This seems wrong, and breaks a few tests when remote storage is enabled. The `src` here will be the `file_path` of a `File` instance, which is supposed to be using the storage already. So we should be using `storage.open()` here.

diox: I'm taking care of this in #3214 so you can merge as is

washort:
Merged as 308069a.

PR #3212, opened 2015-08-03
Convert facets to aggregations (bug 1116800)

robhudson:


mkt/stats/views.py
```diff
@@ -299,15 +299,15 @@ def get_query(self, metric, field, app_id=None):
...
+            count = agg.get('count', 0)
 
-            # We filter out facets with count=0 to avoid returning things
+            # We filter out aggs with count=0 to avoid returning things
             # like `'max': u'-Infinity'`.
             if count > 0:
-                for field in ('max', 'mean', 'min', 'std_deviation',
-                              'sum_of_squares', 'total', 'variance'):
-                    value = facet.get(field)
+                for field in ('avg', 'max', 'min', 'std_deviation',
```
diox: We're exposing these fields directly in the API and one of them (`mean`) is no longer present, isn't that going to be a problem ?

diox: Also, tests were not covering this ?

robhudson: Yeah, I couldn't find any tests other than to test the status codes and permission stuff. I could rename the changed fields so it remains consistent. (And add a test?) This was just a quick work at the end of the day.

ngokevin:
Going to close for age.

PR #3214, opened 2015-08-04
Fix more tests when using remote storage (bug 1167271) 

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1167271

### DIFF
--- a/lib/crypto/packaged.py
+++ b/lib/crypto/packaged.py
@@ -88,9 +88,8 @@ def _sign_app(src, dest, ids, reviewer, tempname):
     except:
         log.error('App signing failed', exc_info=True)
         raise SigningError('App signing failed')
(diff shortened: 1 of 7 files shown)


lib/crypto/packaged.py
```diff
@@ -117,7 +117,7 @@ def _no_sign(src, dest):
     # so that everything seems to work locally.
     log.info('Not signing the app, no signing server is active.')
     with storage.open(dest, 'w') as destf:
-        shutil.copyfileobj(open(src), destf)
+        shutil.copyfileobj(storage.open(src), destf)
```
robhudson: If you use `with` it will close the file for you.

diox: Fixed

washort:
r+

PR #3221, opened 2015-08-06
Adjust code to use public or private storage (bug 1191449)

robhudson:
This is a first pass.

### DIFF
--- a/lib/crypto/packaged.py
+++ b/lib/crypto/packaged.py
@@ -5,7 +5,6 @@ import tempfile
 from base64 import b64decode
 
 from django.conf import settings
-from django.core.files.storage import default_storage as storage
 
 import commonware.log
 import requests
@@ -14,6 +13,8 @@ from django_statsd.clients import statsd
 from signing_clients.apps import JarExtractor
 
 from mkt.versions.models import Version
+from mkt.site.storage_utils import (local_storage, private_storage,
+                                    public_storage)
 
 
 log = commonware.log.getLogger('z.crypto')
@@ -23,10 +24,16 @@ class SigningError(Exception):
     pass
 
 
-def sign_app(src, dest, ids, reviewer=False):
+def sign_app(src, dest, ids, reviewer=False, local=False):
+    """
+    Sign a packaged app.
+
+    If `local` is True, we never copy the signed package to remote storage.
+
+    """
     tempname = tempfile.mktemp()
     try:
-        return _sign_app(src, dest, ids, reviewer, tempname)
+        return _sign_app(src, dest, ids, reviewer, tempname, local)
     finally:
         try:
             os.unlink(tempname)
@@ -35,7 +42,7 @@ def sign_app(src, dest, ids, reviewer=False):
             pass
 
 
-def _sign_app(src, dest, ids, reviewer, tempname):
+def _sign_app(src, dest, ids, reviewer, tempname, local=False):
     """
     Generate a manifest and signature and send signature to signing server to
     be signed.
@@ -88,7 +95,15 @@ def _sign_app(src, dest, ids, reviewer, tempname):
     except:
         log.error('App signing failed', exc_info=True)
         raise SigningError('App signing failed')
-    with open(tempname) as temp_f, storage.open(dest, 'w') as dest_f:
+
+    storage = public_storage  # By default signed packages are public.
+    if reviewer:
+        storage = private_storage
+    elif local:
+        storage = local_storage
+
+    with local_storage.open(tempname) as temp_f, \
+            storage.open(dest, 'w') as dest_f:
         shutil.copyfileobj(temp_f, dest_f)
 
 
@@ -115,7 +130,7 @@ def _no_sign(src, dest_path):
     # If this is a local development instance, just copy the file around
     # so that everything seems to work locally.
     log.info('Not signing the app, no signing server is active.')
-    with storage.open(dest_path, 'w') as dest_f:
+    with local_storage.open(dest_path, 'w') as dest_f:
         shutil.copyfileobj(src, dest_f)
 
 
@@ -140,6 +155,8 @@ def sign(version_id, reviewer=False, resign=False, **kw):
     path = (file_obj.signed_reviewer_file_path if reviewer else
             file_obj.signed_file_path)
 
+    storage = private_storage if reviewer else public_storage
+
     if storage.exists(path) and not resign:
         log.info('[Webapp:%s] Already signed app exists.' % app.id)
         return path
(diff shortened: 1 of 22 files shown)


lib/crypto/tests.py
```diff
@@ -16,6 +15,7 @@
 from lib.crypto.receipt import crack, sign, SigningError
 from mkt.site.storage_utils import copy_to_storage
 from mkt.site.fixtures import fixture
+from mkt.site.storage_utils import private_storage as storage
```
diox: We should avoid `import <something> as <somethingelse>` if we can IMHO, it gets confusing. Especially so in this case.

mkt/commonplace/management/commands/deploy_build_id.py
```diff
@@ -3,10 +3,10 @@
 import sys
 
 from django.conf import settings
-from django.core.files.storage import default_storage as storage
 from django.core.management.base import BaseCommand
 
 from mkt.commonplace.models import DeployBuildId
+from mkt.site.storage_utils import local_storage
```
diox: I love that we have these explicit storage classes now. private for what must be private, public for what must be public, local for what absolutely needs to stay local, it's awesome. :+1:

mkt/site/monitors.py
```diff
@@ -240,7 +241,7 @@ def package_signer():
         msg = 'Error on package signing (%s): %s' % (destination, e)
         return msg, msg
     finally:
-        storage.delete(signed_path)
+        local_storage.delete(signed_path)
```
diox: So that part makes sense as `signed_path` is a local `tempfile`, but it's then used as the `dest` argument when calling `sign_app()`, which has this:
```
storage = private_storage if reviewer else public_storage
...
with open(tempname) as temp_f, storage.open(dest, 'w')
```

So the `signed_path` will be opened by `sign_app()` using remote storage. Not sure how to fix this. Maybe consider copying the file to remote (public) storage first ?

mkt/developers/tasks.py
```diff
@@ -121,7 +122,7 @@ def run_validator(file_path, url=None):
     # uploaded file is on the local filesystem.
     temp_path = tempfile.mktemp()
     with open(temp_path, 'wb') as local_f:
-        with storage.open(file_path) as remote_f:
+        with private_storage.open(file_path) as remote_f:
```
diox: So, that means `File.file_path` and `FileUpload.path` are always private (since they are passed to `run_validator()`, but `File.signed_file_path` is public ? That makes sense to me, but I'm not sure it is at the moment.

lib/crypto/packaged.py
```diff
@@ -115,7 +118,7 @@ def _no_sign(src, dest_path):
     # If this is a local development instance, just copy the file around
     # so that everything seems to work locally.
     log.info('Not signing the app, no signing server is active.')
-    with storage.open(dest_path, 'w') as dest_f:
+    with private_storage.open(dest_path, 'w') as dest_f:
```
diox: It's in `_no_sign()` so it does not matter for prod, but we probably want `public_storage` here. We are making a fake sign and the resulting package will be used as `File.signed_file_path`.

mkt/webapps/cron.py
```diff
@@ -124,11 +124,11 @@ def clean_old_signed(seconds=60 * 60):
     for nextroot, dirs, files in walk_storage(root):
         for fn in files:
             full = os.path.join(nextroot, fn)
-            age = now() - storage.modified_time(full)
+            age = now() - private_storage.modified_time(full)
```
diox: See the comment about delete below

lib/crypto/packaged.py
diox: Unless the comment above is wrong and it's sometimes used in a different context ? I hope not :)

mkt/webapps/cron.py
```diff
@@ -124,11 +124,11 @@ def clean_old_signed(seconds=60 * 60):
...
             if age.total_seconds() > seconds:
                 log.debug('Removing signed app: %s, %dsecs old.' % (
                     full, age.total_seconds()))
-                storage.delete(full)
+                private_storage.delete(full)
```
diox: For a second I thought it was wrong, but no, these are reviewer signed packages, so carry on :)

diox:
Nice work. Left a few comments on things that seemed wrong or weird, but it looks like a good first pass.

mkt/developers/tasks.py
diox: (I wondered for a second if we needed to explicitly copy paths from one storage to another, but I guess we don't since signing will take care of that ? Unless something specific needs to be done for hosted apps, but I don't think so)

lib/crypto/tests.py
robhudson: I agree.

mkt/developers/tasks.py
robhudson: Yeah, I think we (or I) will have to walk the code path and make sure these are right once done. But this one seemed obvious enough to do it now.

mkt/commonplace/management/commands/deploy_build_id.py
robhudson: `\o/`

mkt/site/monitors.py
robhudson: I wonder if it's worth passing in a kwarg to tell it to use `local_storage` instead of these? It seems inefficient to copy it to S3 just to delete it right after.

lib/crypto/packaged.py
robhudson: Wouldn't local development use `local_storage` in that case then?

robhudson:
I did a pass through the signing and file upload stuff. My goal was any file unsigned should be private. Signed files are public unless they are reviewer signed packages. Also, if the file changes from a public status to disabled the files need to move from public to private and vice versa.

mkt/files/models.py
```diff
@@ -103,7 +105,7 @@ def addon(self):
     def generate_hash(self, filename=None):
         """Generate a hash for a file."""
         hash = hashlib.sha256()
-        with storage.open(filename or self.file_path, 'rb') as obj:
+        with private_storage.open(filename or self.file_path, 'rb') as obj:
```
robhudson: I think I need to look more closely at this. If the `File` is public this will be incorrect. Assuming the `mv` stuff below works as it should we can toggle the storage backend based on `file_ == mkt.STATUS_DISABLED`.

mkt/site/monitors.py
robhudson: I updated to use `local_storage`.

mkt/files/models.py
diox: On the other hand, we should only call it once when creating the `File` instance. We re-use `File` instances for hosted apps, but I wonder if we really use their hash (or their `File` for that matter)

diox:
r+ there are probably adjustments to make but that's a very good first pass.

PR #3222, opened 2015-08-06
Adds tags for restricted websites (bugs 1190549, 1190550).

chuckharmston:
r? @ngokevin

### DIFF
--- a/mkt/search/forms.py
+++ b/mkt/search/forms.py
@@ -7,6 +7,7 @@ from mkt.constants import (CATEGORY_CHOICES, CATEGORY_CHOICES_DICT,
                            CATEGORY_REDIRECTS, TARAKO_CATEGORIES_MAPPING,
                            TARAKO_CATEGORY_CHOICES)
 from mkt.constants.applications import DEVICE_LOOKUP
+from mkt.constants.regions import REGIONS_CHOICES
 
 
 SORT_CHOICES = [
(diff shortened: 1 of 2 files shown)


mkt/search/forms.py
```diff
@@ -71,7 +72,10 @@
     ('featured-game-adventure', 'featured-game-adventure'),
     ('featured-game-puzzle', 'featured-game-puzzle'),
     ('featured-game-strategy', 'featured-game-strategy'),
+    ('featured-website', 'featured-website'),
 ]
+REG_TAG = lambda r: ('featured-website-%s' % r, 'featured-website-%s' % r)
+TAG_CHOICES += [REG_TAG(region) for region in dict(REGIONS_CHOICES).keys()]
```
ngokevin: wouldn't mind a quick test

mkt/search/forms.py
```diff
@@ -71,7 +72,10 @@
     ('featured-game-adventure', 'featured-game-adventure'),
     ('featured-game-puzzle', 'featured-game-puzzle'),
     ('featured-game-strategy', 'featured-game-strategy'),
+    ('featured-website', 'featured-website'),
 ]
+REG_TAG = lambda r: ('featured-website-%s' % r, 'featured-website-%s' % r)
```
ngokevin: don't know if we need an intermediate lambda, it might be more readable without it, although i understand what it's doing.

```[('featured-website-%s' % r, featured-website-%s' % r) for r in dict(REGION_CHOICES.keys())]

ngokevin: pep don't like it either

chuckharmston: Seemed like it'd be more readable this way if it had to be multiple lines, but maybe not. I've been writing a lot of JS lately.

ngokevin:
r+

PR #3223, opened 2015-08-06
Adds featured websites to FeedView response (bugs 1190551, 1190553, 1190555).

chuckharmston:
r? @robhudson @ngokevin

### DIFF
--- a/mkt/feed/tests/test_views.py
+++ b/mkt/feed/tests/test_views.py
@@ -24,8 +24,11 @@ from mkt.fireplace.tests.test_views import assert_fireplace_app
 from mkt.operators.models import OperatorPermission
 from mkt.site.fixtures import fixture
 from mkt.site.tests import app_factory, ESTestCase, TestCase
+from mkt.tags.models import Tag
 from mkt.users.models import UserProfile
 from mkt.webapps.models import Preview, Webapp
+from mkt.websites.models import Website
+from mkt.websites.utils import website_factory
 
 
 TEST_DIR = os.path.dirname(os.path.abspath(__file__))
@@ -1320,6 +1323,27 @@ class TestFeedView(BaseTestFeedESView, BaseTestFeedItemViewSet):
         self.carrier = 'telefonica'
         self.region = 'restofworld'
 
+    def _make_website(self, tag=None):
+        new_site = website_factory(**{
+            'title': 'something',
+            'categories': [],
+        })
+        if tag:
+            new_site.keywords.add(tag)
+        return new_site
+
+    def featured_mow_factory(self, n_row=0, n_www=0):
+        websites = []
+        www = Tag.objects.get_or_create(tag_text='featured-website')[0]
+        for n in xrange(n_www):
+            websites.append(self._make_website(tag=www.pk))
+        row = Tag.objects.get_or_create(
+            tag_text='featured-website-restofworld')[0]
+        for n in xrange(n_row):
+            websites.append(self._make_website(tag=row.pk))
+        self.reindex(Website)
+        return websites
+
     def _get(self, client=None, **kwargs):
         self._refresh()
         client = client or self.anon
@@ -1341,7 +1365,9 @@ class TestFeedView(BaseTestFeedESView, BaseTestFeedItemViewSet):
         res, data = self._get()
         eq_(res.status_code, 200)
         eq_(len(data['objects']), len(feed_items))
+        eq_(len(data['websites']), 0)
         ok_(statsd_mock.called)
+        return res, data
 
     def test_200_authed(self):
         feed_items = self.feed_factory()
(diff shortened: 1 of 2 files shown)


mkt/feed/views.py
```diff
@@ -811,6 +815,35 @@ def get_es_feed_element_query(self, sq, feed_items):
 
         return sq.filter(es_filter.Bool(should=filters))[0:len(feed_items)]
 
+    def _featured_website_query(self, region=None, n=11):
+        """
+        Query for n featured websites matching an optional region.
+        """
+        TAG = 'featured-website'
```
ngokevin:
```python
TAG = '%s-%s' % (TAG, region.slug) if region else 'featured-website'
```

mkt/feed/views.py
```diff
@@ -811,6 +815,35 @@ def get_es_feed_element_query(self, sq, feed_items):
...
+        if region:
+            TAG = '%s-%s' % (TAG, region.slug)
+        daily_seed = int(datetime.now().strftime('%Y%m%d'))
+        mow_query = query.Q(
+            'function_score',
+            filter=es_filter.Bool(should=[es_filter.Term(tags=TAG)]),
```
ngokevin: ```must```?

mkt/feed/views.py
```diff
@@ -811,6 +815,35 @@ def get_es_feed_element_query(self, sq, feed_items):
 
         return sq.filter(es_filter.Bool(should=filters))[0:len(feed_items)]
 
+    def _featured_website_query(self, region=None, n=11):
```
ngokevin: nitpick, but calling it ```_get_featured_websites``` might make it more clear that it's actually executing the query

mkt/feed/views.py
```diff
@@ -811,6 +815,35 @@ def get_es_feed_element_query(self, sq, feed_items):
...
+        """
+        Get up to n featured MOWs for the request's region.
+
+        If less than n are available, make up the difference with globally-
+        featured MOWs.
+        """
+        featured = list(self._featured_website_query(region=self.request.REGION))
+        if len(featured) < n:
+            diff = n - len(featured)
+            featured += list(self._featured_website_query(n=diff))
```
ngokevin: any chance of dupes?

ngokevin: i'd let the elasticsearch query handle this rather than firing two. you should be able to do this by weighting the region tag, but still allowing for normal ```featured-website``` tags with ```should```

mkt/feed/views.py
chuckharmston: Yeah, good call. This was the remnants of me trying to be getting it into one too-cute query.

mkt/feed/views.py
chuckharmston: That's exactly what I tried to do, but it didn't act predictably. I'm going to try it again with a fresh brain, because I don't see why it wouldn't work.

chuckharmston: Got it, I was missing the filter on the boost, so it was applying it to everything.

mkt/feed/views.py
chuckharmston: I ended up merging it into a single method, so this wasn't necessary at all.

mkt/feed/views.py
```diff
@@ -912,8 +940,14 @@ def _get(self, request, rest_of_world=False, original_region=None,
             return self._handle_empty_feed(feed_ok, region, request, args,
                                            kwargs)
 
-        return response.Response({'meta': meta, 'objects': feed_items},
-                                 status=status.HTTP_200_OK)
+        with statsd.timer('mkt.feed.view.feed_website_query'):
```
robhudson: :+1:

mkt/feed/tests/test_views.py
```diff
@@ -1374,6 +1400,39 @@ def test_shelf_top(self):
...
+        """
+        self.feed_factory()
+        self.featured_mow_factory(n_row=11, n_www=1)
+        res, data = self._get()
+        tags = [mow['keywords'][0] for mow in data['websites']]
+        eq_(len(data['websites']), 11)
+        eq_(tags.count('featured-website-restofworld'), 11)
+        eq_(tags.count('featured-website'), 0)
+
+    def test_websites_random_seed(self):
```
ngokevin: have a test to mock the date and check results are different

mkt/feed/tests/test_views.py
```diff
@@ -1374,6 +1400,39 @@ def test_shelf_top(self):
...
+    def test_websites(self):
+        self.feed_factory()
+        self.featured_mow_factory(n_row=6, n_www=5)
+        res, data = self._get()
+        tags = [mow['keywords'][0] for mow in data['websites']]
+        eq_(len(data['websites']), 11)
+        eq_(tags.count('featured-website-restofworld'), 6)
+        eq_(tags.count('featured-website'), 5)
+
+    def test_websites_prefer_regional(self):
```
ngokevin: have a test where half the websites are regional and then the bottom half is global

ngokevin:
r+wc

mkt/feed/tests/test_views.py
chuckharmston: `test_websites` does that—"restofworld" is treated like a region.

mkt/feed/tests/test_views.py
chuckharmston: https://github.com/mozilla/zamboni/pull/3223/files#diff-46b7ba21e1c662b3270b812d6a583b8fR1437

PR #3224, opened 2015-08-07
Support both external-app and moz-external-app permissions (bug 1192080)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1192080

This reverts my earlier commit that made `"external-app"` a special case and instead use the new validator commit that moves it and `"moz-external-app"` to the prerelease permissions set.


eviljeff:
r+

PR #3228, opened 2015-08-10
When storing public files on S3, cache them for a year (bug 1189766)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1189766


diox:
Rationale: for all files that are public, when we want to modify the file, we already change the instance entirely (`File`, `FileUpload`, `Preview`), the path (`File.file_path` vs `File.signed_file_path`) or add a query string (`Webapp` icons are served with `?modified`), so we should be able to serve caching headers for them globally.

We'll also need to set caching headers on redirects to them in bug 1189763.

robhudson:
Simple enough. :) r+

PR #3229, opened 2015-08-10
Remove Tarako checkbox and subscription model Bug 741679 and bug 1059385

eviljeff:


diox:
Is there any reason to have those 2 bugs in a single PR, and if not, can you split that ? Easier if we need to revert one of the bugs later.

eviljeff:
no reason really... I was working on something else that touched the same few lines and remembered about these bugs.

eviljeff:
Alright, changed this pull to be just bug 741679.

eviljeff:
Added a commit for the 2nd bug.  Both commits are independent of either other; they just touch adjacent lines of code.

diox:
r+

PR #3232, opened 2015-08-10
Add settings for s3dev (bug 1191449)

robhudson:


jasonthomas:
zamboni/blob/s3dev/sites/s3dev/settings_mkt.py#L160 needs to be removed.

r+wc

robhudson:
Moved to `settings_mkt.py`.

PR #3235, opened 2015-08-11
Explicitely floor() values in FeaturesBitField (bug 1172487)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1172487

Because the result of an integer division in python *2* is already floored, the math.ceil() call didn't have any effect, but it's wrong, our indexes begin at 0, so I removed it for clarity.

Also add a couple dumb tests regarding base64 conversion.


robhudson:
r+

PR #3242, opened 2015-08-12
Fix en-GB test now that it is supported (bug 1190115)

robhudson:


mkt/submit/tests/test_views.py
```diff
@@ -391,7 +391,7 @@ def test_short_locale(self):
         eq_(addon.versions.latest().supported_locales, 'es')
 
     def test_unsupported_detail_locale(self):
-        # This manifest has a locale code of "en-GB" which is unsupported, so
+        # This manifest has a locale code of "en-CA" which is unsupported, so
         # we default to "en-US".
         self.manifest = self.manifest_path('unsupported-default-locale.webapp')
```
chuckharmston: r-, it's "unsuppourted" in en-CA.

chuckharmston:
r+ if tests pass

PR #3249, opened 2015-08-15
use storage for pngcrush_image (bug 1191787)

washort:


mkt/developers/tasks.py
```diff
@@ -217,6 +217,8 @@ def pngcrush_image(src, hash_field='image_hash', **kw):
                   instance through set_modified_on
     """
     log.info('[1@None] Optimizing image: %s' % src)
+    with public_storage.open(src) as srcf, open(src, 'w') as destf:
```
diox: If you want to write to the full path on the local filesystem, I'd use our local_storage class so that the directories are all created for you if they don't exist. But I think I'd prefer for us to write a tempfile on the local storage instead of re-using the path, it's easier to understand and will help in the long term when we do path cleanup.

diox: btw, you should delete both local files when you are done.

mkt/developers/tasks.py
```diff
@@ -217,35 +218,40 @@ def pngcrush_image(src, hash_field='image_hash', **kw):
...
 
-        shutil.move(tmp_path, src)
-        log.info('Image optimization completed for: %s' % src)
-
         # Return hash for set_modified_on.
-        with open(src) as fd:
+        with open(tmp_path) as fd:
             image_hash = _hash_file(fd)
 
+        copy_to_storage(tmp_path, src, storage=public_storage)
```
robhudson: The arguments change in @diox's patch.

washort: guess i'll wait and rebase then. :)

robhudson: Or merge quick and he'll have to fix his. :grinning:

PR #3251, opened 2015-08-17
Make export_data task work with remote storage (bug 1191878)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1191878

### DIFF
--- a/mkt/site/storage_utils.py
+++ b/mkt/site/storage_utils.py
@@ -150,12 +150,17 @@ def walk_storage(path, topdown=True, onerror=None, followlinks=False,
         roots[:] = new_roots
 
 
-def copy_to_storage(src_path, dest_path, storage=private_storage):
+def copy_to_storage(src_path, dst_path, src_storage=local_storage,
+                    dst_storage=private_storage):
     """
-    Copy a local path (src_path) to a store path (dest_path).
+    Copy a path (src_path) from a storage (src_storage) to a path (dst_path)
+    on a different storage (dst_storage).
+
+    Defaults to copying from the local storage to the private storage.
     """
-    with open(src_path) as src_f, storage.open(dest_path, 'w') as dest_f:
-        shutil.copyfileobj(src_f, dest_f)
+    with src_storage.open(src_path) as src_f, \
+            dst_storage.open(dst_path, 'w') as dst_f:
+        shutil.copyfileobj(src_f, dst_f)
 
 
 def copy_stored_file(src_path, dest_path,
--- a/mkt/webapps/tasks.py
+++ b/mkt/webapps/tasks.py
@@ -5,6 +5,7 @@ import logging
 import os
 import shutil
 import subprocess
+import tempfile
 
 from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
@@ -31,7 +32,9 @@ from mkt.reviewers.models import EscalationQueue, RereviewQueue
 from mkt.site.decorators import set_task_user, use_master
 from mkt.site.helpers import absolutify
 from mkt.site.mail import send_mail_jinja
-from mkt.site.storage_utils import public_storage
+from mkt.site.storage_utils import (copy_to_storage, local_storage,
+                                    private_storage, public_storage,
+                                    walk_storage)
 from mkt.site.utils import JSONEncoder, chunked
 from mkt.users.models import UserProfile
 from mkt.users.utils import get_task_user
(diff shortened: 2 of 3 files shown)


diox:
WIP: it's not entirely working right now, and tests don't show that, so I need to fix it.

mkt/webapps/tasks.py
```diff
@@ -383,45 +382,85 @@ def export_data(name=None):
...
 
-    # Put some .txt files in place.
-    cmd = ['tar', 'czf', target_file, '-C',
-           settings.DUMPED_APPS_PATH] + files
-    task_log.info(u'Creating dump {0}'.format(target_file))
-    subprocess.call(cmd)
-    return target_file
+    # Clean-up.
+    local_target_file.close()
+    # rm_directory(local_source_dir)
```
diox: I need to uncomment that

mkt/site/storage_utils.py
```diff
@@ -150,12 +150,17 @@ def walk_storage(path, topdown=True, onerror=None, followlinks=False,
...
+    Copy a path (src_path) from a storage (src_storage) to a path (dst_path)
+    on a different storage (dst_storage).
+
+    Defaults to copying from the local storage to the private storage.
     """
-    with open(src_path) as src_f, storage.open(dest_path, 'w') as dest_f:
-        shutil.copyfileobj(src_f, dest_f)
+    with src_storage.open(src_path) as src_f, \
+            dst_storage.open(dst_path, 'w') as dst_f:
+        shutil.copyfileobj(src_f, dst_f)
```
robhudson: I feel like I've written this same code before. But maybe it was another method and not this one.

mkt/webapps/tasks.py
```diff
@@ -333,8 +336,6 @@ def unindex_webapps(ids, **kw):
 @task
 def dump_app(id, **kw):
     from mkt.webapps.serializers import AppSerializer
-    # Because @robhudson told me to.
```
robhudson: The blame is finally lifted.

mkt/webapps/tasks.py
```diff
@@ -383,45 +382,85 @@ def export_data(name=None):
...
         name = today
-    root = settings.DUMPED_APPS_PATH
-    directories = ['apps']
-    for directory in directories:
-        rm_directory(os.path.join(root, directory))
-    files = directories + compile_extra_files(date=today)
+
+    # Clean up the path where we'll store the individual json files from each
+    # app dump.
+    for dirpath, dirnams, filenames in walk_storage(settings.DUMPED_APPS_PATH):
```
robhudson: How does `walk_storage` know which bucket to walk?

diox: It defaults to `private_storage`. I debated explicitly passing it as argument anyway, but thought this function was hard enough to read as it is.

mkt/webapps/tasks.py
```diff
@@ -383,45 +382,85 @@ def export_data(name=None):
...
+    task_log.info(u'Creating dump {0}'.format(local_target_file.name))
+    subprocess.call(cmd)
 
-    if not os.path.exists(target_dir):
-        os.makedirs(target_dir)
+    # Now copy the local tgz to the public storage.
+    remote_target_filename = os.path.join(
+        settings.DUMPED_APPS_PATH, 'tarballs', '%s.tgz' % tarball_name)
+    copy_to_storage(local_target_file.name, remote_target_filename,
+                    dst_storage=public_storage)
```
robhudson: you don't need `src_storage` here?

mkt/site/storage_utils.py
robhudson: src_storage defaulting to local storage seems odd to me when almost everythign else defaults to private.

robhudson:
r+wc. I didn't think much on all the paths strings but I assume they're good.

mkt/webapps/tasks.py
diox: `src_storage` defaults to `local_storage` so like above I didn't set arguments I didn't need to set

mkt/site/storage_utils.py
diox: It seems strange, but it actually makes sense as the method was originally designed to copy stuff from local filesystem to the remote storage. The remote storage does default to `private_storage`.

diox:
Tests are now passing when configured with a remote storage, and I've added the explicit `storage=private_storage` argument to `walk_storage()` calls.

PR #3253, opened 2015-08-18
Add basic model and submission API for FxOS Add-ons (bug 1195817)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1195817

This is not testable using the API yet, but it's just the basics to get us started. The idea is to follow the same process as apps / langpacks during the submission:
- First you go through the validation API
- Then you wait for it to be processed
- Then if it's valid you submit by POSTing to the extension API


ngokevin:
How come we aren't calling them addons?

ngokevin:
How come we aren't calling them addons? Because the previous widespread use of its name, or to leave room for desktop addons?

diox:
I debated calling them `Addons` internally, but while https://github.com/mozilla/zamboni/pull/3239 is not merged that seems... dangerous. Externally the name that will be exposed will be "add-ons".

ngokevin:
Ah, I didn't know that wasn't merged yet. I might prefer they be called the same thing everywhere to start clean, maybe we should look at getting that PR in?

mkt/extensions/models.py
```diff
@@ -0,0 +1,157 @@
...
+    default_language = models.CharField(default=settings.LANGUAGE_CODE,
+                                        max_length=10)
+    manifest = JSONField()
+    version = models.CharField(max_length=255, default='')
+
+    # Fields that can be modified using the API.
+    authors = models.ManyToManyField('users.UserProfile')
+    name = TranslatedField(default=None)
+    slug = models.CharField(max_length=35, unique=True)
+    status = status = models.PositiveSmallIntegerField(
```
ngokevin: typo?

ngokevin: @chuckharmston advocated a separate model for submission where most of the fields have ```null=True```. It allows flexibility for submission to be done parts at a time rather than needing submit everything at once. Chuck, does that logic still apply here, or do you think this project is more specced out enough we don't have to?

chuckharmston: I'm not familiar enough with the needs here to say.  The reason we did it for website submission is because there was a multi-step data collection process whose various validation rules would have made it really difficult to implement in one model.

diox:
mmm I fear there might be potential issues forcing us to delay or even revert #3239 depending on how the migration goes, so I wasn't really a fan of depending on it.

I can *try* to ignore that PR and `s/Extension/Addon/` and see how it goes though.

mkt/extensions/models.py
diox: I studied that implementation but I don't think it's necessary here. Unlike `Website`s (and like `Webapp`s and `LangPack`s) there are 2 steps to the submission process, one is submitting the file, which gives you a `FileUpload` that validation runs on, and then the actual object creation, that sets all the fields the model will need.

diox: Also, yeah, typo :)

ngokevin:
How about let's continue with Extensions and look to rename it once everything settles.

mkt/extensions/tests/test_models.py
```diff
@@ -0,0 +1,96 @@
+# -*- coding: utf-8 -*-
+import mock
+from nose.tools import eq_, ok_
+
+from django.forms import ValidationError
+
+from mkt.files.tests.test_models import UploadCreationMixin, UploadTest
+from mkt.extensions.models import Extension
```
ngokevin: nit: alphabetical imports

mkt/extensions/models.py
```diff
@@ -0,0 +1,157 @@
...
+            for lang_code, lang_name in languages.items()]
+
+
+LANGUAGE_CHOICES = lazy(_make_language_choices, list)(settings.LANGUAGES)
+
+
+class Extension(ModelBase):
+    # Fields for which the manifest is the source of truth - can't be
+    # overridden by the API.
+    default_language = models.CharField(default=settings.LANGUAGE_CODE,
```
ngokevin: i'm sure we need it, but i'm wondering what cases would we need attach languages to FxOS addons?

mkt/extensions/models.py
```diff
@@ -0,0 +1,157 @@
...
+        raise NotImplementedError
+
+    @property
+    def manifest_url(self):
+        raise NotImplementedError
+
+    def clean_slug(self):
+        return clean_slug(self, slug_field='slug')
+
+    def __unicode__(self):
```
ngokevin: nit: maybe alphabetize method names

mkt/extensions/models.py
```diff
@@ -0,0 +1,157 @@
...
+    def is_public(self):
+        return self.status == STATUS_PUBLIC
+
+    def get_package_path(self):
+        raise NotImplementedError
+
+    def get_minifest_contents(self, force=False):
+        raise NotImplementedError
+
+    def get_manifest_json(self):
```
ngokevin: do we need a proxy method for this?

mkt/extensions/utils.py
```diff
@@ -0,0 +1,81 @@
...
+
+        required_fields = ('name', 'version')
+        for field in required_fields:
+            if not data.get(field):
+                raise forms.ValidationError(
+                    _(u'The "%s" field is missing or empty in the'
+                      u' add-on manifest.' % field))
+            output[field] = data[field]
+
+        allowed_fields = ('author', 'default_locale', 'description', 'icons')
```
ngokevin: we going to add a description field to the model?

ngokevin:
lgtm, not much to review since it's just the foundation

mkt/extensions/models.py
diox: While the code to parse translations for `name`, and `description` does not exist yet, they can be translated and therefore we need a `default_language`. It's coming from the manifest.

mkt/extensions/models.py
diox: Not now, but I suspect we will :) I'll remove it for now.

mkt/extensions/utils.py
diox: Not now, but chrome extensions have it (and I suspect we will), so I allowed the field to be present

mkt/extensions/models.py
ngokevin: that there's premature refactorization

PR #3254, opened 2015-08-18
Update babel for better price localisation (bug 1185250)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1185250

This is mostly to fix Dutch where the Euro symbol was being placed after the amount instead of in front of it.

### DIFF
--- a/mkt/prices/models.py
+++ b/mkt/prices/models.py
@@ -41,17 +41,7 @@ def default_providers():
 def price_locale(price, currency):
     lang = translation.get_language()
     locale = get_locale_from_lang(lang)
-    pricestr = numbers.format_currency(price, currency, locale=locale)
-    if currency == 'EUR':
-        # See bug 865358. EU style guide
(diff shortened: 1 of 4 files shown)


mkt/prices/models.py
```diff
@@ -42,15 +42,6 @@ def price_locale(price, currency):
...
-    if currency == 'EUR':
-        # See bug 865358. EU style guide
-        # (http://publications.europa.eu/code/en/en-370303.htm#position)
-        # disagrees with Unicode CLDR on placement of Euro symbol.
-        bare = pricestr.strip(u'\xa0\u20ac')
-        if lang.startswith(('en', 'ga', 'lv', 'mt')):
-            return u'\u20ac' + bare
-        else:
-            return bare + u'\xa0\u20ac'
     return pricestr
```
diox: No need for the `pricestr` variable then, you can just `return`.

diox:
r+ let's try it

PR #3256, opened 2015-08-18
Upgrade to Django 1.7.10

chuckharmston:
https://www.djangoproject.com/weblog/2015/aug/18/security-releases/

Getting tests started.


robhudson:
Aside from the `foo.bar` file, r+

chuckharmston:
Yeah, I don't know where that came from. It's this weird floating commit on master.

PR #3258, opened 2015-08-18
remove redundant storage_utils bits

washort:


mkt/site/storage_utils.py
```diff
@@ -158,35 +158,26 @@ def copy_to_storage(src_path, dst_path, src_storage=local_storage,
 
     Defaults to copying from the local storage to the private storage.
     """
-    with src_storage.open(src_path) as src_f, \
-            dst_storage.open(dst_path, 'w') as dst_f:
-        shutil.copyfileobj(src_f, dst_f)
+    copy_stored_file(src_path, dst_path, src_storage=src_storage,
+                     dest_storage=dst_storage)
```
robhudson: may as well make the calling args consistent. One is `dest_storage`, the other `dst_storage`.

robhudson:
r+

But I'd prefer to do away with one or the other. The difference is subtle. I think having a single method with explicit arg when called would be clearer.

washort:
Sure, removing one would be the next step. This was just an easy way to reduce subtlety.

PR #3263, opened 2015-08-19
Improve documentation regarding curling usage to manipulate the API

diox:


eviljeff:
r+ on formatting.  Not familiar enough with curling to comment on that.

PR #3266, opened 2015-08-19
remove uses of default_storage in files, langpacks

washort:


mkt/files/cron.py
```diff
@@ -4,13 +4,13 @@
...
 from django.core.cache import cache
-from django.core.files.storage import default_storage as storage
 
 import commonware.log
 import cronjobs
 
 from mkt.files.models import FileValidation
-from mkt.site.storage_utils import storage_is_remote, walk_storage
+from mkt.site.storage_utils import (storage_is_remote,
+                                    private_storage, walk_storage)
```
robhudson: p > s

mkt/files/helpers.py
```diff
@@ -92,8 +92,12 @@ def extract(self):
         Will make all the directories and expand the files.
         Raises error on nasty files.
         """
+        if self.file.status == mkt.STATUS_PUBLIC:
```
robhudson: Should we use `in mkt.LISTED_STATUSES` here? I'm thinking of the approved apps that aren't listed but still should exist in the public storage. I'd have to think harder about approved but private apps.

robhudson:
No test changes required here?

washort:
test failures here, still working on 'em

mkt/files/helpers.py
washort: when are files copied to public? is it done anywhere other than signing?

mkt/langpacks/models.py
```diff
@@ -121,7 +121,7 @@ def handle_file_operations(self, upload):
         upload.path = smart_path(nfd_str(upload.path))
         if not self.uuid:
             self.reset_uuid()
-        if storage.exists(self.filename):
+        if private_storage.exists(self.filename):
```
diox: I think this was wrong to begin with and should be `self.file_path`. Furthermore, it should be `public_storage`, since this is the final, signed version we are looking at here.

mkt/langpacks/models.py
```diff
@@ -150,11 +150,12 @@ def sign_and_move_file(self, upload):
...
                 # This will read the upload.path file, generate a signature
                 # and write the signed file to self.file_path.
-                sign_app(storage.open(upload.path), self.file_path, ids)
+                sign_app(private_storage.open(upload.path),
+                         self.file_path, ids)
             except SigningError:
                 log.info('[LangPack:%s] Signing failed' % self.pk)
-                if storage.exists(self.file_path):
-                    storage.delete(self.file_path)
+                if private_storage.exists(self.file_path):
```
diox: This should be public, not private, same below.

mkt/langpacks/models.py
```diff
@@ -150,11 +150,12 @@ def sign_and_move_file(self, upload):
             try:
                 # This will read the upload.path file, generate a signature
                 # and write the signed file to self.file_path.
-                sign_app(storage.open(upload.path), self.file_path, ids)
+                sign_app(private_storage.open(upload.path),
```
diox: That one looks good, it's initially private since it comes from a `FileUpload`

diox:
I would have expected langpack test changes as well.

mkt/files/helpers.py
robhudson: Yeah, when they are signed. I believe that's it. Unsigned files should be private.

diox:
@washort : `langpacks/tests/test_models.py` and `langpacks/tests/test_views.py` are still using `default_storage`, did you check those ? I don't think this is right.

mkt/langpacks/tests/test_models.py
```diff
@@ -136,7 +136,7 @@ def test_upload_new(self):
         eq_(langpack.filename, '%s-%s.zip' % (langpack.uuid, langpack.version))
         ok_(langpack.filename in langpack.file_path)
         ok_(langpack.file_path.startswith(langpack.path_prefix))
-        ok_(storage.exists(langpack.file_path))
+        ok_(private_storage.exists(langpack.file_path))
```
diox: We sign langpacks as soon as they are uploaded so I think it should be public

PR #3267, opened 2015-08-19
Add Validation & Submission API for FxOS Add-ons (bug 1196223)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1196223

Goes nicely with https://github.com/andymckay/curling/pull/28 (but it's also testable using curl + shared secret, as long as you use `--data-binary` and set the expected headers correctly)

### DIFF
--- a/docs/api/topics/addon_submission.rst
+++ b/docs/api/topics/addon_submission.rst
@@ -18,8 +18,8 @@ Like apps, submitting an addon involves a few steps. The client must be logged
 in for all these steps and the user submitting the addon must have accepted the
 terms of use.
 
-1. :ref:`Validate your addon <validation-post-label>`. The validation will
-   return a validation id.
+1. :ref:`Validate your addon <addon_validation-post-label>`. The validation
+   will return a validation id.
 2. :ref:`Post your app <addon-post-label>` using the validation id.
    This will create an addon and populate the data with the
    contents of the manifest. It will return the current app data.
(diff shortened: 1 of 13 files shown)


mkt/webapps/views.py
```diff
@@ -75,7 +75,7 @@ def create(self, request, *args, **kwargs):
             uuid = request.DATA.get('manifest', '')
             is_packaged = False
         if not uuid:
-            raise serializers.ValidationError(
```
diox: This was throwing a 500 error before.

mkt/webapps/views.py
```diff
@@ -93,7 +93,10 @@ def create(self, request, *args, **kwargs):
             raise exceptions.PermissionDenied('You do not own that app.')
 
         # Create app, user and fetch the icon.
-        obj = Webapp.from_upload(upload, is_packaged=is_packaged)
+        try:
+            obj = Webapp.from_upload(upload, is_packaged=is_packaged)
+        except serializers.ValidationError as e:
```
diox: `from_upload` can return a `ValidationError`, if left uncatched it throws a 500 error like above, what we want is a 400.

mkt/extensions/urls.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework.routers import SimpleRouter
+
+from mkt.extensions.views import ExtensionViewSet, ValidationViewSet
+
+extensions = SimpleRouter()
+extensions.register(r'validation', ValidationViewSet,
+                    base_name='extension-validation')
+extensions.register(r'extension', ExtensionViewSet)
```
ngokevin: How about just ```^$```? We can just have the path ```/extensions/``` be the endpoint rather than ```/extensions/extension```.

diox: I thought about that, but then I would need to relocate `validation`, or forbid `validation` as a string, since we'll want `{prefix}/{slug}` for the detail API

docs/api/topics/addon_submission.rst
```diff
@@ -27,4 +27,76 @@ terms of use.
...
+    **Response**
+
+    Returns a :ref:`validation <addon_validation-response-label>` result.
+
+    :status 201: successfully created, processed.
+    :status 202: successfully created, still processing.
+
+.. _addon_validation-response-label:
+
+.. http:get:: /api/v2/apps/validation/(string:id)/
```
diox: Wrong copy/paste, it's `/api/v2/extensions/validation/(string:id)/`, I'll fix.

ngokevin:
lgtm, i guess indexers are coming next?

diox:
Next on my list is getting the list of addons you've developed and then the ES work yeah

PR #3269, opened 2015-08-19
Replace os.path.join(ROOT, ...) with call to path()

dchanm:
A couple lines were still using os.path.join to join ROOT and a path


diox:
Thanks

PR #3271, opened 2015-08-20
disallow feedback with only blank spaces in the API (bug 1066155)

luckylavish:


mkt/account/serializers.py
```diff
@@ -64,6 +64,8 @@ def validate(self, attrs):
         else:
             attrs['user'] = None
 
+        if attrs['feedback'].strip().__len__() < 1:
```
diox: Don't call `__len__` like that. Better call `len()` instead. But this is not even necessary, you can just do `if not attrs['feedback'].strip()`.

diox:
Please add a test, see `mkt/accounts/tests/test_views.py`

luckylavish:
closing this PR, I will create new PR with all required changes.

PR #3272, opened 2015-08-20
Make dump_user_installs_cron and associated code work with a remote storage (bug 1195267)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1195267

The app export code is super-similar, but I'm strongly resisting the urge to refactor because I can't afford to go to deep in yak shaving at the moment


washort:
r+

PR #3273, opened 2015-08-20
disallow feedback with only blank spaces in the API (bug 1066155)

luckylavish:


mkt/account/serializers.py
```diff
@@ -64,6 +64,9 @@ def validate(self, attrs):
         else:
             attrs['user'] = None
 
+        # ensure feedback is not submitted with only white spaces
+        if not attrs['feedback'].strip():
```
diox: Can you make a `validate_feedback()` method and move that code in there ? That way, the error will be associated to `feedback` instead of going in `non_field_errors`. You'll have to update the test.

Nice work otherwise!

diox:
Great, thanks.

PR #3274, opened 2015-08-20
Make mkt_gc cron() work with a remote storage (bug 1195271)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1195271

### DIFF
--- a/mkt/webapps/cron.py
+++ b/mkt/webapps/cron.py
@@ -17,8 +17,8 @@ from mkt.api.models import Nonce
 from mkt.developers.models import ActivityLog
 from mkt.files.models import File, FileUpload
 from mkt.site.decorators import use_master
-from mkt.site.storage_utils import (private_storage, storage_is_remote,
-                                    walk_storage)
+from mkt.site.storage_utils import (private_storage, public_storage,
+                                    storage_is_remote, walk_storage)
 from mkt.site.utils import chunked, days_ago, walkfiles
 
 from .indexers import WebappIndexer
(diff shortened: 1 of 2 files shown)


washort:
r+
what a mess, nice job coping with that

mkt/webapps/cron.py
```diff
@@ -564,22 +569,26 @@ def mkt_gc(**kw):
     Nonce.objects.filter(created__lt=days_ago(1)).delete()
 
     # Delete the dump apps over 30 days.
-    _remove_stale_files(settings.DUMPED_APPS_PATH,
+    _remove_stale_files(os.path.join(settings.DUMPED_APPS_PATH, 'tarballs'),
```
diox: Note for posterity: it was actually broken before, and we could retrieve all older dumps, because `tarballs` was not included, so `_remove_stale_files()` was listing the parent dir, only finding the `tarballs` directory, which always had a recent modification date since there are new dumps put in there every day.

mkt/webapps/cron.py
```diff
@@ -564,22 +569,26 @@ def mkt_gc(**kw):
...
                         settings.DUMPED_APPS_DAYS_DELETE,
-                        'Deleting old tarball: {0}')
+                        'Deleting old tarball: {0}',
+                        storage=public_storage)
 
     # Delete the dumped user installs over 30 days.
-    _remove_stale_files(settings.DUMPED_USERS_PATH,
+    _remove_stale_files(os.path.join(settings.DUMPED_USERS_PATH, 'tarballs'),
```
diox: Same as above

PR #3275, opened 2015-08-20
User dumps should be using private_storage, not public (bug 1195267)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1195267


robhudson:
r+

PR #3278, opened 2015-08-20
remove further uses of default_storage (bug 1167271)

washort:
(sto)rage against the machine

### DIFF
--- a/mkt/files/models.py
+++ b/mkt/files/models.py
@@ -5,7 +5,6 @@ import unicodedata
 import uuid
 
 from django.conf import settings
-from django.core.files.storage import default_storage as storage
 from django.core.urlresolvers import reverse
 from django.db import models
 from django.dispatch import receiver
@@ -17,12 +16,12 @@ from jingo.helpers import urlparams
 from uuidfield.fields import UUIDField
 
 import mkt
-from mkt.site.storage_utils import copy_stored_file, move_stored_file
 from mkt.site.decorators import use_master
 from mkt.site.helpers import absolutify
 from mkt.site.models import ModelBase, OnChangeMixin
+from mkt.site.storage_utils import (copy_stored_file, move_stored_file,
+                                    private_storage, public_storage)
 from mkt.site.utils import smart_path
-from mkt.site.storage_utils import private_storage, public_storage
 
 
 log = commonware.log.getLogger('z.files')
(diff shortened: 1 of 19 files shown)


mkt/files/models.py
```diff
@@ -231,7 +230,7 @@ def cleanup_file(sender, instance, **kw):
             return
         # FIXME: This needs to check if the file is public or not and choose
         # the appropriate storage backend.
-        if filename and storage.exists(filename):
+        if filename and public_storage.exists(filename):
```
robhudson: I think this should just check both public and private (and remove the FIXME comment)

washort: good point, fixed

mkt/files/models.py
```diff
@@ -229,12 +228,12 @@ def cleanup_file(sender, instance, **kw):
...
-        # FIXME: This needs to check if the file is public or not and choose
-        # the appropriate storage backend.
-        if filename and storage.exists(filename):
+        if filename and (public_storage.exists(filename) or
+                         private_storage.exists(filename)):
             log.info('Removing filename: %s for file: %s'
                      % (filename, instance.pk))
-            storage.delete(filename)
+            public_storage.delete(filename)
+            private_storage.delete(filename)
```
robhudson: Will this throw an error if it exists on one but not the other?

mkt/site/utils.py
```diff
@@ -265,11 +264,11 @@ def __iter__(self):
             return iter([])
 
         chunk = 4096
-        fp = storage.open(self.path, 'rb')
+        fp = private_storage.open(self.path, 'rb')
         if 'wsgi.file_wrapper' in self.request.META:
             return self.request.META['wsgi.file_wrapper'](fp, chunk)
         else:
-            self['Content-Length'] = storage.size(self.path)
+            self['Content-Length'] = private_storage.size(self.path)
```
robhudson: I wonder if this should be `local_storage` here after my patch. `HttpResponseSendFile` should never be called if we are using S3.

mkt/submit/tests/test_views.py
```diff
@@ -16,7 +15,7 @@
 from mkt.files.tests.test_models import UploadTest as BaseUploadTest
 from mkt.reviewers.models import EscalationQueue
 from mkt.site.fixtures import fixture
-from mkt.site.storage_utils import copy_to_storage
+from mkt.site.storage_utils import public_storage, copy_to_storage
```
robhudson: c < p

robhudson:
r+wc

mkt/files/models.py
washort: Docs say delete does not throw exceptions when the path doesn't exist.

mkt/site/utils.py
washort: OK.

mkt/files/models.py
robhudson: Ok cool

PR #3281, opened 2015-08-21
Add API listing FxOS add-ons you've developped (bug 1196224)

diox:
Bonus: basic detail API as well.

https://bugzilla.mozilla.org/show_bug.cgi?id=1196224


docs/api/topics/app_validation.rst
```diff
@@ -1,112 +0,0 @@
-.. _app_validation:
```
diox: I swear I'm done moving this file around. I initially wanted to share it between extensions and apps but decided against it to I'm moving it back to the `addons_submission` topic.

mkt/extensions/views.py
```diff
@@ -85,6 +86,15 @@ class ExtensionViewSet(CORSMixin, MarketplaceView,
     queryset = Extension.objects.all()
     serializer_class = ExtensionSerializer
 
+    def filter_queryset(self, qs):
+        if self.action == 'list':
+            # The listing API only allows you to see extensions you've
+            # developed.
```
chuckharmston: Why is that so?

mkt/extensions/views.py
```diff
@@ -85,6 +86,15 @@ class ExtensionViewSet(CORSMixin, MarketplaceView,
...
+            if not self.request.user.is_authenticated():
```
chuckharmston: Is there any reason for this? I presume we don't allow anonymous submission, and we don't allow anonymous listing, so we could just remove `RestAnonymousAuthentication`.

mkt/extensions/tests/test_views.py
```diff
@@ -179,3 +179,63 @@ def test_not_an_addon(self):
...
+                           kwargs={'pk': self.extension.pk})
+        self.url2 = reverse('api-v2:extension-detail',
+                            kwargs={'pk': self.extension2.pk})
+
+    def test_list_anonymous(self):
+        response = self.anon.get(self.list_url)
+        eq_(response.status_code, 403)
+
+    def test_list_logged_in(self):
+        response = self.client.get(self.list_url)
```
chuckharmston: I like this pattern to keep boilerplate out of tests:

    def _get(self, client, url=None, **params):
        if not url:
            url = self.list_url
        response = client.get(url, params)
        json.loads(res.content)
        return response, data

mkt/extensions/tests/test_views.py
```diff
@@ -179,3 +179,63 @@ def test_not_an_addon(self):
...
+        response = self.anon.get(self.url)
+        eq_(response.status_code, 403)
+
+        self.extension.update(status=STATUS_PUBLIC)
+        response = self.anon.get(self.url)
+        eq_(response.status_code, 200)
+        data = response.json
+        eq_(data['id'], self.extension.id)
+        eq_(data['status'], 'public')
+        eq_(data['name'], {"en-US": self.extension.name})
```
chuckharmston: Mix of `"` and `'` here.

docs/api/topics/addons.rst
```diff
@@ -0,0 +1,60 @@
...
+    **Response**
+
+    :param meta: :ref:`meta-response-label`.
+    :type meta: object
+    :param objects: A :ref:`listing <objects-response-label>` of :ref:`addons <addon-response-label>`.
+    :type objects: array
+
+.. _addon-response-label:
+
+.. http:get:: /api/v2/extensions/extension/(int:id)|(string:slug)/
```
chuckharmston: We use addons in some places, but extensions in others. Can we fix that?

chuckharmston: We should also make sure it's clear in wording that we're dealing with fxos addons and not desktop.

mkt/extensions/views.py
diox: The detail API is also in there, and also available to anonymous users

mkt/extensions/views.py
diox: The public API will be search, like for apps (which have the same restriction about listing)

docs/api/topics/addons.rst
diox: The public name should be "Add-on", but the code and API uses "extensions" because that's the technical name (we're following Chrome "extensions" spec, the platform calls it "extensions", etc) and because #3239 is not fixed.

diox: I'll add a bunch of `Firefox OS` in the docs.

chuckharmston: Can we do it in the code, too? There's a pretty plausible future where Marketplace hosts Firefox addons as well. Would be nice to head it off and avoid something as sweeping as #3239 for this.

diox: If we end up hosting Firefox Desktop Addons, there will be nothing left to do since they will also be extensions following the same Chrome spec (in fact, one of the big advantages of that format is having the same code for both mobile and desktop)

mkt/extensions/tests/test_views.py
diox: Since nowadays all our responses in tests are passed through `mkt.site.tests._JSONifiedResponse`, which does the json loading, I feel there is little boilerplate here.

Compare:

```
response = self.client.get(self.list_url)
# response.json
```

```
response, response_json = response.get()
```

The former also has the json version if needed, which is only returned if accessed, and is more explicit - it's easier to follow the test.

diox:
Merging to unblock me.

PR #3287, opened 2015-08-24
Add view listing all public FxOS Add-ons (bug 1196225)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1196225

### DIFF
--- a/mkt/extensions/models.py
+++ b/mkt/extensions/models.py
@@ -3,13 +3,14 @@ import os.path
 
 from django.conf import settings
 from django.db import models
-from django.utils.functional import lazy
+from django.dispatch import receiver
 
 import commonware.log
 from django_extensions.db.fields.json import JSONField
 
 from mkt.constants.base import (STATUS_CHOICES, STATUS_NULL, STATUS_PENDING,
                                 STATUS_PUBLIC)
+from mkt.extensions.indexers import ExtensionIndexer
 from mkt.extensions.utils import ExtensionParser
 from mkt.files.models import cleanup_file, nfd_str
 from mkt.translations.fields import save_signal, TranslatedField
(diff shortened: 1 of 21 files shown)


mkt/extensions/models.py
```diff
@@ -23,14 +24,6 @@
 log = commonware.log.getLogger('z.extensions')
 
 
-def _make_language_choices(languages):
-    return [(to_language(lang_code), lang_name)
-            for lang_code, lang_name in languages.items()]
-
-
-LANGUAGE_CHOICES = lazy(_make_language_choices, list)(settings.LANGUAGES)
```
diox: Not needed because I relaxed the `default_language` field, we don't actually need to limit choices in there.

mkt/extensions/models.py
```diff
@@ -149,6 +146,20 @@ def __unicode__(self):
...
+          dispatch_uid='extension_index')
+def update_search_index(sender, instance, **kw):
+    instance.get_indexer().index_ids([instance.id])
+
+
+# Delete from ElasticSearch index on delete.
+@receiver(models.signals.post_delete, sender=Extension,
+          dispatch_uid='extension_unindex')
+def delete_search_index(sender, instance, **kw):
+    instance.get_indexer().unindex(instance.id)
```
robhudson: Are we removing from the index or just filtering them out? For apps we filter so things like the lookup tool can still find them.

robhudson:
very nice. r+wc.

mkt/extensions/models.py
diox: For the initial walking skeleton prototype we don't have soft-deletion, so we're directly deleting. Well, for now deletion is not even exposed through the API / UI :)

PR #3292, opened 2015-08-24
Fix more s3

washort:


mkt/developers/tasks.py
```diff
@@ -154,7 +154,8 @@ def _hash_file(fd):
 
 @post_request_task
 @set_modified_on
-def resize_icon(src, dst, sizes, storage=private_storage, **kw):
+def resize_icon(src, dst, sizes, src_storage=private_storage,
+                dest_storage=public_storage, **kw):
```
robhudson: we should unify where we use `dst` and `dest`. I have to look every time.

mkt/developers/tasks.py
```diff
@@ -162,11 +163,11 @@ def resize_icon(src, dst, sizes, storage=private_storage, **kw):
         for s in sizes:
             size_dst = '%s-%s.png' % (dst, s)
             resize_image(src, size_dst, (s, s), remove_src=False,
-                         src_storage=storage)
-            pngcrush_image.delay(size_dst, storage=storage, **kw)
-        with storage.open(src) as fd:
+                         src_storage=src_storage, dst_storage=dest_storage)
```
robhudson: Like this.

robhudson:
r+wc

PR #3300, opened 2015-08-25
Fix remaining S3 tests

washort:


mkt/site/tests/test_storage_utils.py
```diff
@@ -115,12 +115,3 @@ def test_get_storage_local(self):
...
         eq_(get_public_storage().__class__.__name__, 'LocalFileStorage')
-
-    @override_settings(
-        DEFAULT_FILE_STORAGE='mkt.site.storage_utils.LocalFileStorage')
-    def test_simple_lazy_object(self):
-        from mkt.site.storage_utils import private_storage, public_storage
-
-        assert not storage_is_remote()
-        eq_(private_storage.__class__.__name__, 'LocalFileStorage')
-        eq_(public_storage.__class__.__name__, 'LocalFileStorage')
```
washort: This cannot be handled reasonably in a unit test; the settings override doesn't help because the lazy object resolves to its target on its first use, which can be in tests run previous to this one.

PR #3301, opened 2015-08-26
Implement signing in FxOS Add-ons (bug 1198696)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1198696

### DIFF
--- a/mkt/extensions/migrations/0002_extension_add_uuid.py
+++ b/mkt/extensions/migrations/0002_extension_add_uuid.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+import uuidfield.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('extensions', '0001_initial'),
+    ]
+
+    operations = [
+        # Add the field, without a unique constraint for now. This is splitted
+        # into 3 migrations, in order for it to work no matter which state the
+        # Extension model is in.
+        # See https://docs.djangoproject.com/en/dev/howto/writing-migrations/#migrations-that-add-unique-fields
+        migrations.AddField(
+            model_name='extension',
+            name='uuid',
+            field=uuidfield.fields.UUIDField(null=True, max_length=32,
+                                             editable=False, blank=True),
+        ),
+    ]
--- a/mkt/extensions/migrations/0003_populate_extension_uuid.py
+++ b/mkt/extensions/migrations/0003_populate_extension_uuid.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def add_uuid_to_exising_extensions(apps, schema_editor):
+    Extension = apps.get_model('extensions', 'Extension')
+    extensions = Extension.objects.all()
+    for extension in extensions:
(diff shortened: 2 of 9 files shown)


mkt/extensions/migrations/0002_extension_add_uuid.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
chuckharmston: These are great work.

diox: Can't take the credit for it, the idea comes from the django docs. I was about to write a single python migration, the same kind that would sometimes break back when we were using schematic. Was pretty happy to read how to do it properly in the docs :)

chuckharmston:
Not super comfortable with the signing plans for addons, so I'll defer the r+ to somebody who knows expectations. Looks great though: endpoints, migrations, tests. Well done.

washort:
r+

mkt/extensions/migrations/0003_populate_extension_uuid.py
```diff
@@ -0,0 +1,26 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def add_uuid_to_exising_extensions(apps, schema_editor):
+    Extension = apps.get_model('extensions', 'Extension')
+    extensions = Extension.objects.all()
+    for extension in extensions:
+        extension.uuid = extension._meta.get_field('uuid')._create_uuid()
```
robhudson: Super secret internal fields and methods. :scream:

mkt/extensions/migrations/0003_populate_extension_uuid.py
```diff
@@ -0,0 +1,26 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('extensions', '0002_extension_add_uuid'),
+    ]
+
+    operations = [
+        # Update the uuid field on each row.
+        migrations.RunPython(
+            add_uuid_to_exising_extensions,
```
robhudson: This is just an arg to the `RunPython` call right? It seems odd to include the trailing comma to me.

mkt/extensions/models.py
```diff
@@ -140,18 +155,52 @@ def is_public(self):
...
+            # 'version' should be an integer and should be monotonically
+            # increasing.
+            'version': self.file_version
+        })
+        with statsd.timer('extensions.sign'):
+            try:
+                # This will read the file from self.file_path, generate a
+                # signature and write the signed file to self.signed_file_path.
+                sign_app(private_storage.open(self.file_path),
+                         self.signed_file_path, ids)
```
robhudson: `sign_app` will work for both packaged apps and extensions?

robhudson:
nice work

mkt/extensions/migrations/0003_populate_extension_uuid.py
diox: Yeah. It's not that secret, really all it does is a `uuid.uuid4()` call. I could replace it - I wanted to use what the field really uses behind the scenes

mkt/extensions/models.py
diox: Yeah, it doesn't care about the manifest, and they are both zips, signed the same way. It *should* work on the device, but I really have no idea at this point : I need to have the manifest and signing working to test it.

PR #3303, opened 2015-08-26
Remove copy_to_storage, harmonize variable names (bug 1195744)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1195744

### DIFF
--- a/lib/crypto/packaged.py
+++ b/lib/crypto/packaged.py
@@ -13,8 +13,8 @@ from django_statsd.clients import statsd
 from signing_clients.apps import JarExtractor
 
 from mkt.versions.models import Version
-from mkt.site.storage_utils import (local_storage, private_storage,
-                                    public_storage)
+from mkt.site.storage_utils import (copy_stored_file, local_storage,
+                                    private_storage, public_storage)
 
 
 log = commonware.log.getLogger('z.crypto')
@@ -102,9 +102,9 @@ def _sign_app(src, dest, ids, reviewer, tempname, local=False):
     elif local:
         storage = local_storage
 
-    with local_storage.open(tempname) as temp_f, \
-            storage.open(dest, 'w') as dst_f:
-        shutil.copyfileobj(temp_f, dst_f)
+    copy_stored_file(
+        tempname, dest,
+        src_storage=local_storage, dst_storage=storage)
 
 
 def _get_endpoint(reviewer=False):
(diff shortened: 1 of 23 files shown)


lib/crypto/packaged.py
```diff
@@ -103,8 +103,8 @@ def _sign_app(src, dest, ids, reviewer, tempname, local=False):
         storage = local_storage
 
     with local_storage.open(tempname) as temp_f, \
-            storage.open(dest, 'w') as dest_f:
-        shutil.copyfileobj(temp_f, dest_f)
+            storage.open(dest, 'w') as dst_f:
+        shutil.copyfileobj(temp_f, dst_f)
```
washort: could use `copy_stored_file` here.

lib/crypto/tests.py
```diff
@@ -30,8 +31,8 @@ def mock_sign(version_id, reviewer=False):
     file_obj = version.all_files[0]
     path = (file_obj.signed_reviewer_file_path if reviewer else
             file_obj.signed_file_path)
-    with private_storage.open(path, 'w') as dest_f:
-        shutil.copyfileobj(private_storage.open(file_obj.file_path), dest_f)
+    with private_storage.open(path, 'w') as dst_f:
+        shutil.copyfileobj(private_storage.open(file_obj.file_path), dst_f)
```
washort: here too

washort:
i liked `dest` better but consistency is good :) r+wc

lib/crypto/packaged.py
diox: Good call, I wanted to do a pass on `copyfileobj` usage but forgot, I'll take a look.

robhudson:
looks good to me

lib/crypto/packaged.py
```diff
@@ -126,12 +126,12 @@ def _get_endpoint(reviewer=False):
...
 
-def _no_sign(src, dest_path):
+def _no_sign(src, dst_path):
     # If this is a local development instance, just copy the file around
     # so that everything seems to work locally.
     log.info('Not signing the app, no signing server is active.')
-    with public_storage.open(dest_path, 'w') as dest_f:
-        shutil.copyfileobj(src, dest_f)
+    with public_storage.open(dst_path, 'w') as dst_f:
+        shutil.copyfileobj(src, dst_f)
```
diox: Note to self: this is one of the places where we can *not* use `copy_stored_file` instead of `copyfileobj`, because we don't know the source storage.

PR #3305, opened 2015-08-27
Fix calls to copy_stored_file() without the right source storage (bug 1195744)

diox:


mkt/site/storage_utils.py
```diff
@@ -120,7 +120,7 @@ def storage_is_remote():
 
 
 def walk_storage(path, topdown=True, onerror=None, followlinks=False,
-                 storage=private_storage):
+                 storage=None):
```
diox: Let's see how many tests fail when I force all code to explicitly pass storage classes.

mkt/webapps/tasks.py
```diff
@@ -462,6 +462,7 @@ def compress_export(tarball_name, date):
     remote_target_filename = os.path.join(
         settings.DUMPED_APPS_PATH, 'tarballs', '%s.tgz' % tarball_name)
     copy_stored_file(local_target_file.name, remote_target_filename,
+                     src_storage=local_storage,
```
diox: This was the main issue I was tracking (and the one below too).

diox:
Turns out we were already doing the right thing everywhere except in `mkt/webapps/tasks.py` but I think it's best to be explicit here and always pass the `storage` to use to avoid potential issues.

robhudson:
r+. If we're always calling it explicitly would it be better to remove the defaults on the args?

diox:
I'm not sure if we should remove the defaults just yet. I did it temporarily to find bugs, but I'm not sure it's the best thing to do right now. It's still early IMHO, we've been messing with that for the past few days, so I want to let it live a little and see how we use it to make this kind of decisions.

PR #3306, opened 2015-08-27
Add manifest and download views for FxOS Add-ons (bug 1199186, bug 1199189)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1199186
https://bugzilla.mozilla.org/show_bug.cgi?id=1199189

### DIFF
--- a/mkt/extensions/indexers.py
+++ b/mkt/extensions/indexers.py
@@ -57,6 +57,7 @@ class ExtensionIndexer(BaseIndexer):
                     },
                     # Name for sorting.
                     'name_sort': cls.string_not_analyzed(doc_values=True),
+                    'guid': cls.string_not_analyzed(),
                     'slug': {'type': 'string'},
                     'status': {'type': 'byte'},
                     'version': cls.string_not_indexed(),
@@ -91,6 +92,7 @@ class ExtensionIndexer(BaseIndexer):
         # set to True at the moment, and not present in the model.
         doc['is_disabled'] = False
         doc['name_sort'] = unicode(obj.name).lower()
+        doc['guid'] = unicode(obj.uuid)
 
         # Handle localized fields. This adds both the field used for search and
         # the one with all translations for the API.
--- a/mkt/extensions/models.py
+++ b/mkt/extensions/models.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
 import json
 import os.path
+from copy import deepcopy
 
 from django.conf import settings
+from django.core.urlresolvers import reverse
 from django.db import models
 from django.dispatch import receiver
 
@@ -19,6 +21,7 @@ from mkt.extensions.utils import ExtensionParser
 from mkt.files.models import cleanup_file, nfd_str
 from mkt.translations.fields import save_signal, TranslatedField
 from mkt.translations.utils import to_language
+from mkt.site.helpers import absolutify
 from mkt.site.models import ManagerBase, ModelBase
 from mkt.site.storage_utils import (copy_stored_file, private_storage,
                                     public_storage)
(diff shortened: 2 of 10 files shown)


mkt/extensions/indexers.py
```diff
@@ -57,6 +57,7 @@ def get_mapping(cls):
                     },
                     # Name for sorting.
                     'name_sort': cls.string_not_analyzed(doc_values=True),
+                    'guid': cls.string_not_analyzed(),
```
ngokevin: curious why not ```uuid```

mkt/urls.py
```diff
@@ -38,6 +39,8 @@
         name='detail.manifest'),
     url('^langpack/%s/manifest.webapp$' % mkt.ADDON_UUID,
         mini_langpack_manifest, name='langpack.manifest'),
+    url('^extension/(?P<uuid>[0-9a-f]{32})/manifest.json$',
```
ngokevin: curious why it's in this urls file

mkt/extensions/indexers.py
diox: To be compatible with the way we store it for apps in ES. Not necessary right now but will help in the future (guid lookups are used in the lookup tool), so I figured why not.

mkt/urls.py
diox: no particular reason, didn't feel like it was worth including patterns here just for one url. And we do need to include it from the root, the only other inclusion we have right now is prefixed with `/api/v2/`

ngokevin:
need docs?

ngokevin:
lgtm, i want to try it out before you head out. though @robhudson's eyes would be good

diox:
Oh yeah I need to update the docs to mention `manifest_url` and `download_url`. I'll do that

mkt/extensions/models.py
```diff
@@ -153,7 +156,19 @@ def is_public(self):
...
-        raise NotImplementedError
+        return absolutify(reverse('extension.manifest',
+                                  kwargs={'uuid': self.uuid}))
+
+    @property
+    def mini_manifest(self):
+        """Mini-manifest used for install/update in dict form.
+
+        Actually a maxi-manifest for the moment. Must contain `package_path`,
+        that points to the extension download absolute URL."""
```
diox: So, just to leave a trace somewhere: this will actually not be enough for installing to work. FxOS wants mini-manifests to be like apps, in a different format than the one we use for extensions. This means we'll have to create it from scratch using the data on the instance/actual manifest. I'll do that tomorrow.

diox:
Added documentation about the new properties. I decided not to document the views themselves since they are technically not API, but let me know if that'd be useful to do and I'll reverse my stance if it is :)

ngokevin:
Sounds good

mkt/extensions/views.py
```diff
@@ -185,3 +188,32 @@ def reject(self, request, *args, **kwargs):
...
+
+    if extension.is_public():
+        log.info('Downloading add-on: %s from %s' % (
+                 extension.pk, extension.signed_file_path))
+        extension_etag = hashlib.sha256()
+        extension_etag.update(unicode(extension.uuid))
+        extension_etag.update(unicode(extension.file_version))
+        return get_file_response(request, extension.signed_file_path,
+                                 content_type='application/zip',
+                                 etag=extension_etag.hexdigest(), public=True)
```
robhudson: I pretty sure `content_type` or `etag` do nothing under S3 -- they have to be set on upload. We may have to do a pass looking for things like that when the files are signed.

diox: I *think* we already set it at some point, but I may be wrong about this. We'll bump into this issue for webapps when we switch to S3 so I'm not worried about it.

ngokevin:
the download URL is 404ing for me, maybe you have to work with ops on that.

PR #3308, opened 2015-08-28
Make mini-manifest a Webapp manifest, not an extension manifest (bug 1199189)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1199189

We need the manifest.json inside the zip to follow the Web Extensions format, and the mini-manifest to follow Mozilla's "App Manifest" format to make add-ons to be installable and updatable on FxOS.

See https://etherpad.mozilla.org/converting-fxos-addons and 'PSA: add-ons changes are coming' thread on mozilla.dev.b2g.

### DIFF
--- a/docs/api/topics/firefox_os_addons.rst
+++ b/docs/api/topics/firefox_os_addons.rst
@@ -58,8 +58,8 @@ Add-on
     :type download_url: string
     :param name: The add-on name.
     :type name: string|object
-    :param manifest_url: The (absolute) URL to the mini-manifest for that add-on. That URL may be a 404 if the add-on is not public.
-    :type manifest_url: string
+    :param mini_manifest_url: The (absolute) URL to the `mini-manifest <https://developer.mozilla.org/docs/Mozilla/Marketplace/Options/Packaged_apps#Publishing_on_Firefox_Marketplace>`_ for that add-on. That URL may be a 404 if the add-on is not public.
+    :type mini_manifest_url: string
     :param slug: The add-on slug (unique string identifier that can be used
         instead of the id to retrieve an add-on).
     :type slug: string
--- a/mkt/extensions/models.py
+++ b/mkt/extensions/models.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 import json
 import os.path
-from copy import deepcopy
 
 from django.conf import settings
 from django.core.urlresolvers import reverse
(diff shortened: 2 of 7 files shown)


docs/api/topics/firefox_os_addons.rst
```diff
@@ -58,8 +58,8 @@ Add-on
     :type download_url: string
     :param name: The add-on name.
     :type name: string|object
-    :param manifest_url: The (absolute) URL to the mini-manifest for that add-on.
-    :type manifest_url: string
+    :param mini_manifest_url: The (absolute) URL to the mini-manifest for that add-on.
+    :type mini_manifest_url: string
```
diox: @ngokevin , @chuckharmston : how do you feel about this change ? I think it's less confusing that way. Since the 2 formats are completely different I figured it was probably best to make sure we don't confuse them.

chuckharmston: I like the idea, but I think it probably needs to be paired with a good explanation about what a mini-manifest is, where it comes from, etc.

chuckharmston: Our team is great at coming up with cutesy names that make sense to us, but this is outward-facing documentation to people who might be unfamiliar.

diox: **mini-manifest** is a term used on MDN already though: https://developer.mozilla.org/en-US/Marketplace/Options/Packaged_apps#Publishing_on_Firefox_Marketplace

chuckharmston: Perfect! Something easy to link to.

diox: But I'd be happy to have an explanation, I'm just not sure where it'd be the most appropriate and how to word it. I kinda prefer relying on MDN for details like this

chuckharmston: Stab at it :hocho::

> The mini-manifest is a Marketplace-generated subset of your add-on's manifest that is used for installations and updates. Learn more [at MDN]().

mkt/extensions/models.py
```diff
@@ -160,19 +159,33 @@ def is_public(self):
...
     @property
     def mini_manifest(self):
-        """Mini-manifest used for install/update in dict form.
-
-        Actually a maxi-manifest for the moment. Must contain `package_path`,
-        that points to the extension download absolute URL."""
-
-        mini_manifest = deepcopy(self.manifest)
-        mini_manifest['package_path'] = self.download_url
+        """Mini-manifest used for install/update on FxOS devices, in dict form.
```
chuckharmston: \n

mkt/extensions/models.py
```diff
@@ -160,19 +159,33 @@ def is_public(self):
...
+
+        It follows the Mozilla App Manifest format, *not* the Web Extension
+        manifest format.
```
chuckharmston: I wouldn't mind it if you expounded on this: how it's translated and why this has to be the case. Spending some time thinking on it, I understand it, but it's not very obvious.

chuckharmston:
r+, looks great.

mkt/extensions/models.py
diox: Regarding the why:

Something like ```It follows the Mozilla App Manifest format (because that's what Firefox OS requires to install/update add-ons), *not* the Web Extension manifest format.``` ?

It was not obvious to me at all either. It happens because internally, add-ons are actually implemented as Webapps, they are listed in `mozApps.getInstalled()` etc. I'm not sure how much we want to elaborate on that in the docstring.

How:

It's described below in the comment. I felt it was too internal for it to be part of the docstring.

chuckharmston: > It follows the Mozilla App Manifest format (because that's what Firefox OS requires to install/update add-ons), *not* the Web Extension manifest format.

Perfect. As long as a casual reader can negotiate the pieces as they fit together.

PR #3309, opened 2015-08-28
 enable hindi; bug 1154929 

clouserw:


eviljeff:
needs a migration generating as the langpack model uses this constant for its choices.

washort:
merged,  911dd4a64d

washort:
oops i mean 9fa48502d

PR #3310, opened 2015-08-28
Add an unsigned download URL for authors & reviewers (bug 1199198)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1199198

Note that I'm only providing an unsigned download link for reviewers (and authors), I didn't go all the way and give them a special signed download + special manifest, so they won't be able to directly install non-public add-ons on their devices, but they will be able to download them, and that's enough for the walking skeleton.

### DIFF
--- a/docs/api/topics/firefox_os_addons.rst
+++ b/docs/api/topics/firefox_os_addons.rst
@@ -54,11 +54,11 @@ Add-on
 
     **Response**
 
-    :param download_url: The (absolute) URL to the latest signed package for that add-on.
+    :param download_url: The (absolute) URL to the latest signed package for that add-on. That URL may be a 404 if the add-on is not public.
     :type download_url: string
     :param name: The add-on name.
     :type name: string|object
-    :param manifest_url: The (absolute) URL to the mini-manifest for that add-on.
+    :param manifest_url: The (absolute) URL to the mini-manifest for that add-on. That URL may be a 404 if the add-on is not public.
     :type manifest_url: string
     :param slug: The add-on slug (unique string identifier that can be used
         instead of the id to retrieve an add-on).
@@ -66,6 +66,8 @@ Add-on
     :param status: The add-on current status.
         Can be "incomplete", "pending", "public" or "rejected".
     :type status: string
+    :param unsigned_download_url: The (absolute) URL to the latest *unsigned* package for that add-on. Only the add-on author or users with Extensions:Review permission may access it.
+    :type unsigned_download_url: string
     :param version: The add-on current version number.
     :type version: string
 
--- a/mkt/extensions/models.py
+++ b/mkt/extensions/models.py
@@ -36,6 +36,9 @@ class ExtensionManager(ManagerBase):
     def pending(self):
         return self.filter(status=STATUS_PENDING).order_by('id')
 
+    def public(self):
+        return self.filter(status=STATUS_PUBLIC).order_by('id')
+
 
 class Extension(ModelBase):
     uuid = UUIDField(auto=True)
@@ -62,7 +65,7 @@ class Extension(ModelBase):
     @property
     def download_url(self):
         return absolutify(reverse(
-            'extension.download',
+            'extension.download_signed',
             kwargs={'uuid': self.uuid, 'filename': self.filename}))
 
     @property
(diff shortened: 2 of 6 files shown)


docs/api/topics/firefox_os_addons.rst
```diff
@@ -66,6 +66,8 @@ Add-on
     :param status: The add-on current status.
         Can be "incomplete", "pending", "public" or "rejected".
     :type status: string
+    :param unsigned_download_url: The (absolute) URL to the latest *unsigned* package for that add-on. It may not exist if the add-on is not public. Only the add-on author or users with Extensions:Review permission may access it.
+    :type unsigned_download_url: string
```
chuckharmston: `string|null`?

diox: It's always there, it just might not be accessible depending on your rights.

mkt/extensions/models.py
```diff
@@ -62,8 +65,8 @@ def clean_slug(self):
     @property
     def download_url(self):
         return absolutify(reverse(
-            'extension.download',
-            kwargs={'uuid': self.uuid, 'filename': self.filename}))
+            'extension.download_signed',
+            kwargs={'uuid': self.uuid, 'name': self.filename}))
```
chuckharmston: Maybe we should use `filename`, since there is a `name` field on the add-on itself.

docs/api/topics/firefox_os_addons.rst
chuckharmston: Ah. You should clarify that, the "may not exist" language implies otherwise.

mkt/extensions/tests/test_views.py
```diff
@@ -141,10 +141,10 @@ def test_create_logged_in(self):
                                     json.dumps({'upload': upload.pk}))
         eq_(response.status_code, 201)
         data = response.json
-        eq_(data['version'], '0.1')
         eq_(data['name'], {"en-US": u"My Lîttle Extension"})
-        eq_(data['status'], 'pending')
         eq_(data['slug'], u'my-lîttle-extension')
+        eq_(data['status'], 'pending')
+        eq_(data['version'], '0.1')
```
chuckharmston: :sun_with_face:

mkt/extensions/tests/test_views.py
```diff
@@ -513,22 +525,47 @@ def test_download(self):
...
 
-        self.login(self.user)  # Even logged in you can't access it for now.
+        self.login(self.user)
+        self.grant_permission(self.user, 'Extensions:Review')
         response = self.client.get(self.extension.download_url)
+        # Even authors and reviewers can't access it: it doesn't exist.
         eq_(response.status_code, 404)
 
+    @override_settings(
+        XSENDFILE=True,
```
chuckharmston: Are we still using xsendfile with s3?

mkt/extensions/views.py
```diff
@@ -191,20 +195,38 @@ def reject(self, request, *args, **kwargs):
...
+                             content_type='application/zip',
+                             etag=extension_etag.hexdigest(), public=True)
+
+
+def download_signed(request, uuid, **kwargs):
+    extension = get_object_or_404(Extension.objects.public(), uuid=uuid)
+
+    log.info('Downloading add-on: %s from %s' % (
+             extension.pk, extension.signed_file_path))
+    return _download(request, extension, extension.signed_file_path)
```
chuckharmston: :smiling_imp:

chuckharmston:
Looks fantastic, r+ if tests pass.

mkt/extensions/tests/test_views.py
diox: No, that's why I'm forcing the storage. On the other views I've added a second test with the right storage class to test the S3 variant, but here I haven't. I'll do that.

docs/api/topics/firefox_os_addons.rst
diox: Yeah, I messed it up. Updated the docs, let me know if that works for you now.

mkt/extensions/tests/test_views.py
diox: Added more tests w/ S3 storage

docs/api/topics/firefox_os_addons.rst
chuckharmston: ++

mkt/extensions/models.py
diox: Fixed

chuckharmston:
tumblr_mlqgk23xqb1rjm9fto1_400

diox:
Waiting for travis to do its thing :passport_control:

PR #3312, opened 2015-08-31
Make S3BotoPublicStorage.url() use the CDN (bug 1198514)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1198514

This also works around https://bugzilla.mozilla.org/show_bug.cgi?id=1198967 since we don't bother generating the URL on S3.

Need some configuration on the CDN settings to work (the CDN needs to recognize that `https://cdn.host/_s3/<path>` origin should be `https://<s3_public_bucket>/<path>`)

### DIFF
--- a/mkt/site/storage_utils.py
+++ b/mkt/site/storage_utils.py
@@ -86,6 +86,21 @@ class S3BotoPublicStorage(S3BotoStorage):
         'Cache-Control': 'max-age=31536000',
     }
 
+    def url(self, name, **kwargs):
+        if settings.STATIC_URL.startswith(('/', settings.SITE_URL)):
+            # If STATIC_URL starts with / or SITE_URL, then we're serving
+            # assets locally from a local development environnement. Just
+            # return the S3 URL directly.
+            return super(S3BotoPublicStorage, self).url(name, **kwargs)
+        else:
+            # In "real" environnements (prod/stage/dev), we want to use the CDN
+            # for public resources instead of directly calling S3. To do that,
+            # we manually construct the URL to point to STATIC_URL (which is
+            # our CDN) and we add a "_s3" component to the URL to help the CDN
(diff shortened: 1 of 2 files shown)


mkt/site/storage_utils.py
```diff
@@ -86,6 +86,14 @@ class S3BotoPublicStorage(S3BotoStorage):
         'Cache-Control': 'max-age=31536000',
     }
 
+    def url(self, name, headers=None, response_headers=None):
+        # We want to use the CDN for public ressources instead of directly
+        # calling S3. To do that, we are manually constructing the URL to point
+        # to STATIC_URL (which is our CDN) and we add a "_s3" component to the
+        # URL to help the CDN figure out that the underlying ressource lives on
+        # S3. The bucket is hardcoded in the CDN configuration.
+        return '%s_s3/%s' % (settings.STATIC_URL, name.lstrip('/'))
```
diox: Mmm, this might break on local envs not using a CDN for `STATIC_URL`. Maybe I should do this in a separate class ? Or depend on another setting ?

washort: you could just check `if STATIC_URL.startswith(SITE_URL):` and use the regular S3 url in that case. (Or add an `_s3` route that redirects to AWS.)

diox: I'd prefer to avoid redirects, that would complicate things. I like the `STATIC_URL.startswith(SITE_URL)` idea, I'd also need to check if STATIC_URL starts with `/` but I think it should work.

washort:
r+

diox:
Keeping this unmerged for now as it's useless without the CDN settings changes.

diox:
Closing for now. Let's re-open when we want to do the S3 switch.

PR #3313, opened 2015-08-31
Use @use_master in a few tasks called right after a save() (bug 1200267)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1200267


robhudson:
Does order of the decorators matter here?  r+

diox:
It does, on my first attempt I used `@use_master` as the outermost decorator but that doesn't play nice with tasks. `@task` really needs to be the last one called as it completely modifies the function.

PR #3315, opened 2015-09-01
Allows websites to be regionally-restricted to Colombia (bug 1200810):

chuckharmston:
* Creates a new restricted tag (`website-region-co`) that allows editorial to indicate that a website should only be shown in Colombia.
* Updates the base `MultiSearch` endpoint view to add filters preventing websites with that tag from being shown in regions outside of Colombia.
* Updates the base `MultiSearch` endpoint view to add filters allowing websites with that tag to be shown inside of Colombia.
* Disables those filters by default.
* Enables those filters on the Fireplace `MultiSearch` endpoint.

Yes, this is a horrific hack to get this feature request out of the way and allow us to focus on 2.5. Product is aware of this and will not ask for this to be extended to other regions or used past a trial period. If the trial is successful, this commit will be reverted and we will add full region restriction functionality to websites. If not, this commit will be reverted.

r? @ngokevin @robhudson

### DIFF
--- a/mkt/search/views.py
+++ b/mkt/search/views.py
@@ -5,17 +5,19 @@ import json
 from django.db.transaction import non_atomic_requests
 from django.http import HttpResponse
 
+from elasticsearch_dsl import filter as es_filter
 from rest_framework.generics import ListAPIView
 from rest_framework.permissions import AllowAny
 from rest_framework.response import Response
 
+import mkt
 from mkt.api.authentication import (RestOAuthAuthentication,
                                     RestSharedSecretAuthentication)
 from mkt.api.base import CORSMixin, MarketplaceView
 from mkt.api.paginator import ESPaginator
 from mkt.api.permissions import AnyOf, GroupPermission
 from mkt.operators.permissions import IsOperatorPermission
-from mkt.search.forms import ApiSearchForm
+from mkt.search.forms import ApiSearchForm, COLOMBIA_WEBSITE
 from mkt.search.indexers import BaseIndexer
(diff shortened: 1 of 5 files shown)


chuckharmston:
@robhudson would especially appreciate a close look at the ES approach. I'm not sure it is ideal, but I'm also not sure what the ideal way would be.

mkt/search/views.py
```diff
@@ -101,13 +104,23 @@ def get_serializer_context(self):
...
-                       index=self._get_indices(),
-                       doc_type=self._get_doc_types())
-                .extra(_source={'exclude': excluded_fields}))
+        co_filters = self._get_colombia_filter()
+        qs = (Search(using=BaseIndexer.get_es(),
+                     index=self._get_indices(),
+                     doc_type=self._get_doc_types())
+              .extra(_source={'exclude': excluded_fields}))
+        if co_filters:
+            return qs.filter(co_filters)
```
robhudson: Does `.filter` apply directly on itself or does it return a new `qs`?

chuckharmston: It appears to chain, but examples are scarce:

http://elasticsearch-dsl.readthedocs.org/en/latest/search_dsl.html#filters

robhudson: It does return a new `qs` but you're returning it here which is what we want. I mis-read it at first and didn't see the `return`.

robhudson:
This looks fine to me. r+

chuckharmston:
Thanks Rob!

PR #3317, opened 2015-09-02
silence migration change-detector

washort:
I forgot when adding ceae509788 that the migration stuff couldn't detect that no DB changes were needed; it wanted to generate a migration that would drop some foreign keys and re-add the same constraints with the same names. Thus, here are some changes to convince it everything is OK.


robhudson:
r+

PR #3318, opened 2015-09-02
Adds Romania (bug 1192374).

chuckharmston:
Tests will fail because I'm still waiting on a review for the marketplace-constants version I'm updating to. I'll make sure they pass before merging.

r?

### DIFF
--- a/mkt/regions/migrations/0001_initial.py
+++ b/mkt/regions/migrations/0001_initial.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+from django.db import migrations
+
+from mkt.constants import regions
+from mkt.developers.cron import exclude_new_region
+
+
+def exclude_romania(apps, schema_editor):
+    exclude_new_region([regions.ROU])
+
+
+class Migration(migrations.Migration):
+    operations = [
+        migrations.RunPython(exclude_romania),
+    ]
--- a/mkt/settings.py
+++ b/mkt/settings.py
@@ -158,6 +158,7 @@ INSTALLED_APPS = (
     'mkt.prices',
     'mkt.ratings',
     'mkt.receipts',
+    'mkt.regions',
     'mkt.reviewers',
     'mkt.search',
     'mkt.stats',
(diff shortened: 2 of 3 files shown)


mkt/regions/migrations/0001_initial.py
```diff
@@ -0,0 +1,15 @@
...
+from mkt.developers.cron import exclude_new_region
+
+
+def exclude_romania(apps, schema_editor):
+    exclude_new_region([regions.RO])
+
+
+class Migration(migrations.Migration):
+    operations = [
+        migrations.RunPython(exclude_romania),
```
diox: Why ? It's not a new region, it was already there. Also, the constant you are looking for is `ROU`, not `RO`

diox:
We should only need the marketplace-constants bump. Hell, we don't actually need it (fireplace is the only repos that need it, for the mcc detection).

mkt/regions/migrations/0001_initial.py
diox: Oh, and we should consider making `exclude_new_region` migration-aware if we are going to use it in django migrations in the future. Right now importing it imports a whole lot of other stuff and it uses the django model directly, which is a big no-no. It'd probably break if another migration modifies `Webapp` and someone tries to run both migrations in the same run.

diox:
Read https://github.com/mozilla/zamboni/pull/3318#discussion_r38597731 to understand why tests are not passing :)

chuckharmston:
Yep, I know. In a meeting so I haven't been able to loop back yet.

diox:
Why did you merge it with that migration ? As said in my comment, it was not necessary (the region had already been added), and it's even dangerous (because it uses non-frozen code in migrations)

chuckharmston:
Ugh, I had a changeset to make `exclude_new_region` lazily import a region, stashed it to make a rebase, and forgot to get rid of it. Will do so now, thanks for catching that.

diox:
IMHO you should just remove the migration entirely (or well, make it a no-op). There is no need for it, and just importing `exclude_new_region` in it breaks things since it imports non-frozen code in a migration.

PR #3319, opened 2015-09-03
Fix generated package apps that specify a package_file

eviljeff:


washort:
thanks, looks good. r+

PR #3320, opened 2015-09-03
Implement versions for FxOS Add-ons (bug 1201641)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1201641

Summary of the API changes:
* Three new properties are available on add-ons: `description`, `latest_public_version` and `latest_version`.
* Two new endpoints are available to list the versions of an add-on and to get the detail for a particular version.
* `download_url` and `unsigned_download_url` move to the versions objects/endpoints
* `mini_manifest_url` is unchanged, always points to the latest public version for download
* You no longer publish/reject add-ons, you publish/reject versions, and add-ons status inherit the "best" possible status from their versions, or goes back to *incomplete*.
* The review queue is still the same, but also lists any add-on with a non-public version.
* Submission API parameter is now called `validation_id` instead of `upload`.


docs/api/topics/firefox_os_addon_submission.rst
```diff
@@ -94,6 +91,28 @@ Add-on Creation
 
     **Response**
 
-    An :ref:`add-on <addon-response-label>`.
+    The newly created :ref:`add-on <addon-response-label>`. Note that an add-on
+    version is created automatically for you.
 
     :status: 201 successfully created.
+
+Add-on Version Creation
```
ngokevin: maybe we should call them ```Firefox OS Add-on``` everywhere

docs/api/topics/firefox_os_addon_submission.rst
```diff
@@ -94,6 +91,28 @@ Add-on Creation
...
+
+.. _addon-version-post-label:
+
+.. http:post:: /api/v2/extensions/extension/(int:id)|(string:slug)/versions/
+
+    .. note:: Requires authentication, ownership of the add-on and a successful validation result.
+
+    **Request**
+
+    :param upload: the id of the :ref:`validation result <addon_validation>`
```
ngokevin: ```validation_id``` might have more meaning than ```upload```

docs/api/topics/firefox_os_addon_submission.rst
diox: I'm calling them `Firefox OS Add-on` in all page headers, but that was a bit too much redundancy to do that everywhere.

ngokevin: let's abbreviate them to FAds :)

diox: Excellent idea, I'm sure we could get some nice press that way

mkt/extensions/models.py
```diff
@@ -33,96 +34,62 @@
...
-    status = models.PositiveSmallIntegerField(
-        choices=STATUS_CHOICES.items(), db_index=True, default=STATUS_NULL)
 
     objects = ExtensionManager()
 
-    def clean_slug(self):
-        return clean_slug(self, slug_field='slug')
+    @cached_property(writable=True)
+    def latest_public_version(self):
+        return self.versions.filter(status=STATUS_PUBLIC).latest('pk')
```
ngokevin: maybe specify ```get_latest_by='created'```, then just use ```.latest()```

mkt/extensions/models.py
```diff
@@ -179,26 +120,165 @@ def mini_manifest(self):
...
+        """Update `status`, `latest_version` and `latest_public_version`
+        properties depending on the `status` on the ExtensionVersion
+        instances attached to this Extension."""
+        # If there is a public version available, the extension should be
+        # public. If not, and if there is a pending version available, it
+        # should be pending. Otherwise it should just be incomplete.
+        if self.versions.filter(status=STATUS_PUBLIC).exists():
+            self.update(status=STATUS_PUBLIC)
+        elif self.versions.filter(status=STATUS_PENDING).exists():
+            self.update(status=STATUS_PENDING)
```
ngokevin: would this mean the add-on itself would become de-listed? i figure it should be possible to upload a new version of an add-on, keeping the add-on public, but just have a pending version.

mkt/extensions/models.py
```diff
@@ -179,26 +120,165 @@ def mini_manifest(self):
...
+        else:
+            self.update(status=STATUS_NULL)
```
ngokevin: is status_null just a placeholder?

mkt/extensions/models.py
diox: I was thinking about setting `get_latest_by='pk'` actually. `created` is already the default when inheriting from `ModelBase` but I'm not a big fan, `pk` is roughly equivalent and will always be faster.

mkt/extensions/models.py
diox: It means (and is exposed to the API as) "incomplete".

mkt/extensions/tests/test_indexers.py
```diff
@@ -1,24 +1,23 @@
...
         self.indexer = Extension.get_indexer()()
 
-    def _extension_factory(self):
-        return Extension.objects.create(**{
-            'name': u'Test Êxtension',
-            'slug': u'test-ëxtension',
-            'status': STATUS_PENDING,
-            'version': '0.42',
-        })
+    def _extension_factory(self, status=STATUS_PENDING):
```
ngokevin: might benefit from having one of these in site.tests

mkt/extensions/utils.py
```diff
@@ -60,22 +60,32 @@ def decode_manifest(self, manifest_data):
                 _('The addon manifest is not valid JSON.'))
 
     def parse(self):
-        """Parse archive and return extension data.
+        """Parse archive and return extension data as expected by the models.
+
          May raise forms.ValidationError()"""
```
ngokevin: indent looks off

mkt/extensions/models.py
ngokevin: either way is good for me. pk may be a bit faster (almost negligibly faster), although created semantically makes more sense

ngokevin:
looks good to me, up to you if you want another set of eyes since it's a large amount of code.

might want to think about ```s/upload/validation_id```

mkt/extensions/models.py
```diff
@@ -179,26 +120,165 @@ def mini_manifest(self):
...
+            pass
+
+
+class ExtensionVersion(ModelBase):
+    extension = models.ForeignKey(Extension, related_name='versions')
+
+    default_language = models.CharField(default=settings.LANGUAGE_CODE,
+                                        max_length=10)
+    manifest = JSONField()
+    version = models.CharField(max_length=255, default='')
```
eviljeff: Technically, version can only be 23 length, fwiw.  (x.x.x.x where x is an integer from 0 to 65535)

docs/api/topics/firefox_os_addon_submission.rst
diox: Updated to use `validation_id` instead of `upload` (in both add-on and version creation endpoints)

mkt/extensions/models.py
diox: `created` is not indexed atm so I'm keeping `pk` for now

docs/api/topics/firefox_os_addons_review.rst
```diff
@@ -33,27 +35,17 @@ Add-ons Queue
     :status 200: successfully completed.
     :status 403: not authenticated.
 
-
-.. http:get:: /api/v2/extensions/extension/(int:id)|(string:slug)/
```
diox: This API is gone for now. Use the regular detail API when you want to access the detail API for an add-on in the queue instead (as reviewer, you should have access)

diox:
Note: I had to also implement https://bugzilla.mozilla.org/show_bug.cgi?id=1201888 in this PR (separate commit) to avoid a clash with the addon->webapp renaming.

docs/api/topics/firefox_os_addons.rst
```diff
@@ -98,6 +100,54 @@ Add-on Search
...
+    :param meta: :ref:`meta-response-label`.
+    :type meta: object
+    :param objects: A :ref:`listing <objects-response-label>` of :ref:`add-on versions <addon-version-response-label>`.
+    :type objects: array
+
+    :status 200: successfully completed.
+    :status 403: not allowed.
+    :status 404: add-on not found.
+
+.. _addon-version-response-label:
```
chuckharmston: Can we add an example representation of this? We refer to this label all over the place, it'd be nice to see what it actually looks like.

chuckharmston: Might want to model it after [the feed docs](http://firefox-marketplace-api.readthedocs.org/en/latest/topics/feed.html).

docs/api/topics/firefox_os_addon_submission.rst
```diff
@@ -88,12 +85,34 @@ Add-on Creation
...
 
     **Response**
 
-    An :ref:`add-on <addon-response-label>`.
+    The newly created :ref:`add-on <addon-response-label>`. Note that an add-on
+    version is created automatically for you.
+
+    :status: 201 successfully created.
+
+Add-on Version Creation
```
chuckharmston: We don't really talk in prose, about the versioning system (i.e. each addon has a version, each version is individually reviewed, etc.). We should do that at the top of this file so headings like this make sense.

docs/api/topics/firefox_os_addons.rst
```diff
@@ -98,6 +100,54 @@ Add-on Search
...
 
 
+Add-on Versions
+===============
+
+.. http:get:: /api/v2/extensions/extension/(int:id)|(string:slug)/versions/
+
+    .. note:: Non public add-ons versions can only be viewed by their authors.
+
+    Returns a list of versions attached to an add-on.
```
chuckharmston: > Returns a list of versions ~~attached to~~ for an add-on.

docs/api/topics/firefox_os_addon_submission.rst
chuckharmston: Also the way the queue works ^

chuckharmston:
r+wc on the docs and summary.

diox:
Thanks, useful feedback, I'll make the suggested documentation changes.

PR #3323, opened 2015-09-04
Add new dev settings using environ

jasonthomas:
This merges settings_mkt.py and settings_base.py into one and makes use of django-environ so that we can manage private settings as environment variables. After everything has been migrated to the new setup we will remove settings_mkt and settings_base.

r?


sites/dev/settings.py
```diff
@@ -0,0 +1,271 @@
...
+
+RECOMMENDATIONS_API_URL = env('RECOMMENDATIONS_API_URL')
+RECOMMENDATIONS_ENABLED = True
+
+DEV_PAY_PROVIDERS = {
+    APP_PURCHASE_TYP: SITE_URL + '/mozpay/?req={jwt}',
+}
+
+# Bug 1145338
+IAF_OVERRIDE_APPS = env.list('IAF_OVERRIDE_APPS')
```
kumar303: ooh, fancy lists

sites/dev/settings.py
```diff
@@ -0,0 +1,271 @@
...
+NEWRELIC_ENABLED_LIST = env.list('NEWRELIC_ENABLED_LIST')
+
+NEWRELIC_ENABLE = HOSTNAME in NEWRELIC_ENABLED_LIST
+
+AES_KEYS = env.dict('AES_KEYS')
+
+TASK_USER_ID = env('TASK_USER_ID', default=4757633)
+SERVE_TMP_PATH = False
+
+CSP_SCRIPT_SRC = CSP_SCRIPT_SRC + (STATIC_URL[:-1],)
```
kumar303: could be `STATIC_URL.rstrip('/')` if you're trying to remove the slash

kumar303:
can you also delete `settings_mkt.py` and `settings_base.py` as part of this patch? Or if you're waiting for server synchronization, at least file a bug to follow up by deleting them?

kumar303:
r+wc from me (old guard!)

diox:
How is this going to affect developers going to dev to launch a django shell there ?

jasonthomas:
I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1201980 for cleanup the settings_mkt.py and settings_base.py settings once the migration is completed.

> How is this going to affect developers going to dev to launch a django shell there ?
For dev (and stage) we are writing the environment settings to file and load them using  environ.Env.read_env().

PR #3324, opened 2015-09-04
Update docker to use celery via redis

robhudson:
Github issue: https://github.com/mozilla/marketplace-env/issues/95

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,5 +42,10 @@ ENV SOLITUDE_URL http://solitude:2602
 ENV STYLUS_BIN /srv/zamboni-node/node_modules/stylus/bin/stylus
 ENV UGLIFY_BIN /srv/zamboni-node/node_modules/uglify-js/bin/uglifyjs
 ENV ZAMBONI_DATABASE mysql://root:@mysql:3306/zamboni
+ENV CELERY_BROKER_URL redis://redis:6379/1
+ENV CELERY_RESULT_BACKEND redis://redis:6379/2
+# Required to run celery as root using pickle serialization.
+# TODO: Remove when we move to json serialization.
+ENV C_FORCE_ROOT 1
 
 EXPOSE 2600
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -42,5 +42,10 @@ ENV SOLITUDE_URL http://solitude:2602
 ENV STYLUS_BIN /srv/zamboni-node/node_modules/stylus/bin/stylus
 ENV UGLIFY_BIN /srv/zamboni-node/node_modules/uglify-js/bin/uglifyjs
 ENV ZAMBONI_DATABASE mysql://root:@mysql:3306/zamboni
+ENV CELERY_BROKER_URL redis://redis:6379/1
+ENV CELERY_RESULT_BAKEND redis://redis:6379/2
```
andymckay: should that be BACKEND?

robhudson: heh, yeah. At least I was consistent!

scripts/docker/supervisor.conf
```diff
@@ -18,3 +18,12 @@ port=9001
 
 [rpcinterface:supervisor]
 supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
+
+[program:celeryd]
+command=python manage.py celery worker --autoreload --broker=redis://redis:6379/1 -Q priority,devhub,images,limited,celery --loglevel DEBUG
```
robhudson: I'll remove the loglevel here, it was for testing. Although it *is* nice for logging. Hmm.

andymckay:
The forcing of JSON solves the problems with users? r+

robhudson:
It forces celery to allow being root and using pickle. So that avoids having to create a user and deal with file permissions yeah. My next goal is to get our tasks to not pass objects and then switch to json which is more secure and then we can drop this.

andymckay:
Sounds great.

scripts/docker/supervisor.conf
robhudson: Yeah, I decided to leave this.

robhudson:
Thanks @andymckay

PR #3325, opened 2015-09-08
Validate extension names.

chuckharmston:
r? @ngokevin @diox

### DIFF
--- a/mkt/extensions/tests/test_validator.py
+++ b/mkt/extensions/tests/test_validator.py
@@ -1,0 +1,41 @@
+from rest_framework.exceptions import ParseError
+
+from mkt.site.tests import TestCase
+from mkt.extensions.validation import ExtensionValidator
+
+
+class TestExtensionValidator(TestCase):
+    """
+    Tests the ExtensionValidator class. The following methods are tested in
+    the TestExtensionViewSetPost test case instead, as part of an end-to-end
+    workflow:
+
+    * ExtensionValidator.validate_file
+    * ExtensionValidator.validate_json
+    """
+    def setUp(self):
+        self.validator = ExtensionValidator()
+        super(TestExtensionValidator, self).setUp()
+
+    def test_name_missing(self):
+        with self.assertRaises(ParseError):
+            self.validator.validate_name({})
+
+    def test_name_not_string(self):
+        with self.assertRaises(ParseError):
+            self.validator.validate_name({'name': 42})
+
+    def test_name_too_short(self):
+        with self.assertRaises(ParseError):
+            self.validator.validate_name({'name': ''})
+
+    def test_name_too_long(self):
+        with self.assertRaises(ParseError):
+            self.validator.validate_name({'name': 'X' * 100})
+
+    def test_name_valid(self):
+        NAME = u'My Lîttle Extension'
+        try:
+            self.validator.validate_name({'name': NAME})
+        except:
+            assert False, u'A valid name "%s" fails validation' % NAME
--- a/mkt/extensions/tests/test_views.py
+++ b/mkt/extensions/tests/test_views.py
@@ -117,6 +117,18 @@ class TestExtensionValidationViewSet(MktPaths, RestOAuth):
                 **headers)
         eq_(response.status_code, 400)
 
+    @mock.patch('mkt.extensions.views.ExtensionValidator.validate')
+    def test_validation_called(self, mock_validate):
+        headers = {
+            'HTTP_CONTENT_TYPE': 'application/foobar',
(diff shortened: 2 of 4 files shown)


mkt/extensions/views.py
```diff
@@ -70,8 +70,22 @@ def create(self, request, *args, **kwargs):
         return Response(serializer.data, status=status.HTTP_202_ACCEPTED)
 
     def validate_upload(self, file_obj):
```
diox: I'd move all this code to a separate file, like `mkt/extensions/validator.py`, with one entry point function (`validate(file_obj)`). Otherwise it's going to grow and grow and grow, and it's already starting to pollute the views.

diox: That would also allow you to remove the weird instantiation of the view class in the tests

mkt/extensions/tests/test_views.py
```diff
@@ -125,6 +128,29 @@ def test_view_result_anonymous(self):
...
+    def test_name_too_short(self):
+        with self.assertRaises(ParseError):
+            self.view.validate_name({'name': ''})
+
+    def test_name_too_long(self):
+        with self.assertRaises(ParseError):
+            self.view.validate_name({'name': 'X' * 100})
+
+    def test_name_valid(self):
+        NAME = 'My Lîttle Extension'
```
diox: You probably want a `u` prefix here and in the assert below as well.

chuckharmston: Did not even notice the circumflex, good call.

chuckharmston:
@diox Sufficiently major refactoring that it probably requires a re-r.

mkt/extensions/validation.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        Validate the uploaded file:
+
+        * Ensure that it is a valid zip file.
+        * Ensure that it contains a valid manifest.json file.
+        * Validate the manifest against the spec.
+        """
+        self.file = file_obj
+        self.manifest_file = None
+        self.manifest_json = None
```
chuckharmston: Could probably run `self.validate` here, but this makes it eminently more testable.

diox: It's better without the `self.validate()` call IMHO.

mkt/extensions/validation.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def validate(self):
```
diox: Quick docstring ?

mkt/extensions/views.py
```diff
@@ -58,7 +57,9 @@ def create(self, request, *args, **kwargs):
         if not file_obj:
             raise exceptions.ParseError(_('Missing file in request.'))
 
-        self.validate_upload(file_obj)  # Will raise exceptions if appropriate.
+        # Will raise exceptions if appropriate.
+        ExtensionValidator(file_obj).validate()
```
diox: At some point we'll have to take care of the `FIXME` a few lines below, or at least not forget it once we run the validation asynchronously, but I can't think of a good way to integrate that with your changes right now.

mkt/extensions/validation.py
```diff
@@ -0,0 +1,92 @@
+import json
+from zipfile import BadZipfile, ZipFile
+
+from rest_framework.exceptions import ParseError
```
diox: I think I would have preferred a custom `Exception` class that would be caught and re-thrown as `ParseError` in the view to make the validation code more agnostic, but it doesn't really matter now

mkt/extensions/views.py
diox: Maybe add a quick view test to make sure we are indeed calling `ExtensionValidator().validate()` (you can just mock it and check that is has been called with the file-like object)

mkt/extensions/validation.py
chuckharmston: > 9:41 AM <mat> chuck: ParseError for the win
> 9:41 AM <mat> catch everything we can and re-throw as ParseError

mkt/extensions/tests/test_validator.py
```diff
@@ -0,0 +1,42 @@
...
+    workflow:
+
+    * ExtensionValidator.validate_file
+    * ExtensionValidator.validate_json
+    """
+    def setUp(self):
+        self.validator = ExtensionValidator()
+        super(TestExtensionValidator, self).setUp()
+
+    def test_name_missing(self):
```
diox: You should test `validate` as well. With a file-like object (IIRC `StringIO` or a mock should work).

mkt/extensions/validation.py
diox: Yeah, but that was true when we were still in the `View` :)

diox:
r+wc

PR #3326, opened 2015-09-09
Remove references to redis in monitor template

jasonthomas:
This fixes jinja2.environment in getattr UndefinedError: 'dict object' has no attribute 'redis'
r?


chuckharmston:
r+

PR #3327, opened 2015-09-09
Validate extension description.

chuckharmston:


mkt/extensions/validation.py
```diff
@@ -13,6 +13,8 @@ class ExtensionValidator(object):
     """
     errors = {
         'BAD_CONTENT_TYPE': _('The file sent has an unsupported content-type'),
+        'DESCRIPTION_TOO_LONG': _(('The `description` property cannot be '
+                                   'longer than 45 characters.')),
```
diox: 123, not 45

mkt/extensions/tests/test_validation.py
```diff
@@ -0,0 +1,100 @@
...
+        })
+        try:
+            ExtensionValidator(extension).validate()
+        except ParseError:
+            assert False, u'Valid extension failed validation.'
+
+    @mock.patch('mkt.extensions.validation.ExtensionValidator.validate_description')
+    @mock.patch('mkt.extensions.validation.ExtensionValidator.validate_name')
+    @mock.patch('mkt.extensions.validation.ExtensionValidator.validate_json')
+    @mock.patch('mkt.extensions.validation.ExtensionValidator.validate_file')
```
chuckharmston: This is really ugly, would love a better way to do this.

diox: Do you just hate the decorator repetition ?

You can probably do this, which is at least a little shorter:
```
    @mock.patch.object(ExtensionValidator, 'validate_description')
    @mock.patch.object(ExtensionValidator, 'validate_name')
    @mock.patch.object(ExtensionValidator, 'validate_json')
    @mock.patch.object(ExtensionValidator, 'validate_file')
```

Alternatively, you could build a list and then use [patch.start](https://docs.python.org/3/library/unittest.mock.html#start-and-stop) but I find it annoying - you have to call the stop method after you test even if it fails, otherwise you end up polluting the rest of the tests. This ends up being verbose and messy.

diox: Oh, there is also http://www.voidspace.org.uk/python/mock/patch.html#patch-multiple which you can use as a context manager. Kinda neat. Doesn't seem to be present in Python 3, it's not a big deal for us but I wonder if its usage has been deprecated.

chuckharmston: `@mock.patch.object` helps, but any time you add a new validation method, you have to add it in 3 places, which kind of sucks. Not much to be done about it, I suppose.

diox: With `patch.multiple`:

```
    def test_calls(self):
        expected_methods = ['validate_description', 'validate_name',
                            'validate_json', 'validate_file']
        mocks = {method: mock.DEFAULT for method in expected_methods}
        with mock.patch.multiple(ExtensionValidator, **mocks):
            self.test_full()
            for method in expected_methods:
                mocked_method = getattr(ExtensionValidator, method)
                eq_(mocked_method.call_count, 1)
```

chuckharmston: Oh, good call. I never considered using `mock.patch.multiple` as a context manager.

diox:
r+

PR #3330, opened 2015-09-10
Updated to stop passing objects to celery (bug 1174219)

robhudson:


mkt/developers/forms.py
```diff
@@ -701,8 +714,8 @@ def save(self, addon, commit=True):
 
             remove_icons(destination)
             tasks.resize_icon.delay(upload_path, destination,
-                                    mkt.CONTENT_ICON_SIZES,
-                                    set_modified_on=[addon])
+                                    mkt.CONTENT_ICON_SIZES)
+            addon.update(modified=datetime.now())
```
diox: How is the `icon_hash` going to be set ? It calls `pngcrush_image` task, but since `set_modified_on` no longer exists, the `icon_hash` is never going to be set I'm afraid. (and there might be other calls to `resize_icon` with the same issue)

mkt/developers/forms.py
```diff
@@ -253,12 +257,21 @@ def save(self, addon, commit=True):
                                    .replace('-', '/'))
                 if filetype in mkt.VIDEO_TYPES:
                     self.instance.update(filetype=filetype)
-                    vtasks.resize_video.delay(upload_path, self.instance.pk,
-                                              user_pk=mkt.get_user().pk)
+                    chain(
```
diox: I wonder if we should have wrappers for these `chain` calls in tasks. Maybe even rename the task to `_resize_video` and move the chaining to a wrapper that would be called `resize_video` ?

mkt/developers/forms.py
diox: If I'm right we're missing a couple tests to cover this issue

mkt/developers/tasks.py
```diff
@@ -361,9 +359,8 @@ def save_promo_imgs(obj, img_content):
     dirname = obj.get_promo_img_dir()
     destination = os.path.join(dirname, '%s' % obj.pk)
     remove_promo_imgs(destination)
-    resize_promo_imgs(
-        tmp_dst, destination, mkt.PROMO_IMG_SIZES,
-        set_modified_on=[obj])
+    kw = resize_promo_imgs(tmp_dst, destination, mkt.PROMO_IMG_SIZES)
```
diox: Add a comment explaining that this only works because you are calling it directly instead of delaying it. Otherwise you'd have to chain a `update_from_data` call

mkt/feed/views.py
```diff
@@ -96,7 +98,16 @@ def post_save(self, obj, created=True):
                 path = obj.image_path(suffix)
                 with public_storage.open(path, 'wb') as f:
                     i.save(f, 'png')
-                pngcrush_image.delay(path, set_modified_on=[obj])
+                print "update hash on %s" % obj._meta.model_name
```
diox: :poodle:

robhudson: heh :penguin:

mkt/lookup/forms.py
```diff
@@ -106,8 +107,8 @@ def save(self, obj, commit=True):
             destination = os.path.join(dirname, '%s' % obj.id)
 
             remove_promo_imgs(destination)
-            resize_promo_imgs(upload_path, destination,
-                              mkt.PROMO_IMG_SIZES, set_modified_on=[obj])
+            resize_promo_imgs(upload_path, destination, mkt.PROMO_IMG_SIZES)
```
diox: Now that you are no longer passing the `obj` here it won't update the hash correctly. You want to check the return value and use it to update the obj or call it async + chain w/ `update_from_data`

diox: Again a test is missing if that didn't trigger a test failure

diox:
Closing for now as robhudson is no longer actively working on it.

PR #3329, opened 2015-09-10
When uploading a new FxOS Add-on version, make older pending versions obsolete (bug 1203594)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1203594

PR #3332, opened 2015-09-11
When uploading a new FxOS Add-on version, make older pending versions obsolete (bug 1203594)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1203594

Made #3329 from the wrong branch on my fork.


diox:
Carrying the r+wc from #3329

PR #3335, opened 2015-09-11
late customization API

washort:


chuckharmston:
Drive-by: can we put it all in a new app? This actually has nothing to do with feed.

mkt/feed/serializers.py
```diff
@@ -485,3 +485,20 @@ def fake_object(self, data):
                     item_type)))
 
         return feed_item
+
+
+class LateCustomizationSerializer(URLSerializerMixin,
+                                  serializers.ModelSerializer):
+    """
+    A serializer for the FeedItem class, which wraps all items that live on the
```
diox: confusing docstring, this doesn't seem feed-related

diox:
+1 on @chuckharmston suggestion, I was confused as to why this is called feed initially.

diox:
Also, bug number ?

mkt/feed/constants.py
```diff
@@ -94,6 +94,7 @@
 FEED_TYPE_BRAND = 'brand'
 FEED_TYPE_COLL = 'collection'
 FEED_TYPE_SHELF = 'shelf'
+FEED_TYPE_LATE_CUSTOMIZATION = 'late-customization'
```
diox: unneeded ?

washort: oops, a leftover from my earlier bad idea :)

mkt/latecustomization/migrations/0001_initial.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+import mkt.constants.carriers
+import mkt.constants.regions
```
diox: Doing mkt imports in migrations is dangerous, as it can attempt to load models before they are ready and break the migration later when changes to that model are introduced in the future. I'm curious, did django generate this one or did you alter it by hand ?

mkt/latecustomization/migrations/0001_initial.py
```diff
@@ -0,0 +1,35 @@
...
+                ('carrier', models.IntegerField(choices=[(15, mkt.constants.carriers.TELENOR), (7, mkt.constants.carriers.KDDI), (11, mkt.constants.carriers.SINGTEL), (13, mkt.constants.carriers.SPRINT), (14, mkt.constants.carriers.TELECOM_ITALIA_GROUP), (17, mkt.constants.carriers.VIMPELCOM), (9, mkt.constants.carriers.MEGAFON), (2, mkt.constants.carriers.AMERICA_MOVIL), (6, mkt.constants.carriers.HUTCHINSON_THREE_GROUP), (19, mkt.constants.carriers.CONGSTAR), (22, mkt.constants.carriers.ORANGE), (21, mkt.constants.carriers.MTN), (0, mkt.constants.carriers.UNKNOWN_CARRIER), (20, mkt.constants.carriers.O2), (12, mkt.constants.carriers.SMART), (5, mkt.constants.carriers.ETISALAT), (18, mkt.constants.carriers.GRAMEENPHONE), (8, mkt.constants.carriers.KT), (10, mkt.constants.carriers.QTEL), (3, mkt.constants.carriers.CHINA_UNICOM), (1, mkt.constants.carriers.TELEFONICA), (16, mkt.constants.carriers.TMN), (4, mkt.constants.carriers.DEUTSCHE_TELEKOM)])),
```
diox: Related: I wish we had a solution for this, it's a bit ridiculous to have the choices in the migration when we know the database is not going to use them.

docs/api/topics/latecustomization.rst
```diff
@@ -0,0 +1,68 @@
...
+.. http:post:: /api/v2/latecustomization/
+
+  Set the late-customization list for a given region and carrier. Replaces
+  existing list.
+
+  **Request**
+
+  :param app: A webapp ID.
+  :type app: int
+  :param region: A region ID.
```
diox: The API (and tests) appear to be accepting slugs, not IDs.

docs/api/topics/latecustomization.rst
```diff
@@ -0,0 +1,68 @@
...
+.. http:get:: /api/v2/latecustomization/?carrier=(str:carrier)&region=(str:region)
+
+  Returns a list of late customization items for this carrier/region.
+
+  **Response**
+
+  :param app: A webapp ID.
+  :type app: int
+  :param region: A region ID.
+  :type region: int
```
diox: Same as below, the API seems to be returning slugs, not IDs (a good thing IMHO, but docs should be updated)

mkt/latecustomization/views.py
```diff
@@ -0,0 +1,50 @@
...
+from mkt.constants.carriers import CARRIER_MAP
+from mkt.constants.regions import REGIONS_DICT
+from mkt.feed.permissions import FeedPermission
+from mkt.feed.views import FeedShelfPermissionMixin, RegionCarrierFilter
+from mkt.latecustomization.models import LateCustomizationItem
+from mkt.latecustomization.serializers import LateCustomizationSerializer
+
+
+class LateCustomizationViewSet(FeedShelfPermissionMixin, CORSMixin,
+                               viewsets.ModelViewSet):
```
diox: `ModelViewSet` has all the HTTP verbs, but you don't seem to want them ? (They are not documented / tested). From the docs it looks like you only need `GenericViewSet` and `mixins.DestroyModelMixin` ?

diox: Also, you're missing `MarketplaceView`, so you don't get the exception handling and the tastypie-style pagination that the rest of the APIs have.

mkt/latecustomization/views.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    Operations for late customization items.
+    """
+    authentication_classes = [RestOAuthAuthentication,
+                              RestSharedSecretAuthentication]
+
+    permission_classes = [FeedPermission]
```
diox: Seems weird that you use `FeedShelfPermissionMixin` and `FeedPermission` outside of the feed when it doesn't seem related to feed.

mkt/latecustomization/views.py
```diff
@@ -0,0 +1,50 @@
...
+        return super(LateCustomizationViewSet, self).list(request, *args,
+                                                          **kwargs)
+
+    def create(self, request, *args, **kwargs):
+        data = self.req_data()
+        try:
+            self.require_operator_permission(
+                request.user, data['carrier'], data['region'])
+        except (KeyError, MultiValueDictKeyError):
+            raise ParseError()
```
diox: Return a meaningful error message ?

mkt/latecustomization/migrations/0001_initial.py
washort: We could make a custom field that doesn't serialize this info in the migrations.

mkt/latecustomization/migrations/0001_initial.py
washort: This was generated.

diox: Damn, that's bad. Those imports should be ok, but that was close.

mkt/latecustomization/views.py
washort: since this is for use by carriers it seems to be the right permission checks.

chuckharmston: It should use https://github.com/mozilla/zamboni/blob/master/mkt/operators/permissions.py#L6

diox: yeah `IsOperatorPermission` seems more suited to that

mkt/urls.py
```diff
@@ -41,6 +42,12 @@
         mini_langpack_manifest, name='langpack.manifest'),
     url('^extension/(?P<uuid>[0-9a-f]{32})/manifest.json$',
         mini_extension_manifest, name='extension.mini_manifest'),
+    url('^late_customization/%s/manifest.webapp$' % mkt.ADDON_UUID,
```
diox: So, we have a different mini-manifest. Is platform doing some transformation ? Or are we going to add a hack to fireplace to have it consider the app is installed no matter what manifest is used ?

washort:
Now with support for addons. re-r?

docs/api/topics/latecustomization.rst
```diff
@@ -0,0 +1,65 @@
...
+  :status 200: Successfully completed.
+
+
+  .. code-block:: json
+
+      {
+          "objects": [
+              {
+                  "app": "carrier-provided-app-1",
+                  "url": "https://marketplace.firefox.com/app/21066c1c-e45d-4538-8f80-96981df51468/manifest.webapp"
```
mstriemer: It would probably be useful to include the type of the object as well. `"type": "app"` or something.

docs/api/topics/latecustomization.rst
```diff
@@ -0,0 +1,65 @@
...
+
+  Set the late-customization list for a given region and carrier. Replaces
+  existing list.
+
+  **Request**
+
+  :param app: A webapp slug, if item is a webapp.
+  :type app: str
+  :param extension: An extension slug, if item is an extension.
+  :type app: str
```
mstriemer: Should this be `:type extension: str`?

mkt/feed/views.py
```diff
@@ -45,6 +45,7 @@
 
 from .fields import DataURLImageField, ImageURLField
 from .models import FeedApp, FeedBrand, FeedCollection, FeedItem, FeedShelf
+
```
mstriemer: This doesn't look like it should be here.

mstriemer:
The API looks pretty limited right now, we'll likely need to add more fields to be able to correctly display it on the frontend.

r+wc

mstriemer:
I had to add the ID for the item to the response so that I could delete them in the management API. In the management API I just hit the regular app URL to get the data for each app but that isn't ideal for a phone.

docs/api/topics/latecustomization.rst
```diff
@@ -0,0 +1,74 @@
...
+                  ...
+              }
+              ...
+          ]
+      }
+
+
+.. http:post:: /api/v2/latecustomization/
+
+  Set the late-customization list for a given region and carrier. Replaces
```
mstriemer: This just creates a single item doesn't it? I don't see how it could replace the existing list. The management portion is currently setup to manage individual items.

washort: No, you're right, that creates a single item, I didn't update the docs after I changed that. :)

mkt/latecustomization/serializers.py
```diff
@@ -0,0 +1,36 @@
+from rest_framework import relations, serializers
+
+import mkt.carriers
+import mkt.regions
+from mkt.api.fields import SlugChoiceField
+from mkt.extensions.serializers import ExtensionSerializer
+from mkt.fireplace.serializers import FireplaceAppSerializer
```
mstriemer: I thought this serializer would work but it doesn't include `regions` which is required by the transonic app widget. Can you switch this to use `mkt.webapps.serializers.SimpleAppSerializer` please?

PR #3336, opened 2015-09-14
Add missing migration following status choices changes

diox:


diox:
No real changes - the choices are actually the same, only the description change.

PR #3337, opened 2015-09-14
Store and expose FxOS Add-ons size in the API (bug 1204532) 

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1204532


ngokevin:
r+

PR #3344, opened 2015-09-16
Moves the account app's migrations to the access app, allowing them to run.

chuckharmston:
r?

### DIFF
--- a/mkt/access/migrations/0003_add_contenttools_groups.py
+++ b/mkt/access/migrations/0003_add_contenttools_groups.py
@@ -2,7 +2,7 @@ from django.db import migrations
 
 
 def create_addon_groups(apps, schema_editor):
-    Group = apps.get_model('Access', 'Group')
+    Group = apps.get_model('access', 'Group')
     Group.objects.create(name='Content Tools: Add-on Submitters',
                          rules='ContentTools:Login,ContentTools:AddonSubmit')
     Group.objects.create(name='Content Tools: Add-on Reviewers',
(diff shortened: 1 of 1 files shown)


mkt/access/migrations/0003_add_contenttools_groups.py
```diff
@@ -1,8 +1,9 @@
 from django.db import migrations
+from django.db.models.loading import get_model
 
 
 def create_addon_groups(apps, schema_editor):
-    Group = apps.get_model('Access', 'Group')
```
diox: lowercase `access`, it's an app name.

chuckharmston: It's actually not an app at all in the Django 1.7 sense, hence tests failing and the below line being added.

diox: and do use `apps.get_model`, never use `django.db.models.loading` in a migration, you might end up with a wrong version of the model.

chuckharmston: You can't use apps.get_model unless you have an AppConfig object, right?

chuckharmston: Okay, I give up. I *definitely* got an exception that said something along the lines of "no AppConfig with that label exists" earlier, but it seems to be working now.

`¯\_(ツ)_/¯`

ngokevin:
r+

PR #3346, opened 2015-09-17
Add information about developer agreement signatory status to FXA login view.

chuckharmston:
r? @ngokevin

### DIFF
--- a/mkt/account/serializers.py
+++ b/mkt/account/serializers.py
@@ -1,4 +1,7 @@
 from functools import partial
+from os import path
+
+from django.conf import settings
 
 from rest_framework import fields, serializers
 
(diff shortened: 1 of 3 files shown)


mkt/account/serializers.py
```diff
@@ -158,3 +161,20 @@ class Meta:
         model = Group
         fields = ('id', 'name', 'restricted')
         read_only_fields = ('id', 'name', 'restricted')
+
+
+class TOSSerializer(serializers.Serializer):
+    has_signed = fields.SerializerMethodField('get_has_signed')
+    tos_url = fields.SerializerMethodField('get_tos_url')
```
ngokevin: maybe just name it url so we can do ```user.tos.url```

ngokevin:
r+

PR #3348, opened 2015-09-18
Don't blindly take Accept-Language; use the lang_from_accept_header.

chuckharmston:
r? @ngokevin


chuckharmston:
In the interest of fixing login on -dev, I'm going to merge this.

PR #3349, opened 2015-09-18
Add last updated date / reviewed date to FxOS Add-ons (bug 1205791)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1205791

### DIFF
--- a/docs/api/topics/firefox_os_addons.rst
+++ b/docs/api/topics/firefox_os_addons.rst
@@ -45,6 +45,7 @@ Detail
             "status": "public",
             "version": "0.1"
           },
+          "last_updated": "2015-09-04T16:16:39",
           "latest_public_version": {
             "id": 1,
             "download_url": "https://example.com/downloads/extension/ce6b52d231154a27a1c54b2648c10379/1/extension-0.1.zip",
@@ -67,6 +68,7 @@ Detail
 
     :resjson int id: The add-on id.
     :resjson string|object|null description: The add-on description.
+    :resjson string|null last_updated: The latest date a version was published at for this add-on. 
     :resjson object latest_version: The latest :ref:`add-on version <addon-version-detail>` available for this extension.
     :resjson object latest_public_version: The latest *public* :ref:`add-on version <addon-version-detail>` available for this extension.
     :resjson string mini_manifest_url: The (absolute) URL to the `mini-manifest <https://developer.mozilla.org/docs/Mozilla/Marketplace/Options/Packaged_apps#Publishing_on_Firefox_Marketplace>`_ for that add-on. That URL may be a 404 if the add-on is not public yet.
--- a/mkt/extensions/indexers.py
+++ b/mkt/extensions/indexers.py
@@ -1,5 +1,7 @@
 from operator import attrgetter
 
+from django.db.models import Min
+
 from mkt.constants.base import STATUS_PUBLIC
 from mkt.search.indexers import BaseIndexer
 from mkt.translations.models import attach_trans_dict
@@ -50,6 +52,8 @@ class ExtensionIndexer(BaseIndexer):
                         'position_offset_gap': 100,
                     },
                     'is_disabled': {'type': 'boolean'},
+                    'last_updated': {'format': 'dateOptionalTime',
+                                     'type': 'date'},
                     'latest_public_version': {
                         'type': 'object',
                         'properties': {
@@ -71,6 +75,8 @@ class ExtensionIndexer(BaseIndexer):
                     },
                     # Name for sorting.
                     'name_sort': cls.string_not_analyzed(doc_values=True),
+                    'reviewed': {'format': 'dateOptionalTime', 'type': 'date',
+                                 'doc_values': True},
                     'guid': cls.string_not_analyzed(),
                     'slug': {'type': 'string'},
                     'status': {'type': 'byte'},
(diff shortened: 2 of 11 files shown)


diox:
In the bug :
>> We can't just use addon.latest_vesion.date_modified?
>
> Mmmm, we could. But then as soon as we introduce a new field that the developer can edit on `ExtensionVersion` we'll need to denormalize. I kinda prefer doing it now, it's not very costly and makes sense.

mkt/extensions/indexers.py
```diff
@@ -113,7 +120,9 @@ def extract_document(cls, pk=None, obj=None):
         else:
             doc['latest_public_version'] = None
         doc['name_sort'] = unicode(obj.name).lower()
-        doc['guid'] = unicode(obj.uuid)
+        # Find the first reviewed date (used in sort).
```
chuckharmston: Is the sort here "first time approved in Marketplace"? That's not particularly intuitive.

mkt/extensions/models.py
```diff
@@ -52,10 +53,11 @@ def public(self):
 
 class Extension(ModelBase):
     # Automatically handled fields.
-    uuid = UUIDField(auto=True)
+    last_updated = models.DateTimeField(db_index=True, null=True)
```
chuckharmston: Why is this nullable? Seems to me we'd want `default=timezone.now` instead.

mkt/extensions/models.py
```diff
@@ -359,9 +363,13 @@ def is_version_number_higher(cls, version1, version2):
         return v1 > v2
 
     def publish(self):
-        """Publish this extension version to public."""
+        """Publish this extension version to public.
```
chuckharmston: This sentence is pretty awkward. Some suggestions:

1. Make this extension version public.
2. Change this extension version's status to public.
3. Publicize this extension version.

mkt/extensions/tests/test_indexers.py
```diff
@@ -12,11 +14,17 @@ class TestExtensionIndexer(TestCase):
     def setUp(self):
         self.indexer = Extension.get_indexer()()
 
-    def _extension_factory(self, status=STATUS_PENDING):
+    def _extension_factory(self, status=STATUS_PENDING, reviewed=None):
+        if reviewed is None:
+            # Microseconds are not saved by MySQL, so set it to 0 to make sure
+            # our comparisons still work once the model is saved to the db.
```
chuckharmston: Good catch. I'm surprised we don't have a helper for this somewhere.

mkt/extensions/tests/test_views.py
```diff
@@ -148,6 +149,11 @@ def setUp(self):
         self.list_url = reverse('api-v2:extension-list')
         self.user = UserProfile.objects.get(pk=2519)
 
+    def tearDown(self):
+        super(TestExtensionViewSetPost, self).tearDown()
+        # Explicitely delete the Extensions to clean up leftover files.
```
chuckharmston: With a quick skim of this commit, I don't see anywhere where you'd be creating files. Should this be there already?

chuckharmston: In general we should be mocking things instead of using the filesystem, anyway.

docs/api/topics/firefox_os_addons.rst
```diff
@@ -67,6 +68,7 @@ Detail
 
     :resjson int id: The add-on id.
     :resjson string|object|null description: The add-on description.
+    :resjson string|null last_updated: The latest date a version was published at for this add-on.
```
chuckharmston: What about `reviewed`?

chuckharmston:
r+wc

mkt/extensions/tests/test_views.py
diox: Yeah, sorry, this is a drive-by fix. I stumbled upon it when re-launching the tests locally a bunch of times when writing this PR, but it was caused by earlier commits.

mkt/reviewers/tests/test_views.py
```diff
@@ -1249,7 +1249,7 @@ def test_public_sign(self, sign_mock, json_mock, update_cached_manifests):
         with private_storage.open(
                 self.version.files.all()[0].file_path, 'w') as f:
             f.write('.')
-        public_storage.delete(self.version.files.all()[0].file_path)
+        public_storage.delete(self.version.files.all()[0].signed_file_path)
```
diox: Also a drive-by fix.

mkt/files/models.py
```diff
@@ -227,7 +227,7 @@ def cleanup_file(sender, instance, **kw):
     if kw.get('raw') or not instance.filename:
         return
     # Use getattr so the paths are accessed inside the try block.
-    for path in ('file_path', 'guarded_file_path'):
+    for path in ('file_path', 'guarded_file_path', 'signed_file_path'):
```
diox: drive-by fix.

mkt/extensions/tests/test_views.py
diox: Yeah, but for the upload tests it's kinda painful to fully mock the filesystem so I didn't do it. I did it for the rest of the tests though, like reject/publish.

mkt/extensions/indexers.py
diox: It's the sort used by "new". Updates are not displayed there, only new things, ordered by review date.

mkt/extensions/models.py
diox: We don't, because it's only set when a new version is `*published*`. We don't want it to be updated on other saves.

mkt/extensions/indexers.py
chuckharmston: Huh. I'd expect things to be bumped by updates.

¯\_(ツ)_/¯

docs/api/topics/firefox_os_addons.rst
diox: It does not need to be exposed right now.

mkt/extensions/indexers.py
diox: Yeah. I suspect we did that in order to avoid having developers constantly push new updates to stay at the top of the new page (they'd need reviews, but still)

PR #3350, opened 2015-09-21
Add popularity and trending models for FxOS Add-ons (bug 1206737)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1206737

Base code that we'll need to store popularity (installs numbers) for FxOS Add-ons.

### DIFF
--- a/mkt/extensions/indexers.py
+++ b/mkt/extensions/indexers.py
@@ -84,6 +84,12 @@ class ExtensionIndexer(BaseIndexer):
             }
         }
 
+        # Attach boost field, because we are going to need search by relevancy.
+        cls.attach_boost_mapping(mapping)
+
+        # Attach popularity and trending.
+        cls.attach_trending_and_popularity_mappings(mapping)
+
         # Add extra mapping for translated fields, containing the "raw"
         # translations.
         cls.attach_translation_mappings(mapping, cls.translated_fields)
@@ -123,6 +129,9 @@ class ExtensionIndexer(BaseIndexer):
         # Find the first reviewed date (used in sort).
         doc['reviewed'] = obj.versions.public().aggregate(
             Min('reviewed')).get('reviewed__min')
+
+        # Add boost, popularity, trending values.
+        doc.update(cls.extract_popularity_trending_boost(obj))
 
         # Handle localized fields. This adds both the field used for search and
         # the one with all translations for the API.
--- a/mkt/extensions/models.py
+++ b/mkt/extensions/models.py
@@ -147,6 +147,16 @@ class Extension(ModelBase):
     def get_indexer(self):
         return ExtensionIndexer
 
+    def is_dummy_content_for_qa(self):
+        """
+        Returns whether this extension is a dummy extension used for testing
+        only or not.
+
+        Used by mkt.search.utils.extract_popularity_trending_boost() - the
+        method needs to exist, but we are not using it yet.
+        """
+        return False
+
     def is_public(self):
         return self.status == STATUS_PUBLIC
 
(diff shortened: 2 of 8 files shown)


mkt/extensions/models.py
```diff
@@ -145,6 +145,13 @@ def get_fallback(cls):
     def get_indexer(self):
         return ExtensionIndexer
 
+    def is_dummy_content_for_qa(self):
+        """
+        Returns whether this extension is a dummy extension used for testing
```
chuckharmston: This is peculiar. Maybe add a note indicating why this is needed, rather than just what it does?

mkt/extensions/indexers.py
```diff
@@ -78,6 +78,12 @@ def get_mapping(cls):
             }
         }
 
+        # Attach boost field, because we are going to need search by relevancy.
+        cls.attach_boost_mapping(mapping)
```
chuckharmston: I don't see these here. Are they included via inheritance?

mkt/extensions/migrations/0009_auto_20150921_0817.py
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*-
```
chuckharmston: Can we rename to something descriptive? I really don't like the `auto_timestamp` thing.

mkt/extensions/indexers.py
diox: What's "these" ?

chuckharmston: `attach_boost_mapping`, `cls.attach_trending_and_popularity_mappings`, `doc.update(cls.extract_popularity_trending_boost`

mkt/extensions/models.py
diox: Yeah, `mkt.search.utils.extract_popularity_trending_boost()` uses that. I'll modify the comment.

mkt/extensions/indexers.py
diox: Yep, it's in the base Indexer class. Re-used in webapps, websites and extensions.

mkt/extensions/tests/test_indexers.py
```diff
@@ -79,6 +80,43 @@ def test_extract_with_translations(self):
         eq_(doc['name_l10n_english'], [name['en-US']])
         eq_(doc['name_sort'], name['en-US'].lower())
 
+    def test_popularity(self):
+        extension, version = self._extension_factory()
+        # No installs.
+        doc = self._get_doc(extension)
+        # Boost is multiplied by 4 if it's public.
```
chuckharmston: Is this a magic number? Can we use a constant somewhere instead?

mkt/extensions/tests/test_indexers.py
```diff
@@ -79,6 +80,43 @@ def test_extract_with_translations(self):
...
+        # Adolescent regions popularity value is not stored.
+        ok_('popularity_2' not in doc)
+
+    def test_trending(self):
+        extension, version = self._extension_factory()
+        extension.trending.create(region=0, value=10.0)
+        # Test an adolescent region.
+        extension.trending.create(region=2, value=50.0)
+        # Test a mature region.
+        extension.trending.create(region=7, value=50.0)
```
chuckharmston: This is a little brittle, I don't really like that if we change a region to adolescent, tests will break. They should be independent of the data, you know? Don't see a clean alternative though.

mkt/extensions/tests/test_indexers.py
diox: Somewhat magic number yeah. I'll make it a constant.

mkt/extensions/tests/test_indexers.py
diox: I'll look into mocking the regions so set the adolescent property.

mkt/extensions/tests/test_indexers.py
diox: Done

mkt/extensions/tests/test_indexers.py
diox: Done

mkt/extensions/models.py
diox: Done

mkt/extensions/migrations/0009_auto_20150921_0817.py
diox: Done

PR #3351, opened 2015-09-21
Adds standalone view with simple CSS for developer agreement TOS, useful for iframing.

chuckharmston:
r? @ngokevin

### DIFF
--- a/mkt/account/tests/test_serializers.py
+++ b/mkt/account/tests/test_serializers.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 
 import mock
-from nose.tools import eq_, ok_
+from nose.tools import eq_
 
 import mkt
 import mkt.site.tests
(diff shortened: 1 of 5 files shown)


mkt/developers/templates/developers/terms_standalone.html
```diff
@@ -0,0 +1,29 @@
+<html>
+  <head>
+    {{ css('mkt/devreg') }}
+    <style type="text/css">
```
chuckharmston: This is sort of lame, but some of the CSS is very specific to devhub. This is just enough for a nice, iframe-able presentation:

screen shot 2015-09-21 at 12 16 24 pm

mkt/developers/templates/developers/terms_standalone.html
```diff
@@ -0,0 +1,29 @@
+<html>
+  <head>
+    {{ css('mkt/devreg') }}
```
ngokevin: this seems more CSS than we need, do we actually need this?

chuckharmston: It's a pretty simple way to add all the fonts and resets. I was shooting for simplicity over optimization, given the very specific nature of it.

mkt/account/tests/test_serializers.py
```diff
@@ -86,25 +86,14 @@ class TestTOSSerializer(mkt.site.tests.TestCase):
     def setUp(self):
         self.account = UserProfile()
 
-    def serializer(self, lang='pt-BR'):
```
ngokevin: are we supporting en-gb yet??

ngokevin:
r+wc

mkt/account/tests/test_serializers.py
chuckharmston: Yes, but this is in a test.

chuckharmston: That is no longer necessary, since the standalone TOS view is locale-aware.

chuckharmston:
@ngokevin Which changes?

chuckharmston:
Just pulling in reset, base, and typography now. Should be about the minimum amount of CSS required.

PR #3354, opened 2015-09-22
Add API to delete a FxOS Add-on or one of its versions (bug 1207104)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1207104

### DIFF
--- a/docs/api/topics/firefox_os_addons.rst
+++ b/docs/api/topics/firefox_os_addons.rst
@@ -116,6 +116,18 @@ Search
     :status 200: successfully completed.
 
 
+Delete
+------
+
+.. _addon-delete:
+
+.. http:delete:: /api/v2/extensions/extension/(int:id)|(string:slug)/
+
+    .. note:: Requires authentication. Only works on your own Add-ons.
+
+    Delete an add-on. This action is irreversible.
+
+
 Add-on Versions
 ===============
 
@@ -175,6 +187,19 @@ List
     :status 200: successfully completed.
     :status 403: not allowed.
     :status 404: add-on not found.
+
+Delete
+------
+
+.. _addon-version-delete:
+
+.. http:delete:: /api/v2/extensions/extension/(int:id)|(string:slug)/versions/(int:version_id)/
+
+    .. note::
+        Requires authentication. Only works on versions attached to your
+        your own add-ons.
+
+    Delete an add-on version. This action is irreversible.
 
 .. _addon_statuses:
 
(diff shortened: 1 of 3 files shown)


docs/api/topics/firefox_os_addons.rst
```diff
@@ -116,6 +116,18 @@ Search
...
+Delete
+------
+
+.. _addon-delete:
+
+.. http:delete:: /api/v2/extensions/extension/(int:id)|(string:slug)/
+
+    .. note:: Requires authentication. Only works on your own Add-ons.
+
+    Delete an add-on. This action is irreversible.
```
chuckharmston: ...don't we want soft deletion?

chuckharmston: I'm pretty sure that's free with `ModelBase`. Might be wrong on that though.

diox: It's not. Having it causes subtle problems, so as long as no story requires it, I'd prefer not having it for the moment. When/If we add it, I'll change the docs.

mkt/extensions/tests/test_views.py
```diff
@@ -704,6 +744,113 @@ def test_reject(self, remove_signed_file_mock):
...
+        ok_(ExtensionVersion.objects.filter(
+            pk=self.public_version.pk).exists())
+        ok_(not ExtensionVersion.objects.filter(
+            pk=self.pending_version.pk).exists())
+        ok_(ExtensionVersion.objects.filter(
+            pk=self.other_version.pk).exists())
+        self.extension.reload()
+        eq_(self.extension.status, STATUS_PUBLIC)
+
+    def test_delete_both_versions_logged_in_has_rights(self):
```
chuckharmston: This makes me appreciate that mocha pattern:

    it('should delete both versions when logged in if you have rights', function () {

diox: Yeah. OTOH it's nice to have the method names to directly double-click the test name in the results to copy it and relaunch only this one, for instance.

mkt/extensions/views.py
```diff
@@ -115,8 +116,8 @@ def create(self, request, *args, **kwargs):
 
 
 class ExtensionViewSet(CORSMixin, MarketplaceView, CreateExtensionMixin,
-                       SlugOrIdMixin, ListModelMixin, RetrieveModelMixin,
-                       GenericViewSet):
+                       DestroyModelMixin, ListModelMixin, RetrieveModelMixin,
+                       SlugOrIdMixin, GenericViewSet):
```
chuckharmston: Composition is beautiful.

mkt/extensions/tests/test_views.py
ngokevin: also shout out to mocha since you can just do:

```js
it.only('should delete both versions when logged in if you have rights', function () {
```

:)

PR #3355, opened 2015-09-22
Refactor FxOS Add-on statuses constants (bug 1207115)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1207115

### DIFF
--- a/docs/api/topics/firefox_os_addons.rst
+++ b/docs/api/topics/firefox_os_addons.rst
@@ -74,7 +74,7 @@ Detail
     :resjson string slug: The add-on slug (unique string identifier that can be used
         instead of the id to retrieve an add-on).
     :resjson string status: The add-on current status.
-        Can be "incomplete", "pending", or "public".
+        Can be *incomplete*, *pending*, or *public*.
 
     :param int id: The add-on id
     :param string slug: The add-on slug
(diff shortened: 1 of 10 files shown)


docs/api/topics/firefox_os_addons.rst
```diff
@@ -74,7 +74,7 @@ Detail
     :resjson string slug: The add-on slug (unique string identifier that can be used
         instead of the id to retrieve an add-on).
     :resjson string status: The add-on current status.
-        Can be "incomplete", "pending", or "public".
+        Can be "incomplete", *pending*, or *public*.
```
eviljeff: nit: "incomplete"

mkt/constants/base.py
```diff
@@ -92,6 +93,18 @@
     'unlisted': STATUS_UNLISTED,
 }
 
+STATUS_FILE_CHOICES_API_v2 = {
+    STATUS_NULL: 'incomplete',
+    STATUS_PENDING: 'pending',
+    STATUS_PUBLIC: 'public',
+    STATUS_DISABLED: 'obsolete',
```
eviljeff: STATUS_OBSOLETE?

mkt/constants/base.py
```diff
@@ -92,6 +93,18 @@
     'unlisted': STATUS_UNLISTED,
 }
 
+STATUS_FILE_CHOICES_API_v2 = {
```
eviljeff: Could we limit this to the valid choices we know are possible for add-ons (and apps)?

diox: `STATUS_FILE_CHOICES` had the whole list so I also used the whole list here, even though some might be useless. I prefer to not limit this right now, I don't really want to dive deep into the code to find out which ones are not useful right now.

eviljeff:
r+wc

PR #3356, opened 2015-09-22
Add API to let developers disable their FxOS Add-ons (bug 1207158)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1207158

### DIFF
--- a/docs/api/topics/firefox_os_addons.rst
+++ b/docs/api/topics/firefox_os_addons.rst
@@ -38,6 +38,7 @@ Detail
         {
           "id": 1,
           "description": null,
+          "disabled": false,
           "latest_version": {
             "id": 1,
             "download_url": "https://example.com/downloads/extension/ce6b52d231154a27a1c54b2648c10379/1/extension-0.1.zip",
@@ -68,6 +69,8 @@ Detail
 
     :resjson int id: The add-on id.
     :resjson string|object|null description: The add-on description.
+    :resjson boolean disabled: Boolean indicating whether the developer has disabled
+        their add-on or not.
     :resjson string|null last_updated: The latest date a version was published at for this add-on. 
     :resjson object latest_version: The latest :ref:`add-on version <addon-version-detail>` available for this extension.
     :resjson object latest_public_version: The latest *public* :ref:`add-on version <addon-version-detail>` available for this extension.
@@ -100,6 +103,26 @@ List
     :status 200: successfully completed.
     :status 403: not authenticated.
 
+Update
+------
+
+.. http:patch:: /api/v2/extensions/extension/(int:id)|(string:slug)/
+
+    .. note:: Requires authentication and ownership of the Add-on.
+
+    Update some properties of an add-on.
+
+    :param int id: The add-on id
+    :param string slug: The add-on slug
+
+    :reqjson boolean disabled: Boolean indicating whether the developer has disabled
+        their add-on or not.
+    :reqjson string slug: The add-on slug (unique string identifier that can be used
+        instead of the id to retrieve an add-on).
+
+    :status 200: successfully completed.
+    :status 403: not allowed to access this object.
+    :status 404: not found.
 
 Search
 ------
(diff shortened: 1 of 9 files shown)


chuckharmston:
1 test error, 1 test failure.

docs/api/topics/firefox_os_addons.rst
```diff
@@ -38,6 +38,7 @@ Detail
         {
           "id": 1,
           "description": null,
+          "disabled": false,
```
chuckharmston: We probably want to be consistent with other APIs, but I like booleans prefixed with `is_` or `has_`.

diox:
Oops, bad merge, just force-pushed a new version without the 2 offending lines causing the test failures.

docs/api/topics/firefox_os_addons.rst
diox: I could change it to `is_disabled` everywhere, including in the model, if you prefer.

mkt/extensions/views.py
```diff
@@ -197,6 +205,12 @@ def check_permissions(self, request):
         super(ExtensionVersionViewSet, self).check_permissions(request)
 
         extension = self.get_extension_object()
+        # You can't modify or create versions on disabled extensions, even if
+        # you are the owner.
+        if extension.disabled and self.request.method not in SAFE_METHODS:
+            raise PermissionDenied(
+                _(u'Modifying or submitting versions is forbidden for disabled'
+                  u' Add-ons.'))
```
chuckharmston: Do we typically translate API error messages besides for the validator?

docs/api/topics/firefox_os_addons.rst
chuckharmston: Nah, not worth it now. Just musing.

mkt/extensions/views.py
diox: We typically don't. But this message could end up being shown by the frontend, since it can happen at version creation time, so I thought it'd be best to translate it.

docs/api/topics/firefox_os_addons.rst
```diff
@@ -346,6 +379,6 @@ Publishing
     :param int version_id: The add-on version id
 
     :status 202: successfully published.
-    :status 403: not allowed to access this object.
+    :status 403: not allowed to access this object or disabled add-on.
```
chuckharmston: This sentence is a bit unclear. Perhaps: "not allowed to access this object, or you are trying to change properties on a disabled addon, which [is not allowed]()." (link to [a label inserted here](https://github.com/mozilla/zamboni/pull/3356/files#diff-e1e81b1c2b8c89a397b0df2556da9950R216))

mkt/extensions/migrations/0010_auto_20150922_0704.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
```
chuckharmston: Rename migration?

mkt/extensions/models.py
```diff
@@ -36,10 +36,11 @@
 
 class ExtensionManager(ManagerBase):
     def pending(self):
-        return self.filter(versions__status=STATUS_PENDING).order_by('id')
+        return self.filter(
+            disabled=False, versions__status=STATUS_PENDING).order_by('id')
 
     def public(self):
-        return self.filter(status=STATUS_PUBLIC).order_by('id')
+        return self.filter(disabled=False, status=STATUS_PUBLIC).order_by('id')
```
chuckharmston: Maybe add + use a `.enabled` method?

chuckharmston:
+ if tests pass.

mkt/extensions/models.py
diox: I had one originally, but it wasn't used except here so I removed it. Don't want to preliminary refactor things that don't need to be yet.

chuckharmston: I generally write a helper if it'd be used more than once (i.e. here + pending), but it's a very minor point.

PR #3362, opened 2015-09-25
Implement soft-deletion for FxOS Add-ons and versions (bug 1208099)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1208099

### DIFF
--- a/mkt/extensions/indexers.py
+++ b/mkt/extensions/indexers.py
@@ -51,6 +51,7 @@ class ExtensionIndexer(BaseIndexer):
                         'analyzer': 'default_icu',
                         'position_offset_gap': 100,
                     },
+                    'is_deleted': {'type': 'boolean'},
                     'is_disabled': {'type': 'boolean'},
                     'last_updated': {'format': 'dateOptionalTime',
                                      'type': 'date'},
@@ -114,6 +115,7 @@ class ExtensionIndexer(BaseIndexer):
         doc = dict(zip(attrs, attrgetter(*attrs)(obj)))
 
         doc['guid'] = unicode(obj.uuid)
+        doc['is_deleted'] = obj.deleted
         doc['is_disabled'] = obj.disabled
         if obj.status == STATUS_PUBLIC:
             doc['latest_public_version'] = {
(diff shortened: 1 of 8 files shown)


mkt/extensions/indexers.py
```diff
@@ -51,6 +51,7 @@ def get_mapping(cls):
                         'analyzer': 'default_icu',
                         'position_offset_gap': 100,
                     },
+                    'is_deleted': {'type': 'boolean'},
```
diox: In case I forget when merging: this will need a reindex to work properly.

ngokevin:
use the deleted manager here to https://github.com/mozilla/zamboni/blob/master/mkt/comm/models.py#L204

mkt/extensions/models.py
```diff
@@ -35,29 +35,43 @@
...
+        else:
+            return self.get(slug=identifier)
+
     def pending(self):
-        return self.filter(
-            disabled=False, versions__status=STATUS_PENDING)
+        return self.filter(disabled=False, versions__status=STATUS_PENDING)
 
     def public(self):
         return self.filter(disabled=False, status=STATUS_PUBLIC)
```
ngokevin: need to check deleted here?

diox:
I made the filtering of deleted Extensions explicit (the opposite of what's done for Webapps), so there is nothing to change in commbadge, I think, since you want to be able to display the threads from deleted extensions.

mkt/extensions/models.py
diox: It was a conscious decision not to, but up for debate. I separated the status from the deletion state, so technically, if you want to find all public extensions including deleted, you can with `Extension.objects.public()`, and if you want public non-deleted ones, you'd do `Extension.objects.without_deleted().public()`. This might be over-engineering on my part, let me know what you think.

ngokevin: that seems fine as long as we're not using it in the views

ngokevin:
r+ from me

mkt/extensions/models.py
diox: The views have been changed to use `.without_deleted().public()`

PR #3363, opened 2015-09-25
Don't allow add-ons with just whitespace as name (bug 1205474)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1205474


ngokevin:
r+

PR #3366, opened 2015-09-28
Add FxOS Add-ons reviewer mini_manifest and reviewer signed packages (bug 1209120)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1209120

This allows reviewers to install specific versions of add-ons, regardless of their status. Since the manifest URL and the ids in the signature are different, they can be installed side-by-side with the regular public add-on, and even side-by-side other non-public versions of the same add-on. Apps work the same way.


ngokevin:
r+

PR #3370, opened 2015-09-29
Extract and store author name from FxOS Add-ons manifests (bug 1209582)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1209582

Bonus: add searching by author to the API.


mkt/extensions/validation.py
```diff
@@ -168,3 +171,12 @@ def validate_version(self, manifest_json):
             except ValueError:
                 # Not a valid number.
                 raise ParseError(self.errors['VERSION_INVALID'])
+
+    def validate_author(self, manifest_json):
+        """
+        Ensures that, if present, the author property is no longer than
+        255 characters.
+        """
+        description = manifest_json.get('author')
```
eviljeff: description?

mkt/extensions/validation.py
```diff
@@ -19,6 +19,8 @@ class ExtensionValidator(object):
     error.
     """
     errors = {
+        'AUTHOR_TOO_LONG': _(
+            u'The `author` property cannot be longer than 255 characters.'),
```
eviljeff: Device side UX would prefer 128 and that seems enough to me.  Can we change to 128?

mkt/extensions/validation.py
diox: oops

mkt/extensions/validation.py
diox: Sure. Do you have a link for the decision ? I was looking for one and couldn't find any.

eviljeff: I just asked Jacqueline on IRC.  And then confirmed with David Almstrom. If there is a wiki page for the spec it should be (added) there though

chuckharmston: https://wiki.mozilla.org/Marketplace/Add-ons/Validation

diox: Updated to 128. A little weird since description is at 132 though, but I guess that's fine ? I also added a bunch of validation tests.

docs/api/topics/firefox_os_addons.rst
```diff
@@ -133,12 +135,15 @@ Search
...
-    * If a query is passed, order by relevance.
-    * If no query is passed, order by popularity descending.
+    All query parameters are optional. The default sort order when the `sort`
+    parameter is absent depends on whether a search query (`q`) is present or
+    not:
+    * If a search query is passed, order by relevance.
+    * If no search query is passed, order by popularity descending.
 
     :param string q: The search query.
+    :param string author: Filter by author. Requires a case-insensitive
```
chuckharmston: `string|null`?

docs/api/topics/firefox_os_addons.rst
```diff
@@ -68,6 +69,7 @@ Detail
         depending on the query. See :ref:`translations <overview-translations>`.
 
     :resjson int id: The add-on id.
+    :resjson string author: The add-on author, if specified in the manifest.
```
chuckharmston: `string|null`?

mkt/extensions/indexers.py
```diff
@@ -16,12 +16,17 @@ class ExtensionIndexer(BaseIndexer):
...
+        'trending_*',
+        'boost',
+        # 'name' and 'description', as well as their locale variants
+        # ('name_l10n_<language>', etc.) are only used for the query matches,
+        # and are never returned to the client through the API. The fields that
+        # are returned to the API are '*_translations'.
         'name',
         'name_l10n_*',
+        'description',
+        'description_l10n_*',
```
chuckharmston: Not sure what the changes to this file have to do with this commit?

mkt/extensions/migrations/0015_extension_author.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('extensions', '0014_soft_delete'),
```
chuckharmston: Is this a real dependency? Seems to me that it depends on nothing but `0001_initial`

mkt/extensions/models.py
```diff
@@ -75,6 +75,7 @@ class Extension(ModelBase):
 
     # Fields for which the manifest is the source of truth - can't be
     # overridden by the API.
+    author = models.CharField(default='', editable=False, max_length=128)
```
chuckharmston: Why not nullable?

mkt/extensions/tests/test_validation.py
```diff
@@ -108,16 +118,60 @@ def test_description_missing_valid(self):
...
+
+    def test_author_invalid(self):
+        with self.assertRaises(ParseError):
+            self.validator.validate_author({'author': 42})
+        with self.assertRaises(ParseError):
+            self.validator.validate_author({'author': []})
+        with self.assertRaises(ParseError):
+            self.validator.validate_author({'author': {}})
+        with self.assertRaises(ParseError):
+            self.validator.validate_author({'author': None})
```
chuckharmston: `None` is valid here, isn't it?

chuckharmston: Imagining a manifest with `{ "author": null }`

chuckharmston:
Seems like there's a lot of drive-by cleanup in here. It'd be nice if we could factor that out into a separate commit, but...

mkt/extensions/indexers.py
diox: drive-by optimization, I forgot to do this earlier

mkt/extensions/migrations/0015_extension_author.py
diox: It's how makemigrations generated it - it does that, and I guess it's good to avoid conflicts

mkt/extensions/tests/test_validation.py
diox: That's precisely what I was imagining and it doesn't feel valid to me.

mkt/extensions/models.py
diox: I try to avoid it for `CharField`s for new code when I can, because of https://docs.djangoproject.com/en/1.8/ref/models/fields/#null

docs/api/topics/firefox_os_addons.rst
diox: It defaults to an empty string, see below.

mkt/extensions/indexers.py
diox: Splitted into a different commit

diox:
I splitted everything into 3 different commits

PR #3371, opened 2015-09-30
Support FxOS Add-ons in multi-search API (bug 1210029)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1210029

### DIFF
--- a/mkt/search/views.py
+++ b/mkt/search/views.py
@@ -4,6 +4,7 @@ import json
 
 from django.db.transaction import non_atomic_requests
 from django.http import HttpResponse
+from django.utils.functional import lazy
 
 from elasticsearch_dsl import filter as es_filter
 from rest_framework.generics import ListAPIView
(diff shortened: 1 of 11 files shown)


mkt/search/views.py
```diff
@@ -71,37 +74,58 @@ class MultiSearchView(SearchView):
...
+            'index': WebsiteIndexer.get_index()
+        }
+    }, dict)()
+    serializer_classes = {
+        'extension': ESExtensionSerializer,
+        'webapp': ESAppSerializer,
+        'website': ESWebsiteSerializer
+    }
+
+    def get_doc_types_and_indices(self):
```
ngokevin: does this handle the case where doc_type is ```all```?

ngokevin:
flake8

mkt/search/views.py
diox: `all` is unsupported to it just uses the default - ` 'webapp,website'`

ngokevin:
r+

PR #3372, opened 2015-10-02
Implement "rejected" status for FxOS Add-ons (bug 1211023)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1211023

We already had it for versions, but we want it for add-ons as well.


chuckharmston:
And I'll even merge it, so you don't have to.

PR #3374, opened 2015-10-05
Require user to be shown the developer agreement before allowing them to sign it (bug 1210191).

chuckharmston:
r? @ngokevin

### DIFF
--- a/docs/api/topics/accounts.rst
+++ b/docs/api/topics/accounts.rst
@@ -237,9 +237,29 @@ carrier/region pairs.
 Sign Developer Agreement
 ========================
 
-.. http:post:: /api/v2/account/devtos/
+.. _show-agreement:
 
-    Sign the developer agreement for the authenticating user.
+.. http:post:: /api/v2/account/dev-agreement/show/
+
+    Get the developer agreement URL for the authenticating user.
+
+    .. note:: Authentication is required.
+
+    **Response**
+
+    :status 200: successfully viewed developer agreement.
+    :status 201: successfully viewed developer agreement for the first time.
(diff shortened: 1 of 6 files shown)


mkt/users/models.py
```diff
@@ -83,6 +82,9 @@ class UserProfile(OnChangeMixin, ModelBase, AbstractBaseUser):
                             editable=False)
     enable_recommendations = models.BooleanField(default=True)
 
+    shown_dev_agreement = models.DateTimeField(null=True, blank=True)
+    read_dev_agreement = models.DateTimeField(null=True, blank=True)
```
ngokevin: shown vs read? maybe shown vs signed?

docs/api/topics/accounts.rst
```diff
@@ -237,9 +237,29 @@ carrier/region pairs.
...
+    :status 200: successfully viewed developer agreement.
+    :status 201: successfully viewed developer agreement for the first time.
+      The user can now :ref:`sign the agreement <read-agreement>`.
+    :status 400: user has already signed terms of service.
+    :status 403: authentication required.
+    :status 405: invalid HTTP method; only POST is allowed on this endpoint.
+
+.. read-agreement:
+
+.. http:post:: /api/v2/account/dev-agreement/read/
```
ngokevin: feel like s/read/sign makes sense

mkt/users/models.py
chuckharmston: That's what I'd prefer, but I didn't want to change wording with the existing field, which is `read` :/

ngokevin: hmm...add comments at least

ngokevin:
r+wc

PR #3375, opened 2015-10-05
Add addons-related Fireplace views to routes.

chuckharmston:
r? @ngokevin @jasonthomas


ngokevin:
r-, this only affects local zamboni and isn't worth doing in this case.

you want to update nginx configs at https://github.com/mozilla-services/svcops-puppet/

chuckharmston:
We need this eventually, right?

ngokevin:
not really, i already added the routes at https://github.com/mozilla-services/svcops-puppet/commit/d0dfc42982c20022f2ab986b5b0c4d68b432add4

we'd only need specific routes if we wanted route reversal (e.g., link to app detail page in a commbadge email)

PR #3377, opened 2015-10-06
Refactors extension validator to return error key in addition to error message.

chuckharmston:
Opening to run the full test suite.


ngokevin:
there could be an argument made for making constants...your call r+

chuckharmston:
I thought about that, but you can't use constants as keys in Python when using literal dict definition syntax, and manually setting everything in `__init__` seemed cumbersome.

PR #3378, opened 2015-10-06
Correctly add detail to validation exceptions.

chuckharmston:
DRF exceptions apparently don't handle messages in the same way as Python's do by default.


ngokevin:
r+

chuckharmston:
Thanks! Waiting on tests to pass, missed one thing.

PR #3379, opened 2015-10-07
Changes default extension sort to id DESC (bug 1210949).

chuckharmston:
r? @ngokevin 

The right way to fix this is to set up a sort query param, but beta and whatnot. This is probably good enough for now.


ngokevin:
r+

chuckharmston:
Thanks! Tests failed, just pushed a fix. Will let tests finish before merging.

PR #3385, opened 2015-10-09
Allow extension reviewers access to commbadge APIs.

chuckharmston:
r? @ngokevin

### DIFF
--- a/mkt/comm/models.py
+++ b/mkt/comm/models.py
@@ -45,7 +45,8 @@ def check_acls(user, obj, acl_type):
     elif acl_type == 'admin':
         return acl.action_allowed_user(user, 'Admin', '%')
     elif acl_type == 'reviewer':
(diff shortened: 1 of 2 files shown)


mkt/comm/models.py
```diff
@@ -45,7 +45,8 @@ def check_acls(user, obj, acl_type):
     elif acl_type == 'admin':
         return acl.action_allowed_user(user, 'Admin', '%')
     elif acl_type == 'reviewer':
-        return acl.action_allowed_user(user, 'Apps', 'Review')
+        return (acl.action_allowed_user(user, 'Apps', 'Review') or
```
ngokevin: do you care about checking the obj type so you don't give addon reviewers access to apps and vice versa?

ngokevin:
r+with consideration

PR #3386, opened 2015-10-12
Accept MCC/MNC for late-customization listing (bug 1213328)

washort:


mkt/latecustomization/views.py
```diff
@@ -13,6 +16,8 @@
 from mkt.latecustomization.serializers import LateCustomizationSerializer
 from mkt.operators.models import OperatorPermission
 
+COUNTRIES_BY_MCC = {c['mcc']: c['slug'] for c in COUNTRY_DETAILS.itervalues() if 'mcc' in c}
```
mstriemer: This fails PEP8 on Travis.

mkt/latecustomization/views.py
```diff
@@ -41,24 +46,42 @@ class LateCustomizationViewSet(FeedShelfPermissionMixin, CORSMixin,
     filter_backends = [RegionCarrierFilter]
     cors_allowed_methods = ('get', 'post', 'delete')
 
-    def list(self, request, *args, **kwargs):
+    def get_carrier_region(self, request):
         data = self.req_data() or self.request.GET
-        carrier, region = data['carrier'], data['region']
+        if (('region' in data or 'carrier' in data) and
```
mstriemer: This conditions seems redundant. Looks like `(a || b) && (a && b)` to me which is just `a && b`.

mstriemer:
I'm a bit surprised that there's no code to detect this in the `mkt.carriers` or `mkt.regions` modules but I couldn't find anything. I guess fireplace normally does this conversion and passes the slug in.

It might be useful to move the MCC/MNC detection code into the respective modules.

r+wc+tests

mkt/latecustomization/views.py
washort: Yeah it's redundant now, I originally had different error messages for the region/carrier and mnc/mcc cases but that all got smushed together here.

PR #3389, opened 2015-10-15
Adds validation for `icons` property to WebExtension validator.

chuckharmston:
r? @ngokevin 

Intentionally omitted testing that requires an actual extension file; following precedent in `validate_file` and `validate_json`, I'll add that as part of an e2e test through the viewset.

### DIFF
--- a/mkt/extensions/validation.py
+++ b/mkt/extensions/validation.py
@@ -32,6 +32,10 @@ class ExtensionValidator(object):
         'DESCRIPTION_TOO_LONG': _(
             u'The `description` property cannot be '
             u'longer than 132 characters.'),
+        'ICONS_DO_NOT_EXIST': _(u'Not all specified icons exist.'),
+        'ICONS_NO_128': _(
+            u'If defining `icons`, you must include a 128x128 variant.'),
+        'ICONS_INVALID_FORMAT': _(u'Only PNG icons are permitted.'),
         'INVALID_JSON': _(
             u"'manifest.json' in the archive is not a valid JSON"
             u" file."),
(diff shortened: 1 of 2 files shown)


chuckharmston:
Hm. They were passing, now they aren't.

mkt/extensions/validation.py
```diff
@@ -87,6 +92,8 @@ def validate(self):
         self.validate_description(self.data)
         self.validate_version(self.data)
         self.validate_author(self.data)
+        self.validate_icons(self.data)
+        # self.validate_icon_files(self.data)
```
diox: that part should not be commented

diox: or are you leaving that for later ?

chuckharmston: Leaving it for later.

chuckharmston: I wasn't 100% clear on the engineering path forward, so I started stubbing things out here.

mkt/extensions/validation.py
```diff
@@ -202,3 +210,32 @@ def validate_author(self, manifest_json):
...
+    def validate_icons(self, manifest_json):
+        """
+        Validate the `icons` property in the manifest:
+
+        * Ensure that, if the icons property is present, a 128px icon is
+          provided.
+        * Ensure that each icon referenced has a .png extension.
+        """
+        icons = manifest_json.get('icons', {})
+        if len(icons):
```
diox: can just test `if icons:` I think.

diox:
r+wc, we can validate the icon files when working on the extraction code.

PR #3390, opened 2015-10-17
Use the right language when adding name/description for FxOS Add-ons (bug 1215094)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1215094

Contains the code to prevent the bug causing add-ons to have no name from happening again + migration to fix the existing add-ons, as well as a commit to make translations app behaviour more consistent.


washort:
r+

PR #3391, opened 2015-10-19
Never consider deleted versions in Extension.objects.pending() (bug 1213027)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1213027

It's a little unfortunate there is no easy way to combine queryset methods properly here. It would have been great to be able to write something like:

`Extension.objects.without_deleted().filter(disabled=False).filter(
    versions=ExtensionVersion.objects.pending().without_deleted())`

But this causes a nasty subselect. We probably never care about deleted versions when looking for pending add-ons, so let's keep it simple and modify the pending() queryset method directly.

### DIFF
--- a/mkt/extensions/models.py
+++ b/mkt/extensions/models.py
@@ -46,7 +46,8 @@ class ExtensionQuerySet(models.QuerySet):
             return self.get(slug=identifier)
 
     def pending(self):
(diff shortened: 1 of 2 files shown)


ngokevin:
r+

PR #3392, opened 2015-10-19
[wip] Generate new file name when duplicate files are uploaded (bug 1209114).

chuckharmston:


mkt/extensions/models.py
```diff
@@ -483,19 +483,18 @@ def handle_file_upload_operations(self, upload):
...
+        destination = self.file_path
 
         if private_storage.exists(self.file_path):
-            # The filename should not exist. If it does, it means we are trying
-            # to re-upload the same version. This should have been caught
-            # before, so just raise an exception.
-            raise RuntimeError(
-                'Trying to upload a file to a destination that already exists:'
-                ' %s' % self.file_path)
+            destination = private_storage.get_available_name(self.file_path)
```
chuckharmston: I'm pretty sure that this will work in production, but our mock file storage system doesn't appear to handle this well. @diox, thoughts on the approach?

chuckharmston: We could alternately store the latest uploaded version numbers, store it on the addon, and validate that any future version numbers are higher than that.

diox:
I don't follow your approach. The rest of the code relies on `file_path`, not your custom destination, so that should not work.

I think the right way of doing this is to make `filename` depend on the version pk, and add a migration for existing files.

PR #3394, opened 2015-10-20
Homescreen submission & review

washort:


mkt/constants/applications.py
```diff
@@ -58,3 +58,5 @@ def get_device(request):
 
 def get_device_id(request):
     return getattr(get_device(request), 'id', None)
+
+HOMESCREEN = 'homescreen'
```
diox: Is that the tag name ? Weird place to put it, this file only contains devices info.

mkt/webapps/indexers.py
```diff
@@ -297,6 +297,8 @@ def extract_document(cls, pk=None, obj=None):
         d['is_rereviewed'] = is_rereviewed
         d['rereview_date'] = (obj.rereviewqueue_set.get().created
                               if is_rereviewed else None)
+        is_homescreen = obj.tags.filter(tag_text=mkt.HOMESCREEN).exists()
+        d['is_homescreen'] = is_homescreen
```
diox: Would be nice to start adding tests to `mkt/webapps/tests_indexers.py` - it does not cover much right now, but the more we add the quicker coverage will improve :)

mkt/reviewers/utils.py
```diff
@@ -682,6 +684,28 @@ def get_pending_queue(self):
             addon__disabled_by_user=False,
             addon__status=mkt.STATUS_PENDING)
             .exclude(addon__id__in=self.excluded_ids)
+            .exclude(addon__tags__tag_text='homescreen')
```
diox: Do we need to do that in other queues as well ? In what queues should updated homescreens end up ?

washort: I don't know the answer to that one.

washort: Updated homescreens now go in homescreen queue; re-reviews/escalations for homescreens go in app re-review/escalation queues.

PR #3397, opened 2015-10-21
Expose FxOS Add-ons uuid in mini-manifest and API (bug 1216974)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1216974

### DIFF
--- a/mkt/extensions/models.py
+++ b/mkt/extensions/models.py
@@ -270,6 +270,9 @@ class Extension(ModelBase):
         except ExtensionVersion.DoesNotExist:
             return {}
         mini_manifest = {
+            # 'id' here is the uuid, like in sign_file(). This is used by
+            # platform to do blocklisting.
+            'id': self.uuid,
             'name': version.manifest['name'],
             'package_path': version.download_url,
             'size': version.size,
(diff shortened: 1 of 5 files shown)


mkt/extensions/models.py
```diff
@@ -267,6 +267,7 @@ def mini_manifest(self):
         except ExtensionVersion.DoesNotExist:
             return {}
         mini_manifest = {
+            'id': self.uuid,
```
diox: I chose to expose it as `id` to be coherent with signing (which *needs* it to be called `id`). I'm open to comments about why that's a terrible idea.

chuckharmston: It's terrible, but fine.

chuckharmston: We should mention in documentation endpoint that it's a UUID and not auto-incremented, just for posterity or whatever.

mkt/extensions/models.py
```diff
@@ -613,6 +610,16 @@ def reviewer_sign_if_necessary(self):
         if not private_storage.exists(self.reviewer_signed_file_path):
             self.reviewer_sign_file()
 
+    @property
+    def reviewer_identifier(self):
```
chuckharmston: `reviewer_id`? Docstring is great, name is not.

chuckharmston:
r+wc

mkt/extensions/models.py
diox: I don't think the mini-manifest for add-ons is documented anywhere, except on MDN where it basically says "this is a webapp manifest" :( It's really an internal thing.

PR #3398, opened 2015-10-21
Fully validate icons insize zip files for FxOS Add-ons (bug 1212470)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1212470


mkt/extensions/validation.py
```diff
@@ -217,25 +228,45 @@ def validate_icons(self, manifest_json):
...
-        icons = manifest_json.get('icons', {})
-        return icons
+        try:
+            icon_contents = self.zipfile.extract_path(icon_path.lstrip('/'))
+        except KeyError:
+            self.error('ICON_DOES_NOT_EXIST', error_parameter=icon_path)
+            return
+        try:
+            image = Image.open(StringIO(icon_contents))
+            if image.format != 'PNG':
```
chuckharmston: I think we want to use `imghdr.what(None, x)` for this.

mkt/extensions/validation.py
```diff
@@ -217,25 +228,45 @@ def validate_icons(self, manifest_json):
...
+                try:
+                    icon_size = int(icon_size)
+                    if icon_size <= 0:
+                        raise ValueError
+                except ValueError:
+                    self.error('ICON_INVALID_SIZE', error_parameter=icon_path)
+                self._validate_icon_file(icon_path, icon_size)
 
-    def validate_icon_files(self, manifest_json):
+    def _validate_icon_file(self, icon_path, icon_size):
```
chuckharmston: Beautifully clean.

mkt/extensions/validation.py
diox: I'm not sure we need to. According to Pillow docs: 
http://pillow.readthedocs.org/en/3.0.x/reference/Image.html#PIL.Image.open
> This is a lazy operation; this function identifies the file, but the file remains open and the actual image data is not read from the file until you try to process the data

I wonder if accessing `format` and `size` count as really reading image data. I'll look into that, if it does some heavy reads then yeah I'll probably need to use `imghdr`.

mkt/extensions/validation.py
```diff
@@ -70,10 +79,13 @@ def __init__(self, file_obj=None):
...
-    def error(self, error_key):
+    def error(self, error_key, error_parameter=None):
+        message = self.errors[error_key]
+        if error_parameter is not None:
+            message = self.errors[error_key] % error_parameter
         raise ParseError(detail={
             'key': error_key,
-            'message': self.errors[error_key],
+            'message': message,
         })
```
chuckharmston: If it's parameterized, can we return all the parameters in the response as well? That might make construction of a [more user-friendly error message](https://github.com/mozilla/marketplace-content-tools/blob/master/src/media/js/addon/components/validationError.js) a little easier.

mkt/extensions/validation.py
```diff
@@ -70,10 +79,13 @@ def __init__(self, file_obj=None):
         self.file_obj = file_obj
         self.zipfile = None
 
-    def error(self, error_key):
+    def error(self, error_key, error_parameter=None):
```
chuckharmston: `*args` instead of `error_parameter=None`

mkt/extensions/validation.py
```diff
@@ -70,10 +79,13 @@ def __init__(self, file_obj=None):
...
+        message = self.errors[error_key]
+        if error_parameter is not None:
+            message = self.errors[error_key] % error_parameter
```
chuckharmston: `args` instead of `error_parameter`

mkt/extensions/validation.py
chuckharmston: Or I guess we could use keyword arguments as well, to make the specification of multiple params in error messages localizations more flexible.

mkt/extensions/validation.py
diox: any suggestions about what to return to achieve that ?

mkt/extensions/tests/test_views.py
```diff
@@ -1285,7 +1285,7 @@ def test_post_logged_in_with_rights(self):
         }))
         eq_(response.status_code, 201)
         data = response.json
-        eq_(data['size'], 268)  # extension.zip size in bytes.
+        eq_(data['size'], 526)  # extension.zip size in bytes.
```
chuckharmston: Could we read this from the extension file rather than magic numbering it?

chuckharmston: Could read once at the top and store as a constant.

diox: Yeah, I should have done that the last couple times I modified it, I'll do it now. Same for the hash.

mkt/extensions/tests/test_validation.py
```diff
@@ -49,27 +46,41 @@ def assertValidationError(self, key):
...
                 z.writestr('manifest.json', json.dumps(data))
+            if icon is not None:
+                z.writestr('path/to/128.png', icon)
         return self.extension
 
+    def _icon(self, size=(128, 128), format='png'):
+        fake_icon = StringIO()
+        Image.new('RGB', size).save(fake_icon, format)
+        fake_icon.seek(0)
+        return fake_icon.read()
```
chuckharmston: This is great.

mkt/extensions/validation.py
chuckharmston: I was thinking something this simple:

    def error(self, key, *args):
        raise ParseError(detail={
            'key': key,
            'message': self.errors[key] % args,
            'params': args,
        }) 

Then errors can have any number of substitution tokens (often none; `'string' % ()` is valid), so long as the message for the key expects as many as is passed.

chuckharmston: Not sure I made the intent of this totally clear. If using what I suggested above results in a response body of:

    {
        'key': 'ICON_INCORRECT_DIMENSIONS',
        'message': 'The icon file `256.png` dimensions are different from the size',
        'params': [
            '256.png'
        ]
    }

Then we can use a combination of `key` + `params` to construct a far more helpful error message.

mkt/extensions/validation.py
```diff
@@ -32,10 +33,18 @@ class ExtensionValidator(object):
         'DESCRIPTION_TOO_LONG': _(
             u'The `description` property cannot be '
             u'longer than 132 characters.'),
-        'ICONS_DO_NOT_EXIST': _(u'Not all specified icons exist.'),
+        'ICON_INCORRECT_DIMENSIONS': _(
+            u'The icon file `%s` dimensions are different from the size '
+            u'in the manifest'),
```
chuckharmston: The file `%s` is not the specified dimensions of %s x %s.

mkt/extensions/validation.py
```diff
@@ -32,10 +33,18 @@ class ExtensionValidator(object):
...
+        'ICON_DOES_NOT_EXIST': _(
+            u'The icon file `%s` is referenced in the manifest but does not'
+            u' exist in the ZIP file.'),
         'ICONS_NO_128': _(
             u'If defining `icons`, you must include a 128x128 variant.'),
-        'ICONS_INVALID_FORMAT': _(u'Only PNG icons are permitted.'),
+        'ICON_NOT_A_VALID_IMAGE_OR_PNG': _(
+            u'The icon file `%s` is not a valid image or not in PNG format.'),
+        'ICON_NOT_SQUARE': _(
+            u'The icon file `%s` is not a square.'),
```
chuckharmston: You can leave out "a"; square is an adjective as well as a noun.

mkt/extensions/validation.py
diox: Yeah, I can do that.

mkt/extensions/validation.py
```diff
@@ -32,10 +33,18 @@ class ExtensionValidator(object):
...
+            u'The icon file `%s` dimensions are different from the size '
+            u'in the manifest'),
+        'ICON_DOES_NOT_EXIST': _(
+            u'The icon file `%s` is referenced in the manifest but does not'
+            u' exist in the ZIP file.'),
         'ICONS_NO_128': _(
             u'If defining `icons`, you must include a 128x128 variant.'),
-        'ICONS_INVALID_FORMAT': _(u'Only PNG icons are permitted.'),
+        'ICON_NOT_A_VALID_IMAGE_OR_PNG': _(
+            u'The icon file `%s` is not a valid image or not in PNG format.'),
```
chuckharmston: The icon file `%s` is not a valid PNG.

This reads better and is equally accurate (invalid images are also not PNGs).

chuckharmston:
r+wc, nice work!

mkt/extensions/validation.py
diox: Yeah I'll use kwargs thanks

PR #3402, opened 2015-10-22
homescreen nav api (bug 1214745)

washort:


mkt/search/tests/test_views.py
```diff
@@ -1447,22 +1460,30 @@ def test_statsd(self, _mock):
         assert _mock.called
 
     def test_search_no_doc_type_passed(self):
+        self.makeHomescreen()
```
diox: probably best to make a separate test for this IMHO

mkt/search/tests/test_views.py
```diff
@@ -1507,12 +1528,13 @@ def test_search_popularity(self):
         eq_(objs[2]['slug'], self.webapp.app_slug)
 
     def test_search_sort_by_reviewed(self):
+        self.makeHomescreen()
```
diox: same as above, a separate test is probably better, those tests depend on relevancy/reviewed time, so they are quite fragile. your call though.

mkt/search/filters.py
```diff
@@ -162,8 +162,9 @@ def filter_queryset(self, request, queryset, view):
 
 class PublicSearchFormFilter(SearchFormFilter):
     VALID_FILTERS = ['app_type', 'author.raw', 'category', 'device', 'guid',
-                     'installs_allowed_from', 'is_offline', 'manifest_url',
```
diox: I don't see a change here

mkt/search/views.py
```diff
@@ -94,6 +94,7 @@ class MultiSearchView(SearchView):
     serializer_classes = {
         'extension': ESExtensionSerializer,
         'webapp': ESAppSerializer,
+        'homescreen': ESAppSerializer,
```
diox: wait what ? this is not a different type, so this should not be needed

mkt/search/tests/test_views.py
```diff
@@ -1549,6 +1577,20 @@ def test_search_q(self):
         eq_(objs[2]['name'], self.webapp.name)
         eq_(objs[2]['slug'], self.webapp.app_slug)
 
+    def test_homescreen(self):
+        homescreen = self.makeHomescreen()
+        res = self.anon.get(self.url, data={
+            'doc_type': 'extension,webapp,website', 'lang': 'en-US',
+            'is_homescreen': '1'})
```
diox: I think you're missing a test for when it's set to 0

diox: you could have such a test in `test_filters` actually that'd be better

mkt/webapps/indexers.py
```diff
@@ -297,6 +297,7 @@ def extract_document(cls, pk=None, obj=None):
         d['is_rereviewed'] = is_rereviewed
         d['rereview_date'] = (obj.rereviewqueue_set.get().created
                               if is_rereviewed else None)
+        d['is_homescreen'] = obj.is_homescreen()
```
ngokevin: add this to the mapping

mkt/search/tests/test_views.py
ngokevin: +1

mkt/search/tests/test_views.py
```diff
@@ -1404,6 +1404,19 @@ def setUp(self):
             reviewed=self.days_ago(0), status=mkt.STATUS_PUBLIC)
         self.refresh(('webapp', 'website', 'extension'))
 
+    def makeHomescreen(self):
```
ngokevin: underscore

mkt/search/filters.py
```diff
@@ -204,6 +205,23 @@ class ExtensionSearchFormFilter(SearchFormFilter):
     VALID_FILTERS = ['author.raw']
 
 
+class HomescreenFilter(BaseFilterBackend):
+    """
+    A django-rest-framework filter backend that filters in or out homescreens.
+    """
+    def filter_queryset(self, request, queryset, view):
+        is_homescreen = request.GET.get('is_homescreen')
```
ngokevin: maybe we should check webapp_type === 'homescreen' rather than is_homescreen === '0'

ngokevin:
update for re-r

mkt/search/filters.py
washort: explain the difference to me?

mkt/webapps/serializers.py
```diff
@@ -530,6 +534,9 @@ def get_package_path(self, obj):
     def get_file_size(self, obj):
         return obj.es_data.get('file_size')
 
+    def get_is_homescreen(self, obj):
+        return obj.es_data['is_homescreen']
```
diox: this will fail before a reindex is made. use `.get('is_homescreen')`

mkt/search/filters.py
ngokevin: seems to me it's better api design. have one filter webapp_type=whatever rather than is_this and is_that

washort: ah, I see what you mean. fireplace is already using `is_homescreen` so let's fix it after next week

ngokevin: it's not using it, we can just change it here

mkt/search/tests/test_views.py
```diff
@@ -1525,6 +1564,34 @@ def test_search_sort_by_reviewed(self):
...
+
+    def test_search_homescreen_sort_by_reviewed(self):
+        self.makeHomescreen()
+        res = self.anon.get(self.url, data={
+            'doc_type': 'extension,webapp,website', 'lang': 'en-US',
+            'sort': 'reviewed'})
+        eq_(res.status_code, 200)
+        objs = res.json['objects']
+        eq_(len(objs), 4)
+        eq_(objs[0]['doc_type'], 'extension')
```
ngokevin: if testing sort, i imagine just testing the ID is sufficient

ngokevin:
r+wc

PR #3403, opened 2015-10-22
Extract and store icons for FxOS Add-ons, and expose them in the API (bug 1217441)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1217441

### DIFF
--- a/mkt/extensions/migrations/0017_author_field_length.py
+++ b/mkt/extensions/migrations/0017_author_field_length.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('extensions', '0016_reset_extensions_translations_locale'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='extension',
+            name='author',
+            field=models.CharField(default=b'', max_length=128, editable=False),
+            preserve_default=True,
+        ),
+    ]
--- a/mkt/extensions/migrations/0018_change_default_ordering.py
+++ b/mkt/extensions/migrations/0018_change_default_ordering.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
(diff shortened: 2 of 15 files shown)


mkt/webapps/tests/test_crons.py
```diff
@@ -164,7 +164,7 @@ def test_by_webapp(self, sign_mock):
         v1 = self.app.current_version
         call_command('sign_apps', webapps=str(v1.pk))
         file1 = v1.all_files[0]
-        assert sign_mock.called_with(((file1.file_path,
+        sign_mock.assert_called_with(((file1.file_path,
```
diox: drive by fix, unrelated.

mkt/extensions/migrations/0017_author_field_length.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
diox: missing migration.

mkt/extensions/migrations/0018_change_default_ordering.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
```
diox: missing migration.

mkt/extensions/tests/test_tasks.py
```diff
@@ -0,0 +1,76 @@
...
+        self.version = self.extension.versions.create(manifest={
+            'icons': {
+                '128': '/path/to/icon.png'
+            }
+        }, version='0.1')
+
+    @mock.patch('mkt.extensions.tasks.save_icon')
+    @mock.patch('mkt.extensions.tasks.ZipFile')
+    @mock.patch('mkt.extensions.tasks.private_storage')
+    def _test(self, private_storage_mock, ZipFile_mock, save_icon_mock,
```
washort: these kind of mock-heavy internal-behaviour tests make me a little nervous; they definitely get the job done but they're the most exasperating thing when it's time to refactor/change backends. I don't have a suggestion for what specifically to do instead here though.

diox: I agree - I spent a lot of time running around trying to find a better way of doing this, and this is the best I could come up with. I tried to keep the calls and therefore the mocks simple and heavily document + test the various cases to balance the awkwardness of the mocks.

washort:
r+

PR #3410, opened 2015-10-26
Copy content_scripts from manifest to mini-manifest (bug 1218282)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1218282

Firefox OS needs this to display a warning if an add-on modifies the system app.


chuckharmston:
r+

PR #3412, opened 2015-10-26
Add API to block/unblock a FxOS Add-on (bug 1217066)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1217066


washort:
great tests, r+ (after fixing that last one)

PR #3413, opened 2015-10-26
Sort reviewer queue by the creation date, ascending.

chuckharmston:
r? @ngokevin


mkt/extensions/views.py
```diff
@@ -186,7 +186,8 @@ class ReviewerExtensionViewSet(CORSMixin, SlugOrIdMixin, MarketplaceView,
         'post': GroupPermission('ContentTools', 'AddonReview'),
         'get': GroupPermission('ContentTools', 'AddonReview'),
     }),)
-    queryset = Extension.objects.without_deleted().pending()
+    queryset = (Extension.objects.without_deleted().pending()
+                         .order_by('latest_version__created'))
     serializer_class = ExtensionSerializer
 
     @list_route(queryset=Extension.objects.without_deleted().public()
```
ngokevin: here too

ngokevin:
i think a couple of tests would help in case we ever add sorting and such. r+wc

PR #3414, opened 2015-10-26
 Fix fireplace multi-search for FxOS Add-ons (bug 1210029)

diox:


mkt/search/serializers.py
```diff
@@ -141,6 +141,7 @@ def to_native(self, obj):
 
         serializer = self.serializers.get(doc_type)
         if serializer is None:
+            import debug
```
chuckharmston: 😒

diox: lol

diox: interestingly I don't think this branch is ever hit :) I removed it, thanks

chuckharmston:
Better commit message? Should be clear what is being fixed.

chuckharmston:
r+

PR #3415, opened 2015-10-26
Set EXTENSION_ICONS_PATH to the correct path on our sites (bug 1217441)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1217441


jasonthomas:
r+

PR #3420, opened 2015-10-30
Change ExtensionVersion.filename to depend on pk, not version number (bug 1219789)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1219789

Slightly annoying to do this late, but it'd be much more painful to do this migration once we are live.

### DIFF
--- a/mkt/extensions/migrations/0020_change_filename_scheme.py
+++ b/mkt/extensions/migrations/0020_change_filename_scheme.py
@@ -1,0 +1,67 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import os.path
+
+from django.conf import settings
+from django.db import migrations
+
+from mkt.site.storage_utils import (move_stored_file, private_storage,
+                                    public_storage)
+
+
+def move_files_to_their_new_locations(apps, schema_editor):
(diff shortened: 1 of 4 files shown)


ngokevin:
"to depend on pk"? in commit message

diox:
oops

mkt/extensions/migrations/0020_change_filename_scheme.py
```diff
@@ -0,0 +1,59 @@
...
+import os.path
+
+from django.conf import settings
+from django.db import migrations
+
+from mkt.site.storage_utils import (move_stored_file, private_storage,
+                                    public_storage)
+
+
+def move_files_to_their_new_locations(apps, schema_editor):
```
ngokevin: how do we know this migration works? it's a lot of code

ngokevin:
i'm worried about the migration, though testing it would involve a lot of mocking. but everything else looks good r+

mkt/extensions/migrations/0020_change_filename_scheme.py
diox: That's a fair question. I tested it locally manually, but it's super annoying to test migrations (they work on frozen models etc). If you can test it locally (with some existing extensionversions) that'd be cool

PR #3424, opened 2015-11-02
Always return True for content_tools_login and content_tools_addon_submit (bug 1220623).

chuckharmston:


diox:
r+ as a temporary fix if tests pass

chuckharmston:
Yeah, tests won't pass. I'm letting Travis find all the ones that fail for me.

Thanks!

PR #3427, opened 2015-11-04
Separate indexer for homescreens (bug 1219333)

washort:


mkt/search/views.py
```diff
@@ -94,6 +99,7 @@ class MultiSearchView(SearchView):
     serializer_classes = {
         'extension': ESExtensionSerializer,
         'webapp': ESAppSerializer,
+        'homescreen': ESAppSerializer,
```
diox: Also update fireplace's `MultiSearchView`

mkt/detail/tests/test_views.py
```diff
@@ -171,10 +171,11 @@ def test_queries(self, _mock):
 
         The queries we are expecting are:
           * 2 savepoints
+          * 1 tag query for distinguishing homescreen/webapps in indexer task
```
diox: ? This is a `GET`, why is it calling the indexer ?

mkt/downloads/tests/test_views.py
```diff
@@ -56,7 +56,7 @@ def test_download_already_signed(self):
             self.app.sign_if_packaged(self.file.version_id)
 
         # Now download and look at the number of queries.
-        with self.assertNumQueries(2):
+        with self.assertNumQueries(3):
```
diox: Same as above, this should not happen during a `GET`.

mkt/fireplace/views.py
```diff
@@ -31,6 +31,7 @@ class MultiSearchView(BaseMultiSearchView):
     serializer_classes = {
         'extension': ESExtensionSerializer,
         'webapp': FireplaceESAppSerializer,
+        'homescreen': FireplaceESAppSerializer,
```
diox: Add a test showing that homescreens are included when requested in `mkt/fireplace/test_views.py`

mkt/search/tests/test_views.py
```diff
@@ -1479,29 +1481,13 @@ def test_search_no_doc_type_passed(self):
...
-        eq_(objs[1]['slug'], self.homescreen.app_slug)
-        eq_(objs[1]['is_homescreen'], True)
-        eq_(objs[2]['doc_type'], 'website')
-        eq_(objs[2]['id'], self.website.pk)
-        eq_(objs[2]['title'], self.website.title)
-        eq_(objs[2]['url'], self.website.url)
-        eq_(objs[2]['promo_imgs']['1050'], '')
-        eq_(objs[2]['promo_imgs']['640'], '')
-        eq_(objs[2]['promo_imgs']['320'], '')
+        eq_(objs[0]['id'], self.homescreen.pk)
```
diox: Also test `objs[0]['doc_type']`

mkt/reviewers/utils.py
```diff
@@ -663,7 +664,12 @@ def get_escalated_queue(self):
                 es_filter.Term(is_disabled=False),
                 es_filter.Term(is_escalated=True),
             ]
-            return WebappIndexer.search().filter('bool', must=must)
+            return (Search(using=WebappIndexer.get_es(),
+                           index=[settings.ES_INDEXES['homescreen'],
+                                  settings.ES_INDEXES['webapp']],
+                           doc_type=['homescreen', 'webapp'])
+                    .extra(_source={'exclude': WebappIndexer.hidden_fields})
+                    .filter('bool', must=must))
```
diox: Can you figure out a way to avoid copy/pasting this below ?

mkt/webapps/tasks.py
```diff
@@ -323,13 +323,23 @@ def update_supported_locales(ids, **kw):
...
+    if webapps:
+        WebappIndexer.index_ids(list(webapps), no_delay=True)
 
 
 @post_request_task(acks_late=True)
 @use_master
 def unindex_webapps(ids, **kw):
     # DEPRECATED: call WebappIndexer.unindexer directly.
+    HomescreenIndexer.unindexer(ids)
     WebappIndexer.unindexer(ids)
```
diox: Can you check if that is covered by tests? If not, add one deleting only a homescreen and another deleting only a webapp ? I'm surprised it's working and does not cause a 404 on elasticsearch and then an exception in our side.

diox: Maybe it's not called at all ?

diox:
This was already a problem in some way, but can you look at feed collections ? I'm not sure we prevent homescreens from being added to a feed collection, and we probably should.

mkt/webapps/tasks.py
```diff
@@ -323,13 +323,23 @@ def update_supported_locales(ids, **kw):
...
+    homescreens = set(
+        Webapp.tags.through.objects.filter(
+            webapp_id__in=ids,
+            tag__tag_text='homescreen')
+        .values_list('webapp_id', flat=True))
+    webapps = set(ids) - homescreens
+    if homescreens:
+        HomescreenIndexer.index_ids(list(homescreens), no_delay=True)
+    if webapps:
+        WebappIndexer.index_ids(list(webapps), no_delay=True)
```
diox: Urgh.

There is got to be a better way. This task is supposed to be deprecated as `obj.get_indexer().index_ids([obj.pk])` does the same job for single objects, and we almost never massively index multiple webapps. Maybe it's a good time to remove it entirely ? If you are dealing with single objects it becomes much easier.

mkt/search/tests/test_views.py
```diff
@@ -1479,29 +1481,13 @@ def test_search_no_doc_type_passed(self):
 
     def test_search_homescreen(self):
```
diox: Add a homescreen to the test above, `test_search_no_doc_type_passed` to test that no homescreens are returned by default

mkt/detail/tests/test_views.py
washort: why indeed. turns out the indexer call is done in a `post_request_task` which carries over from test setup, and is not done by the code being tested at all.

mkt/webapps/tasks.py
washort: it's called in `tearDown` in various search tests and doesn't raise an exception.

mkt/webapps/tasks.py
washort: OK but I'd feel better doing that in a separate PR, even if it's done before this is merged.

mkt/detail/tests/test_views.py
```diff
@@ -175,6 +176,8 @@ def test_queries(self, _mock):
 
         """
         _mock.return_value = (self._mocked_json(), 'fake_etag')
+        # Don't count things left over from setup
+        _discard_tasks()
```
diox: `reset_queries()` instead

mkt/downloads/tests/test_views.py
```diff
@@ -54,7 +55,8 @@ def test_download_already_signed(self):
         with mock.patch.object(packaged, 'sign', mock_sign):
             # Sign the app before downloading, like it would normally happen.
             self.app.sign_if_packaged(self.file.version_id)
-
+        # Don't count things left over from setup.
+        _discard_tasks()
```
diox: `reset_queries()` instead

washort: insufficient; the query is not made until after the request is done.

diox: The `setUp()` should call `_send_tasks()` then. We do that for `ESTestCase`

mkt/webapps/tasks.py
diox: Ah, yeah, `unindexer()` ignores exceptions. It appears to be only called from tests so it should be ok, but we should remove all those ugly `FIXME`s and not blindly call the 2 different unindexers at some point...

mkt/downloads/tests/test_views.py
```diff
@@ -54,7 +56,7 @@ def test_download_already_signed(self):
         with mock.patch.object(packaged, 'sign', mock_sign):
             # Sign the app before downloading, like it would normally happen.
             self.app.sign_if_packaged(self.file.version_id)
-
+        # Don't count things left over from setup.
```
diox: Move that comment above, rewording it to say you are sending the tasks to make sure they don't mess with the `assertNumQueries` calls below ?

mkt/site/tests/__init__.py
```diff
@@ -244,6 +244,7 @@ def options(self, *args, **kw):
 
 ES_patchers = [mock.patch('elasticsearch.Elasticsearch'),
                mock.patch('mkt.websites.indexers.WebsiteIndexer', spec=True),
+               mock.patch('mkt.webapps.indexers.HomescreenIndexer', spec=True),
```
diox: :+1: Can you also add ExtensionIndexer while you are at it ?

mkt/webapps/tests/test_indexers.py
```diff
@@ -263,6 +256,22 @@ def test_trending(self):
...
+        with self.settings(ES_INDEXES={'homescreen': 'homescreens'}):
+            eq_(HomescreenIndexer.get_index(), 'homescreens')
+
+    def test_model(self):
+        eq_(HomescreenIndexer.get_model(), Webapp)
+
+    def test_mapping(self):
+        mapping = HomescreenIndexer.get_mapping()
+        eq_(mapping.keys(), ['homescreen'])
+
```
diox: Forgot this earlier: add a test on `get_indexable` here and in `TestWebappIndexer`, to show that these indexers properly include/ignore homescreens

mkt/fireplace/tests/test_views.py
```diff
@@ -168,13 +169,15 @@ def setUp(self):
...
+        self.website.save()
         self.webapp = Webapp.objects.get(pk=337141)
+        self.webapp.save()
         self.extension = Extension.objects.create(name='test-ext-lol')
         self.extension.versions.create(status=STATUS_PUBLIC)
         self.extension.popularity.add(ExtensionPopularity(region=0, value=999))
-        self.reindex(Extension)
-        self.reindex(Webapp)
-        self.reindex(Website)
+        self.extension.save()
```
diox: Why this change ? `reindex()` would call `save()` + `refresh()` I think, isn't that what we want here ?

washort: `self.reindex()` won't reindex homescreens.

diox: Is it because you want a specific refresh ? I notice `reindex()` would be missing the refresh on homescreens, that kinda suck...

mkt/reviewers/tests/test_views.py
```diff
@@ -582,8 +583,10 @@ def test_queue_count(self):
 
     def test_homescreen_count(self):
         Tag(tag_text='homescreen').save_tag(self.apps[1])
+        self.apps[1].save()
         if self.uses_es():
-            self.reindex(Webapp)
+            WebappIndexer.unindex(self.apps[1].id)
```
diox: Is this necessary just because you changed the app to homescreens after indexation ? I assume it only happens in tests ?

mkt/reviewers/tests/test_views.py
```diff
@@ -962,8 +965,10 @@ def test_queue_count(self):
 
     def test_homescreen(self):
         Tag(tag_text='homescreen').save_tag(self.apps[1])
+        self.apps[1].save()
+        WebappIndexer.unindex(self.apps[1].id)
```
diox: Same question as above

mkt/search/tests/test_views.py
```diff
@@ -1414,7 +1414,8 @@ def make_homescreen(self):
             device_type=mkt.DEVICE_GAIA.id)
         self.homescreen.update(categories=['health-fitness', 'productivity'])
         self.homescreen.update_version()
-        self.refresh(('webapp', 'website', 'extension'))
+        HomescreenIndexer.index_ids([self.homescreen.pk], no_delay=True)
```
diox: `_send_tasks()` instead to avoid explicitly calling `index_ids` ?

diox:
r+wc

mkt/search/tests/test_views.py
diox: Actually `refresh()` should call `_send_tasks()` so why do we need the explicit indexation ?

mkt/fireplace/tests/test_views.py
diox: It should reindex them but indeed does not refresh them... :/

mkt/reviewers/tests/test_views.py
washort: correct

PR #3430, opened 2015-11-06
Return Feature Profile compatibility in App Serializer (bug 1221345)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1221345

- [x] Fix tests, double-check stuff that was using `to_list()`, etc.
- [x] Add feature profile compatibility info to fireplace serializers as well
- [x] Add tests, check coverage
- [x] Add docs


docs/api/topics/apps.rst
```diff
@@ -208,6 +209,12 @@ App
         is marked as compatible with. Currently available devices names are
         ``desktop``, ``android-mobile``, ``android-tablet``, ``firefoxos``.
     :type device_types: array
+    :param feature_compatibility: Boolean indicating whether the app's current
+        version is compatible with the
+        :ref:`feature profile signature <feature-profile-label>`  passed to the
+        API request. If not profile signature was passed or if the backend is
```
washort: s/not/no/

mkt/reviewers/views.py
```diff
@@ -277,7 +277,7 @@ def _review(request, addon, version):
     all_forms = [form, attachment_formset, testedon_formset]
 
     if version:
-        features_list = [unicode(f) for f in version.features.to_list()]
+        features_list = [unicode(f) for f in version.features.to_names()]
```
washort: should this just be `version.features.to_names()`?

mkt/features/tests/test_utils.py
```diff
@@ -0,0 +1,59 @@
...
+import mkt.site.tests
+from mkt.constants.features import FeatureProfile
+from mkt.features.utils import load_feature_profile
+
+
+class TestLoadFeatureProfile(mkt.site.tests.TestCase):
+    def setUp(self):
+        super(TestLoadFeatureProfile, self).setUp()
+        self.profile = FeatureProfile(apps=True)
+        self.signature = self.profile .to_signature()
```
washort: remove extra space

mkt/reviewers/views.py
diox: the names are lazily translated, so if you don't do this, you get proxy objects instead of unicode strings.

mkt/features/tests/test_utils.py
diox: wow, how the hell do the test pass despite this

diox: TIL that's it valid python...

diox: Fixed and renamed the file to avoid clashes with `test_utils` package (we no longer use it, but just in case, I'm following the convention we used in the other modules...)

mkt/webapps/models.py
```diff
@@ -2589,56 +2589,15 @@ class AppFeatures(ModelBase, DynamicBoolFieldsMixin):
...
-        # Grab the profile part of the signature and convert to binary string.
-        try:
-            profile = bin(int(signature.split('.')[0], 16)).lstrip('0b')
-            n = len(fields) - 1
-            for i, f in enumerate(fields):
-                setattr(self, f, bool(int(profile, 2) & 2 ** (n - i)))
-        except ValueError as e:
-            log.error(u'ValueError converting %s. %s' % (signature, e))
-
-    def to_signature(self):
```
washort: this really confused me because I didn't realize that both `AppFeatures` and `FeatureProfile` had this method

diox: Yeah, it confused me as well ! That's why I removed it, trying to keep things simpler.

mkt/reviewers/views.py
washort: ah yes, that. any reason not to force them to strings in `to_names`?

washort:
r+wc

mkt/webapps/models.py
diox: I'll add that it doesn't make sense to generate the signature of `AppFeatures` anyway - you can't compare it to anything, and in the API, you're better of just exporting something more readable if you want to export the list of required features.

mkt/reviewers/views.py
diox: I just realized I went through this though process and... did it already. So this can probably go, I'll give it a try.

PR #3436, opened 2015-11-10
Document limitation about changing the slug of a FxOS Add-On in the API (bug 1221604)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1221604


ngokevin:
r+

PR #3440, opened 2015-11-11
 Add 'firefoxos+tv' as device type (bug 1205503)

washort:


mkt/constants/applications.py
```diff
@@ -29,7 +29,15 @@ class DEVICE_GAIA(object):
     api_name = 'firefoxos'
 
 
-DEVICE_TYPE_LIST = [DEVICE_DESKTOP, DEVICE_MOBILE, DEVICE_TABLET, DEVICE_GAIA]
+class DEVICE_TV(object):
+    id = 5
+    name = _(u'Firefox OS TV')
+    class_name = 'firefoxos+tv'
+    api_name = 'firefoxos+tv'
```
ngokevin: can't we just call it ```tv``` such that we don't have to differentiate it with FirefoxOS

ngokevin:
are we addressing comments?

mkt/constants/applications.py
diox: It is a variant of firefoxos though, and would help going towards having sensible dev/device parameters, like we do for android (where "device" is really the form factor, and "dev" the os, more or less)

mkt/site/middleware.py
```diff
@@ -169,9 +169,10 @@ class DeviceDetectionMiddleware(object):
     def process_request(self, request):
         dev = request.GET.get('dev')
         if dev:
-            setattr(request, 'MOBILE', dev == 'android')
-            setattr(request, 'GAIA', dev == 'firefoxos')
-            setattr(request, 'TABLET', dev == 'desktop')
+            request.MOBILE = dev == 'android'
```
diox: We should remove or rename that `MOBILE` constant, it's confusing. Hell, of those constants are becoming super confusing, `GAIA` doesn't mean much, especially if it's false for tvs.

mkt/constants/applications.py
diox: Also the "+" in api_name/class_name isn't coherent with Android which use "-", so that's wrong. Sorry about that, might have been a wrong suggestion from me (sadly both separators exist in fireplace, but shouldn't in zamboni)

Ideally we'd have:

```
class DESKTOP:
    'api_name': 'desktop'

class FIREFOXOS_MOBILE:
    'api_name': 'firefoxos_mobile'  # aliased to "firefoxos", but "firefoxos" should never be exposed as a choice

class FIREFOXOS_TV:
   'api_name': 'firefoxos_tv'

class ANDROID_MOBILE:
   'api_name': 'android_mobile'

class ANDROID_TABLET:
    'api_name': 'android_tablet'
```

diox:
No yolos please :(

mkt/constants/applications.py
```diff
@@ -41,6 +49,8 @@ class DEVICE_GAIA(object):
     'mobile': DEVICE_MOBILE.id,
     'tablet': DEVICE_TABLET.id,
     'firefoxos': DEVICE_GAIA.id,
+    'firefoxos+mobile': DEVICE_GAIA.id,
```
diox: "firefoxos" should be kept as an alias but not exposed in choices IMHO. Does not necessarily need a migration though, we can convert when instances using the string (Websites, etc) are re-saved.

mkt/features/utils.py
```diff
@@ -11,7 +11,8 @@ def load_feature_profile(request):
     if hasattr(request, 'feature_profile'):
         return
     profile = None
-    if request.GET.get('dev') in ('firefoxos', 'android'):
+    if request.GET.get('dev') in ('firefoxos', 'firefoxos+mobile',
+                                  'firefoxos+tv', 'android'):
```
diox: This list should be a constant instead now that it's becoming quite complex. Also, not sure we want feature detection on TVs  (does that even work for TVs?)

mkt/search/filters.py
```diff
@@ -259,6 +259,7 @@ def filter_queryset(self, request, queryset, view):
 
         device_id = get_device_id(request)
         data = {
+            'tv': getattr(request, 'TV', False),
```
diox: calling it `tv` here but `firefoxos+tv` elsewhere ?

PR #3442, opened 2015-11-12
TV website import support (bug 1223551)

washort:


mkt/websites/management/commands/import_websites_from_csv.py
```diff
@@ -309,6 +319,12 @@ def create_instances(self, data):
 
             with atomic():
                 try:
+                    devices = []
+                    if row['Mobile URL']:
+                        devices += [DEVICE_GAIA.id, DEVICE_MOBILE.id,
+                                    DEVICE_TABLET.id]
+                    if row['TV URL']:
+                        devices.append(DEVICE_TV.id)
```
ngokevin: can we rename DEVICE_TV's string to just 'tv'?

washort: let's talk with mat about that when he's back

ngokevin: @diox

diox: I liked the firefoxos+tv because it was a way to restore the form factor idea. If firefox os for tablets ever comes back, it will be far less messy with that.

diox:
r+

PR #3444, opened 2015-11-16
featured flag for TV apps/websites (bug 1223547)

washort:


mkt/tvplace/tests/test_serializers.py
```diff
@@ -0,0 +1,91 @@
...
+        return a.data
+
+    def test_no_tv_featured(self):
+        res = self.serialize(self.app)
+        ok_(not res['tv_featured'])
+
+    def test_tv_featured(self):
+        Tag(tag_text='featured-tv').save_tag(self.app)
+        self.app.save()
+        self.refresh()
```
washort: @diox halp

this passes if i put `time.sleep(5)` here. am i doing something wrong with ES?

diox: well, `refresh()` is implemented as `pass` above ?

diox: That test should not use ES, right ? Are you talking about the class below ?

washort: yes, I mean the ES subclass -- it of course has the real `refresh` (and leaving out the call to it results in test failure even with delay)

diox: Assuming you are: `refresh()` needs to be called with a doctype (optionally a doctype list). So the issue is probably that it's missing that doctype here.

I suggest raising an error in the original refresh implementation when no doctype is passed : https://github.com/mozilla/zamboni/blob/master/mkt/site/tests/__init__.py#L924

diox: Also, consider doing a mixin instead of directly sublcassing the test class. By subclassing something that inherit from `TestCase` but then adding `ESTestCase` into the mix it ends up having a confusing inheritance chain IMHO. Would be nice to have a quick comment in the the fake `refresh()` method as well to make the idea behind this more obvious.

washort: argh. a brief glance at `ESTestCase.refresh` suggested that it refreshed everything when doctype was not passed.

diox: I wonder if it would be a big perf hit on tests to just let it refresh all known indices... Worth testing someday to avoid further complications and simplify our tests.

mkt/tvplace/tests/test_serializers.py
```diff
@@ -0,0 +1,95 @@
...
+    def test_tv_featured(self):
+        Tag(tag_text='featured-tv').save_tag(self.app)
+        self.app.save()
+        self.refresh('webapp')
+        res = self.serialize(self.app)
+        ok_(res['tv_featured'])
+
+
+class AppSerializerTests(AppSerializerTestsMixin, TestCase):
+    def refresh(self, _):
```
diox: Add a simple comment

mkt/tvplace/tests/test_serializers.py
```diff
@@ -0,0 +1,95 @@
...
+                                     TVWebsiteSerializer,
+                                     TVESWebsiteSerializer)
+from mkt.webapps.indexers import WebappIndexer
+from mkt.websites.indexers import WebsiteIndexer
+from mkt.websites.utils import website_factory
+
+
+class AppSerializerTestsMixin(object):
+
+    def setUp(self):
```
diox: Ah, I wonder if that's going to bite us, particularly in the ES tests, which do need to call `ESTestCase.setUp()`... I'm starting to regret my mixin suggestion... Damn test classes, it's annoying to end up with something clean.

mkt/tvplace/tests/test_serializers.py
```diff
@@ -0,0 +1,95 @@
...
+        Tag(tag_text='featured-tv').save_tag(self.website)
+        self.website.save()
+        self.refresh('website')
+        res = self.serialize(self.website)
+        ok_(res['tv_featured'])
+
+
+class WebsiteSerializerTests(WebsiteSerializerTestsMixin, TestCase):
+
+    def refresh(self, _):
```
diox: Small comment again

mkt/tvplace/views.py
```diff
@@ -0,0 +1,30 @@
...
+
+class MultiSearchView(BaseMultiSearchView):
+    serializer_classes = {
+        'webapp': TVESAppSerializer,
+        'website': TVESWebsiteSerializer,
+    }
+
+    def get_queryset(self):
+        qs = BaseMultiSearchView.get_queryset(self)
+        return qs.filter(Bool(must=[F('term', device=mkt.DEVICE_TV.id)]))
```
diox: Alternatively, just let the client send `dev=firefoxos-tv` and don't bother re-filtering here ?

mkt/tvplace/tests/test_serializers.py
diox: well, tests *are* passing...

mkt/tvplace/views.py
washort: Seems kind of redundant to have to specify `dev=` on a tv-specific endpoint.

mkt/tvplace/tests/test_serializers.py
washort: I'll add the `super` call here, good point.

washort: hah, wait, you fooled me! `ESTestCase` doesn't add a `setUp` method.

diox: ahah! Let's keep it that way, it works.

mkt/tvplace/views.py
diox: Good point.

mkt/webapps/serializers.py
```diff
@@ -135,7 +136,7 @@ class Meta:
             'privacy_policy', 'promo_imgs', 'public_stats', 'release_notes',
             'ratings', 'regions', 'resource_uri', 'slug', 'status',
             'support_email', 'support_url', 'supported_locales', 'tags',
-            'upsell', 'upsold', 'user', 'versions'
+            'upsell', 'upsold', 'user', 'tv_featured', 'versions'
         ]
```
diox: Do we need to add it to the default serializer ?

diox: I think all tv_featured stuff could and should live in `mkt/tv/serializers.py`, unless you see a reason why not ?

diox:
r+wc

PR #3445, opened 2015-11-17
Remove `requires flash` checkbox as no runtime has flash (bug 1170904)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1170904


eviljeff:
r+

PR #3446, opened 2015-11-17
Add 'firefoxos-tv' as device type (bug 1205503)

washort:
This reverts commit 32a694335b447a78efa5baa1e0967b23f7c6e636.


mkt/search/tests/test_views.py
```diff
@@ -612,6 +612,15 @@ def test_device(self):
...
         eq_(obj['slug'], self.webapp.app_slug)
 
+    def test_device_tv(self):
+        AddonDeviceType.objects.create(
+            addon=self.webapp, device_type=DEVICE_CHOICES_IDS['firefoxos-tv'])
+        self.reindex(Webapp)
+        res = self.anon.get(self.url, data={'dev': 'firefoxos-tv'})
+        eq_(res.status_code, 200)
+        obj = res.json['objects'][0]
+        eq_(obj['slug'], self.webapp.app_slug)
```
diox: Add a test on `obj['device_types']` while you are at it ? Just in case.

mkt/submit/forms.py
```diff
@@ -107,6 +107,7 @@ def get_devices(self, source=None):
             source = self._get_combined()
 
         platforms = {'firefoxos': mkt.DEVICE_GAIA,
```
diox: You can probably replace that by dict + return by 

`return map(mkt.DEVICE_LOOKUP.get, source)`

diox:
I notice you haven't modified `DeviceDetectionMiddleware` this time, any reason ? I don't think we need it (and I'd be happy to remove it entirely) but I'm wondering if this omission was intentional.

diox:
r+wc

washort:
omission was intentional, since TV won't need detection.

PR #3451, opened 2015-11-20
Prevent developers from uploading the same add-on twice (bug 1214363)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1214363


eviljeff:
r+

PR #3452, opened 2015-11-23
Add abuse reporting endpoint for extensions aka add-ons.

eviljeff:
Bonus: remove high abuse reporting trigger code that doesn't work.


mkt/abuse/models.py
```diff
@@ -18,13 +19,15 @@ class AbuseReport(ModelBase):
     reporter = models.ForeignKey(UserProfile, null=True,
                                  blank=True, related_name='abuse_reported')
     ip_address = models.CharField(max_length=255, default='0.0.0.0')
-    # An abuse report can be for an addon, a user, or a website. Only one of
-    # these should be set.
+    # An abuse report can be for an app, a user, a website, or an extension.
+    # Only one of these should be set.
     addon = models.ForeignKey(Webapp, null=True, related_name='abuse_reports')
```
diox: Ah, this is not going to be confusing at all :-)

mkt/abuse/serializers.py
```diff
@@ -52,3 +54,13 @@ class WebsiteAbuseSerializer(BaseAbuseSerializer):
 
     class Meta(BaseAbuseSerializer.Meta):
         fields = BaseAbuseSerializer.Meta.fields + ('website',)
+
+
+class ExtensionAbuseSerializer(BaseAbuseSerializer):
+    extension = SplitField(
+        SlugOrPrimaryKeyRelatedField(source='extension', slug_field='slug',
+                                     queryset=Extension.objects.all()),
```
diox: You probably want `Extension.objects.without_deleted().public()` as the queryset instead. I'm surprised this isn't done like this for apps/websites too.

eviljeff: Can you log an abuse report for a deleted <thing> ?  Also, no slugs for deleted, is that a problem?  I had serializers.PrimaryKeyRelatedField at first btw - I think Fireplace just uses the pk.

mkt/abuse/models.py
eviljeff: I thought about changing it but there would be a likely cascade of addon->app changes needed...

diox: Yeah, definitively not the right time for this

mkt/abuse/serializers.py
diox: My point was that you shouldn't be able to report non-public websites, hence my queryset suggestion. So I don't think using the slug is going to be an issue.

eviljeff: oh, I forgot Extensions was the other way around - deleted included in all()

eviljeff: changed

mkt/abuse/tests/test_views.py
```diff
@@ -170,3 +176,28 @@ def test_invalid_website(self):
...
+            'sprout': 'potato',
+            'extension': self.extension.pk
+        }
+
+    def test_invalid_app(self):
+        res, data = self._call(data={'extension': -1})
+        eq_(400, res.status_code)
+        assert 'does not exist' in data['extension'][0]
+
+    def test_slug_app(self):
```
diox: method name contains "app"

mkt/abuse/tests/test_views.py
```diff
@@ -170,3 +176,28 @@ def test_invalid_website(self):
...
+        self.extension = Extension.objects.create(
+            name=u'Test Êxtension')
+        self.extension.update(status=STATUS_PUBLIC)
+        self.default_data = {
+            'text': 'Lies! This extension is an add-on!',
+            'sprout': 'potato',
+            'extension': self.extension.pk
+        }
+
+    def test_invalid_app(self):
```
diox: method name contains "app"

diox: can you also add a test with a deleted extension ?

eviljeff: done

diox:
r+

PR #3455, opened 2015-11-25
Update pyrepo URL following AwS switch (bug 1227779)

diox:


diox:
@jasonthomas I figured Marketplace needed this too ?

jasonthomas:
Correct, marketplace needs this as well.

jasonthomas:
r+

PR #3457, opened 2015-11-30
Add IARC v2 mappings and a basic de-serializer (bug 1228364, bug 1228371)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1228364
https://bugzilla.mozilla.org/show_bug.cgi?id=1228371

The mapping additions should be backwards-compatible - at the moment they do not add new database fields, but even if they did it would still work (they'd still be false and ignored)

### DIFF
--- a/lib/iarc/serializers.py
+++ b/lib/iarc/serializers.py
@@ -1,0 +1,66 @@
+from rest_framework.exceptions import ParseError
+
+from mkt.constants.iarc_mappings import (BODIES, DESCS_V2, INTERACTIVES_V2,
+                                         RATINGS)
+from mkt.site.utils import cached_property
+
+
+class IARCV2RatingListSerializer(object):
+    """Class that is used to de-serialize json sent/received from IARC to our
+    ratings/descriptors/interactive model instances for a given app.
+
+    Not a true rest-framework serializer, because there is little point: it
+    has to deal with several instances for each model, and we don't care about
+    serializing back to json."""
+
+    def __init__(self, instance=None, data=None):
+        self.instance = instance
+        self.data = data
+
+    @cached_property
+    def validated_data(self):
+        validated_data = {
+            'descriptors': set(),
+            'interactives': set(),
+            'ratings': {},
+
+        }
+        rating_list = self.data.get('RatingList')
+        if not rating_list:
+            return None
+
+        for raw_info in rating_list:
+            body = BODIES.get(raw_info['RatingAuthorityShortText'].lower())
+            if not body:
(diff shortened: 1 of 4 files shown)


diox:
@ngokevin: **r?** this does not add the translation layer between the new IARC constants and text yet, but adds a class that should be able to de-serialize data coming from IARC to save in the db.

lib/iarc/serializers.py
```diff
@@ -0,0 +1,66 @@
+from rest_framework.exceptions import ParseError
+
+from mkt.constants.iarc_mappings import (BODIES, DESCS_V2, INTERACTIVES_V2,
+                                         RATINGS)
+from mkt.site.utils import cached_property
+
+
+class IARCV2RatingListSerializer(object):
+    """Class that is used to de-serialize json sent/received from IARC to our
```
ngokevin: does it work with both SearchCerts + PushCerts?

diox: It should. I tried it with the `SearchCerts` demo, and from the documentation it looks like the same format is used for `GetRatingChanges` and `PushCert` as well (we will still need to generate the response for `PushCert` with something else)

diox:
Test failure has been happening since #3454

mkt/constants/iarc_mappings.py
```diff
@@ -189,15 +191,153 @@
...
-    dict([(unicode(v), unicode(k)) for k, v in body_mapping.iteritems()])
-    for body, body_mapping in DESCS.iteritems()]
+# WARNING: When adding a new rating descriptor here also include a migration.
+#          All descriptor keys must be prefixed by the rating body (e.g. USK_).
+#
+# Together with DESCS above these are used to dynamically generate the field
+# list for the RatingDescriptors Django model in mkt.webapps.models.
+# (The model has both v1 and v2 keys, and it's fine, the extra keys are
+#  simply irrelevant depending on the version).
+DESCS_V2 = {
```
ngokevin: might be useful to have a test that all the slugs in DESCS (V1) are a subset of the slugs in DESCS_V2 to check for typos and such

ngokevin:
i'll check out the failure. r+ awesome work

mkt/constants/iarc_mappings.py
diox: This is unfortunately not the case. There are descriptors removed from v1. See https://bugzilla.mozilla.org/show_bug.cgi?id=1228364#c1

ngokevin: I see, great logical organization on that comment.

PR #3460, opened 2015-12-01
Add texts for IARC mappings (bug 1228364)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1228364

Went back & forth a lot and it seems like the simplest solution. We can't pollute the `REVERSE_*` dicts, as they are used for de-serialization with IARC. Instead, we create a new dict with the human-readable text, and use that where we display text.

### DIFF
--- a/mkt/constants/iarc_mappings.py
+++ b/mkt/constants/iarc_mappings.py
@@ -191,6 +191,13 @@ DESCS = {
     },
 }
 
+# Change {body: {'key': 'val'}} to {'val': 'key'}. Used to deserialize our
+# data back to IARC.
+REVERSE_DESCS = {}
+for mapping in [{unicode(v): unicode(k) for k, v in body_mapping.items() if v}
+                for body, body_mapping in DESCS.items()]:
+    REVERSE_DESCS.update(mapping)
+
 # WARNING: When adding a new rating descriptor here also include a migration.
 #          All descriptor keys must be prefixed by the rating body (e.g. USK_).
 #
@@ -327,16 +334,15 @@ DESCS_V2 = {
         'USK_ShopStreamingService': '',
         'USK_Tabakkonsum': 'has_usk_tobacco',
     }
-
 }
 
-# Change {body: {'key': 'val'}} to {'val': 'key'}, combining v1 and v2 dicts.
-_REVERSE_DESCS_BY_BODY = [
-    {unicode(v): unicode(k) for k, v in body_mapping.items() if v}
-    for body, body_mapping in DESCS_V2.items() + DESCS.items()]
-REVERSE_DESCS = {}
-for mapping in _REVERSE_DESCS_BY_BODY:
-    REVERSE_DESCS.update(mapping)
+# Change {body: {'key': 'val'}} to {'val': 'key'}. Used to deserialize our
+# data back to IARC.
+REVERSE_DESCS_V2 = {}
+for mapping in [{unicode(v): unicode(k) for k, v in body_mapping.items() if v}
+                for body, body_mapping in DESCS_V2.items()]:
+    REVERSE_DESCS_V2.update(mapping)
+
 
 # WARNING: When adding a new interactive element here also include a migration.
 #
@@ -349,6 +355,8 @@ INTERACTIVES = {
     'Digital Purchases': 'has_digital_purchases',
 }
 
+REVERSE_INTERACTIVES = {v: k for k, v in INTERACTIVES.items()}
+
 INTERACTIVES_V2 = {
     'IE_UsersInteract': 'has_users_interact',
     'IE_SharesInfo': 'has_shares_info',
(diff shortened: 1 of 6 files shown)


mkt/webapps/migrations/0005_auto_20151201_1858.py
```diff
@@ -0,0 +1,224 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
```
diox: No actual db changes in that migration, as no new descriptors/interactives have been added yet. The migration is only here because of the `help_text` changes.

mkt/developers/templates/developers/apps/ratings/ratings_summary.html
```diff
@@ -40,10 +40,10 @@
                   {% set desc_icon = body.label in desc_icons and desc_icons[body.label][descriptor] %}
                   {% if desc_icon %}
                     <img src="{{ media(desc_icon) }}" class="icon"
-                         title="{{ mkt.iarc_mappings.REVERSE_DESCS[descriptor] }}">
+                         title="{{ mkt.iarc_mappings.HUMAN_READABLE_DESCS_AND_INTERACTIVES[descriptor] }}">
```
diox: I'll double-check if tests existed for this, might add new ones just in case.

mkt/webapps/models.py
```diff
@@ -2512,15 +2518,20 @@ def to_keys_by_body(self, body):
     def iarc_deserialize(self, body=None):
         """Map our descriptor strings back to the IARC ones (comma-sep.)."""
         keys = self.to_keys()
-        # FIXME: convert IARC v2 constants to text.
+        if waffle.switch_is_active('iarc-upgrade-v2'):
```
diox: TODO: add tests for this.

mkt/webapps/models.py
```diff
@@ -2540,15 +2551,21 @@ def __unicode__(self):
         return u'%s: %s' % (self.id, self.addon.name)
 
     def iarc_deserialize(self):
-        # FIXME: convert IARC v2 constants to text.
         """Map our descriptor strings back to the IARC ones (comma-sep.)."""
-        return ', '.join(iarc_mappings.REVERSE_INTERACTIVES.get(inter)
+        if waffle.switch_is_active('iarc-upgrade-v2'):
```
diox: TODO: add tests for this.

mkt/constants/iarc_mappings.py
```diff
@@ -327,16 +336,130 @@
...
+# data back to IARC.
+REVERSE_DESCS_V2 = {}
+for mapping in [{unicode(v): unicode(k) for k, v in body_mapping.items() if v}
+                for body, body_mapping in DESCS_V2.items()]:
+    REVERSE_DESCS_V2.update(mapping)
+
+
+HUMAN_READABLE_DESCS_AND_INTERACTIVES = {
+    # Descriptors:
+    'has_classind_criminal_acts': _(u'Criminal Acts'),
```
ngokevin: this is a mistake since ESRB and rating bodies are very specific on how their text is translated. in the previous implementation, we did not try any l10n on our side which is why i recommended that we map each rating body to a language and serialize the string as is.

ngokevin: so i would just hardcode the strings we were using before

diox: But that only gives us very few languages... and that's making assumptions about the language based on the region... ;/

diox: One thing I started doing before simplifying: import all translations from https://drive.google.com/a/mozilla.com/file/d/0B3KnWfn_4CqGN2dPVmp4cEFFbzNseVM4M010bElWZjA1OVVn/view , except in a specific "iarc" domain that our translators would not touch (and that would be used only to translate those strings)

ngokevin: Might not worth the effort, most rating bodies display icons anyways. The rating bodies themselves make assumptions about the language based on the region so not sure we should try to spend time on this.

Before it was:

- English for ESRB for North America (I think frontend does translation for Spanish).
- English for PEGI for most of Europe.
- German for USK for Germany.
- Portuguese for CLASSIND for Brazil.

diox: Ok, I'll do that

mkt/constants/iarc_mappings.py
```diff
@@ -356,5 +364,11 @@
...
 
-REVERSE_INTERACTIVES = {v: k for k, v
-                        in INTERACTIVES_V2.items() + INTERACTIVES.items()}
+REVERSE_INTERACTIVES_V2 = {v: k for k, v in INTERACTIVES_V2.items()}
+
+# Human readable text for descriptors and interactives should not be translated
+# by us, the rating bodies are very specific about the texts. Instead, we use
+# the V1 descriptors and interactives, since V1 used text instead of constants.
+# When we phase out v1, we'll just add a dict of strings here instead.
+HUMAN_READABLE_DESCS_AND_INTERACTIVES = dict(
```
ngokevin: nice

mkt/constants/iarc_mappings.py
```diff
@@ -191,6 +191,13 @@
     },
 }
 
+# Change {body: {'key': 'val'}} to {'val': 'key'}. Used to deserialize our
```
ngokevin: we should probably add specific example of what these data structures look like. i wrote these and even i sometimes get lost.

ngokevin:
r+ awesome

PR #3461, opened 2015-12-01
Changed fy to fy_NL per pontoon changes. The former is still symlinke…

ddurst:
…d, but may as well make things line up. (pa --> pa_IN was also done, but that language is not yet added.)


diox:
r+

PR #3462, opened 2015-12-02
Script for importing webapps from CSV (bug 1223551)

washort:


washort:
final details will depend on actual data provided for import.

PR #3464, opened 2015-12-03
Implement IARCv2 PushCert API (bug 1228373)

diox:
Not ready yet, missing tests and the code to store the Cert ID. I'm sharing this early to make sure we are on the same page.

### DIFF
--- a/mkt/developers/urls.py
+++ b/mkt/developers/urls.py
@@ -10,7 +10,8 @@ from mkt.api.base import SubRouter
 from mkt.developers.api_payments import (
     AddonPaymentAccountViewSet, PaymentAccountViewSet, PaymentCheckViewSet,
     PaymentDebugViewSet, UpsellViewSet)
-from mkt.developers.views import ContentRatingList, ContentRatingsPingback
+from mkt.developers.views import (ContentRatingList, ContentRatingsPingback,
+                                  ContentRatingsPingbackV2)
 from mkt.inapp.views import InAppProductViewSet, StubInAppProductViewSet
 from mkt.receipts.urls import test_patterns
 from mkt.site.decorators import use_master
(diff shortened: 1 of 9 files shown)


diox:
@ngokevin this is going to conflict with #3459 but I'll rebase it once you merge, it shouldn't be a problem as it's rather small. Let me know if that matches what you had in mind.

mkt/developers/urls.py
```diff
@@ -210,6 +211,10 @@ def provider_patterns(prefix):
     '',
     url(r'^apps/app/(?P<pk>[^/<>"\']+)/content-ratings/pingback/',
         ContentRatingsPingback.as_view(), name='content-ratings-pingback'),
+    # IARC v2 pingback does not use the app pk, instead it uses IARCRequest
+    # uuid, but sends it as POST data.
+    url(r'^apps/iarc/pingback/', ContentRatingsPingbackV2.as_view(),
```
ngokevin: do we call it pushcerts and mention v2?

mkt/developers/views.py
```diff
@@ -1173,3 +1176,30 @@ def verify_data(self, data):
             return False
 
         return True
+
+
+class ContentRatingsPingbackV2(CORSMixin, UpdateModelMixin, GenericAPIView):
+    """Pingback API for IARC v2.
+
+    Should conform to the PushCert API spec. Assumes that the RatingList is
+    always sent with PushCert, so we don't need to call SearchCert afterwards
```
ngokevin: we used to do SearchCert to verify the data is correct for security. i don't think it's a huge deal, but is the UUID enough?

diox: I can do that with the CertID that is returned (and not yet stored, I still need to do this). The uuid becomes pointless once we have the CertID

mkt/developers/urls.py
diox: Maybe yeah. Or just `/iarc/pushcert/` ? I wanted to avoid the "v2" in the URL because it was confusing with our own API version number

ngokevin: ok we can avoid v2

diox:
@ngokevin updated, with tests. I'm not doing the `SearchCert` call yet though, since it's not yet in master.

mkt/developers/views.py
```diff
@@ -396,7 +399,7 @@ def content_ratings_edit(request, addon_id, addon):
             # No IARC request exists. Create.
             iarc_request = IARCRequest.objects.create(
                 app=addon, uuid=uuid.uuid4())
-        ctx['iarc_request_id'] = iarc_request.uuid
+        ctx['iarc_request_id'] = unicode(uuid.UUID(iarc_request.uuid))
```
diox: This is to make sure we're properly formatting the UUID before sending it to IARC through the questionnaire.

ngokevin:
some test issues

diox:
Fixed the flake8 issue, tests are passing.

ngokevin:
r+

PR #3469, opened 2015-12-22
upgrade to restframework 3

washort:


diox:
A bug number would have been nice :)

mkt/account/serializers.py
```diff
@@ -38,42 +35,38 @@ class Meta:
...
+    platform = fields.CharField(required=False, allow_null=True)
 
     def validate(self, attrs):
         attrs = super(FeedbackSerializer, self).validate(attrs)
-
         if not attrs.get('platform'):
             attrs['platform'] = self.request.GET.get('dev', '')
         if self.request.user.is_authenticated():
-            attrs['user'] = self.request.user
+            attrs['user'] = unicode(self.request.user)
```
diox: This looked suspicious to me. I took a quick glance at the code and I don't understand why we'd need that at all, since the view uses the serialized data only to build a context dict for the email function, and overrides the `user` key when it does that.

mkt/account/tests/test_views.py
```diff
@@ -326,7 +326,7 @@ def test_installed_pagination(self):
         eq_(data['meta']['limit'], 2)
         prev = urlparse(data['meta']['previous'])
         eq_(next.path, self.list_url)
-        eq_(QueryDict(prev.query).dict(), {u'limit': u'2', u'offset': u'0'})
```
diox: Shouldn't be an issue, but it makes me nervous that it changed. Making a note to investigate.

mkt/api/fields.py
```diff
@@ -432,8 +412,13 @@ class SemiSerializerMethodField(serializers.SerializerMethodField):
     Used for fields serialized with a method on the serializer but who
     need to handle unserialization manually.
     """
-    def field_from_native(self, data, files, field_name, into):
-        into[field_name] = data.get(field_name, None)
+    def __init__(self, method_name=None, **kwargs):
+        # Intentionally skipping SerializerMethodField.__init__.
```
diox: Why?

mkt/api/paginator.py
```diff
@@ -60,61 +60,55 @@ def page(self, number):
         return page
 
 
-class MetaSerializer(serializers.Serializer):
+class CustomPagination(pagination.LimitOffsetPagination):
     """
-    Serializer for the 'meta' dict holding pagination info that allows to stay
-    backwards-compatible with the way tastypie does pagination (using offsets
-    instead of page numbers), while still using a "standard" Paginator class.
+    Paginator for DRF API views.
```
diox: I would have kept the part of the comment explaining that it's using offsets for backwards-compatibility

mkt/api/paginator.py
```diff
@@ -60,61 +60,55 @@ def page(self, number):
...
+            self.display_page_controls = True
+        page = queryset[self.offset:self.offset + self.limit]
+        if isinstance(page, Search):
+            page = page.execute()
+        return list(page)
+
+
+class PageNumberPagination(pagination.PageNumberPagination):
+
+    def django_paginator_class(self, queryset, page_size):
```
diox: So, I'm not sure I'm following this right, but shouldn't you need this method in the `CustomPagination` above too ? (sharing it with a mixin)

washort: no, `LimitOffsetPagination` does not use the django paginator at all.

mkt/account/serializers.py
washort: so should this field be removed entirely?

mkt/api/fields.py
washort: `SerializerMethodField.__init__` sets the field read-only. I'll clarify the comment.

mkt/api/paginator.py
diox: You removed `ESPaginator` from some views that do ES queries, and `ESPaginator` was useful for those because it prevented the extra count query in ES. So if `LimitOffsetPagination` does not use `ESPaginator`, then the extra count query is going to take place, which is bad. We need a way to fix that.

mkt/api/tests/__init__.py
```diff
@@ -19,7 +21,8 @@ def _allowed_verbs(self, url, allowed):
             if verb in allowed:
                 continue
             try:
-                res = getattr(self.client, verb)(url)
+                with transaction.atomic():
```
diox: Why ? It's inside a `TestCase` class (not a `TransactionTestCase`), I'm not sure this will behave correctly...

mkt/api/tests/test_paginator.py
```diff
@@ -35,15 +35,19 @@ def monkey_search(*args, **kwargs):
 class TestMetaSerializer(TestCase):
```
diox: Rename this

mkt/account/serializers.py
diox: My guess is that's it's only useful to return the `user` in the json returned by the API. So let's keep it, it's not great code but it should work. Maybe add a comment.

mkt/extensions/tests/test_validation.py
```diff
@@ -50,10 +50,10 @@ def assertValidationError(self, key, **kwargs):
             if kwargs:
                 expected_message = expected_message % kwargs
             eq_(e.__class__, ParseError)
-            eq_(e.detail['key'], key)
-            eq_(e.detail['message'], expected_message)
+            eq_(e.detail['detail']['key'], key)
+            eq_(e.detail['detail']['message'], expected_message)
             if kwargs:
-                eq_(e.detail['params'], kwargs)
+                eq_(e.detail['detail']['params'], kwargs)
```
diox: This is a backwards-incompatible change, we should check the content tools to see if they cope correctly with that

mkt/api/tests/__init__.py
diox: Same question for similar `transaction.atomic()` added below.

mkt/search/serializers.py
```diff
@@ -65,34 +65,13 @@ def get_fields(self):
...
-                self._data = self.to_native(self.object)
-        return self._data
-
-    def field_to_native(self, obj, field_name):
-        # DRF's field_to_native calls .all(), which we want to avoid, so we
-        # provide a simplified version that doesn't and just iterates on the
-        # object list.
-        if hasattr(obj, 'object_list'):
-            return [self.to_native(item) for item in obj.object_list]
-        return super(BaseESSerializer, self).field_to_native(obj, field_name)
```
diox: Not sure this was covered by tests, have you checked that this wasn't needed anymore ? IIRC this caused severe performance issues with serializers using a field depending on ES, like the feed does.

mkt/webapps/serializers.py
```diff
@@ -328,6 +338,14 @@ def save_upsold(self, obj, upsold):
             current_upsell.delete()
 
     def save_price(self, obj, price):
```
diox: Maybe add a comment explaining that this shouldn't be called for free apps, and the caller is responsible for checking this beforehand.

mkt/webapps/serializers.py
```diff
@@ -335,31 +353,17 @@ def save_price(self, obj, price):
...
-                                               for p in valid_prices),))
-        return attrs
-
-    def restore_object(self, attrs, instance=None):
-        # restore_object creates or updates a model instance, during
-        # input validation.
+        return device_types
+
+    def validate_price(self, price):
+            return {'price': price}
```
diox: Weird indent

requirements/prod.txt
```diff
@@ -41,7 +41,7 @@ django-filter==0.9.2
 django-multidb-router==0.5.1
 django-nose==1.4.2
 django-quieter-formset==0.4
-djangorestframework==2.4.5
+djangorestframework==3.3.2
```
diox: You should also be able to removed `drf_compound_fields` now

mkt/search/serializers.py
washort: yes, `.all()` is only called on django managers now: https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/serializers.py#L611

mkt/extensions/tests/test_validation.py
washort: this is a change in the exception API (`detail` is a dict containing the complete response info now), not in the json response sent for errors.

mkt/api/tests/__init__.py
washort: This silenced a bunch of "no more db access allowed in this transaction" errors in the tests.

mkt/api/paginator.py
washort: things that want to do ES pagination set `count_override` which prevents the count call, see https://github.com/mozilla/zamboni/pull/3469/files#diff-5819148273915141de41e0ae9738dab0R60

mkt/extensions/tests/test_validation.py
diox: ah right, carry on

mkt/api/paginator.py
diox: I don't see it on website search or extension search (also not on regular search view but apparently you kept `paginator_class = ESPaginator` there ? https://github.com/washort/zamboni/blob/restframework3/mkt/search/views.py#L55)

mkt/api/tests/__init__.py
diox: that's too weird to silence, IMHO. we should never need to add `transaction.atomic()` calls in tests.

mkt/account/serializers.py
washort: I just noticed this was in `validate`, not `to_representation`, which is part of why it looks so weird.

mkt/account/tests/test_views.py
washort: This is just a change in next/prev url generation, the 'offset' parameter isn't included when it's 0 since that is the default.

mkt/api/paginator.py
washort: This is a very good point and I realized that this can be done correctly from directly inside `CustomPagination`, getting `count` from `response.hits.total` when used on an ES search.

diox: Yeah. I think you should do this in both pagination class, and then remove all references to `ESPaginator` in the views themselves. There are tests in `mkt/search/views.py` that test only one ES query is made, you could steal this for the other ES views as well.

mkt/api/tests/__init__.py
diox: Discussed on IRC and it makes sense, DRF now rolls back the transaction in the exception handler if `ATOMIC_REQUESTS` is `True` (or inside a manual transaction), and we want that.

mkt/api/paginator.py
diox: Works with https://github.com/washort/zamboni/commit/ad2991c35328956adbfab7e6c44d3191ba6d86f2 that uses the total from ES like `ESPaginator` does. Nothing else to do here except maybe a comment now that you added tests as well in https://github.com/washort/zamboni/commit/7716bcbfd3e4ffc53065094e0103b69ac4775a1c

diox:
r+wc

PR #3470, opened 2015-12-28
Fix IARC v2 API URLs, and pass store id/password in each request (bug 1233385)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1233385


diox:
Superseded by #3476

PR #3471, opened 2015-12-28
Tried to resolve Bug [1202409]

waseem18:
Bug : https://bugzilla.mozilla.org/show_bug.cgi?id=1202409 
Removed 'thumbtype' fields which are of no use to the codebase.


mkt/webapps/migrations/0001_initial.py
```diff
@@ -252,7 +252,6 @@ class Migration(migrations.Migration):
                 ('created', models.DateTimeField(auto_now_add=True)),
                 ('modified', models.DateTimeField(auto_now=True)),
                 ('filetype', models.CharField(max_length=25)),
-                ('thumbtype', models.CharField(max_length=25)),
```
diox: Please revert this part - you shouldn't touch existing migrations. Instead, create a new one with `manage.py createmigrations` and add it with `git add`.

PR #3472, opened 2015-12-28
update dependencies (bug 1156928)

washort:


diox:
Seems like you are missing a few jingo.env still.

Got those deps on pyrepo yet ?

washort:
yes, pyrepo is updated -- next step is to replace tower with puente, looks like.

diox:
r+wc (https://github.com/washort/zamboni/commit/facb37e3004c7303a672225445aa0564fb28b31e#commitcomment-15245080)

Had a weird pip segmentation fault (!) locally but it worked anyway, might not be related.

PR #3475, opened 2015-12-30
Upgrade from Tower to Puente

washort:


diox:
It's working for me locally, with both python and javascript translations. I haven't tried extraction though.

r+

PR #3476, opened 2015-12-30
 Implement glue code between views/tasks/models and the new IARC Client (bug 1233385)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1233385


diox:
Could do a lot more tests in reviewer tools, like we do for V1, but that seem unnecessary for now, as it would mean running our slowest tests twice (once for v1, once for v2).

ngokevin:
r+ more tests than code :+1:

PR #3479, opened 2016-01-07
Fix multi-line strings and collapse whitespace following tower removal

diox:
tower used to fix those strings for us magically, removing extraneous whitespace, so now we have to do it ourselves and be careful :)

Follows https://github.com/mozilla/zamboni/commit/18a4f4cfaf7e21fbce5d3fd4ce60824242971d2b and https://github.com/mozilla/zamboni/commit/cf2cea4bcd2e56d3119ad9aca591162925613266

### DIFF
--- a/mkt/developers/templates/developers/api.html
+++ b/mkt/developers/templates/developers/api.html
@@ -25,7 +25,7 @@
           There are two main uses for the Marketplace API. From another web site
           using a traditional OAuth flow or from a command line tool.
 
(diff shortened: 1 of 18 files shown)


ddurst:
r+

ddurst:
My worry is that in a system where we assume that whitespace is handled for us (tower), we're gonna miss things like this in the future.

diox:
We could add a warning in the extract process, and/or ask translators to complain when that happens.

PR #3488, opened 2016-01-20
Don't queue tasks if we're not inside a request-response cycle (bug 1241071)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1241071


lib/post_request_task/task.py
```diff
@@ -69,12 +99,13 @@ def apply_async(self, *args, **kwargs):
...
+
+# Send the tasks to celery and stop queuing when the request is finished.
+request_finished.connect(_send_tasks_and_stop_queuing,
                          dispatch_uid='request_finished_tasks')
-# Also send the tasks when a task is finished (outside the request-response
-# cycle, when a task calls another task).
-task_postrun.connect(_send_tasks, dispatch_uid='tasks_finished_tasks')
 
 # And make sure to discard the task queue when we have an exception in the
 # request-response cycle.
```
robhudson: All the above changes make so much more sense. I like that we enable/disable this task queueing stuff with the request signals starting and stopping. Nicely done.

robhudson:
I mostly skimmed over the test changes. r+ I like the changes to the `post_request_task` lib.

diox:
Waiting a week before merging to make sure the Django 1.8 update sticks and also because we've had issues with the django-post-request-task app in olympia.

PR #3492, opened 2016-01-26
Bug 1236810 - Update TV template and manifest.webapp

jostw:
I would like to update some settings for TV Marketplace.

1. The url of locale files is incorrect now. Adding `data-media` and `data-repo` on body should fix this problem.
2. menu/menuitem has moved to the homepage template, so it should be removed from index. See this [Bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1241402) for more details.
3. Point to TV Marketplace icon.
4. Remove appcache.
5. Name should be 'Web Apps'.

Thanks :)


washort:
r+

jostw:
@washort Will you merge this for me?

PR #3493, opened 2016-01-27
Pass . to stats-po.sh in omg_new_l10n.sh so it works with BSD and gnu…

ddurst:
… find. (Previously done for all other string extraction repos except this one on 2015/12/03.)


ddurst:
I am now officially a member of the team. For better or for worse.

PR #3500, opened 2016-02-03
Bug 1236810 - Update TV Marketplace icon URL to production server

jostw:
Since the assets of TV Marketplace had push to the production server, we should change the icon url in manifest to the production server.


spasovski:
Test failures seem unrelated. r+

PR #3506, opened 2016-02-05
Minor tweaks to wording on content ratings edit page, also don't show initial cert_id

diox:
Showing the initial certificate ID value has no real benefit, we don't want users to resubmit that form anyway.


ngokevin:
r+

PR #3509, opened 2016-02-09
Fix tvplace search ordering

diox:
If `tv_featured` was absent the ordering was unstable. To fix it, sort by relevancy or by reviewed date depending on the query, and make sure to handle missing `tv_featured` value.


washort:
r+

PR #3512, opened 2016-02-15
Show a confirmation before attaching an IARC Certificate

diox:
Per spec : "present the ratings information to the developer and ask if this is the cert they want to use."


ngokevin:
r+

PR #3518, opened 2016-02-24
Add all PEGI categories/descriptors to Generic rating body

diox:
As requested by IARC:
- Generic has no records in Descriptor table, it uses the complete set of PEGI descriptors
- Generic has no records in AgeRating table either, it uses the complete set of PEGI age ratings, but for ageRatingID 55 it shows a 12+ icon, i.e. when PEGI shows an exclamation point icon Generic shows “12+”.


ddurst:
I'm r+ing this from looking through it. Also because this has gone on enough. :)

PR #3519, opened 2016-02-25
remove payments/devices in submission/devhub

washort:


diox:
(don't merge before today's tag, we still have IARC fixes that need to be pushed)

PR #3520, opened 2016-02-26
Fix "Generic" IARC rating body to PEGI categories... without the "+"

diox:
We sadly match on AgeRatingText because it was the most convenient with what we had when we started. It appears that for Generic, IARC rating no longer includes the "+" after the age, at least not always.


ddurst:
r+

PR #3523, opened 2016-03-01
Avoid 500 error when IARC returns invalid JSON, regardless of the reason

diox:


ddurst:
r+ (was waiting on checks)

PR #3530, opened 2016-03-17
Remove websites, disable unsupported apps (bugs 1243209, 1246179)

washort:


mkt/websites/migrations/0005_delete_websites.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+from mkt.constants.applications import DEVICE_TV
+
+
+def delete_websites(apps, schema_editor):
+    Website = apps.get_model("websites", "Website")
+    Website.objects.exclude(devices=[DEVICE_TV.id]).delete()
```
diox: Eh, I didn't expect that to actually work, but it does... Cool.

mkt/webapps/migrations/0013_disable_unsupported_apps.py
```diff
@@ -0,0 +1,23 @@
...
+
+from django.db import migrations
+from mkt.constants.applications import DEVICE_GAIA
+from mkt.constants.base import ADDON_FREE, STATUS_NULL
+
+
+def disable_unsupported_apps(apps, schema_editor):
+    Webapp = apps.get_model("webapps", "Webapp")
+    Webapp.objects.exclude(addondevicetype__device_type=DEVICE_GAIA.id,
+                           premium_type=ADDON_FREE).update(status=STATUS_NULL)
```
diox: You'll be selecting add-ons that are not free and not gaia compatible at the same time, this is not what we want.

diox: I think something like this would work ? `Webapp.objects.only('id').exclude(Q(addondevicetype__device_type=DEVICE_GAIA.id) | Q(premium_type=ADDON_FREE))`

Or just make 2 queries maybe, so it's easier to read ?

mkt/websites/migrations/0005_delete_websites.py
diox: Just to be clear : at the moment TV websites *only* have the TV device set, and no other ?

washort: That's correct.

PR #3532, opened 2016-03-18
In GetRatingChanges, only update descriptors from bodies that changed (bug 1257677)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1257677


ddurst:
r+

PR #3533, opened 2016-03-21
Replace RemoveProduct by Unpublish, per latest IARC spec (bug 1257248)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1257248


ddurst:
r+

PR #3534, opened 2016-03-24
Renovate landing page (bug 1247035)

washort:
nu-landing

Does this look right? @spasovski @dbialer


spasovski:
Looks good r+ from my end

PR #3536, opened 2016-09-13
Remove links to submit new apps to Marketplace (bug 1299561)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1299563


diox:
To keep my sanity I'm not touching the views, so technically it's still possible to submit if you know the URL... but it'd be pointless since you'd end up in a review queue that nobody cares about anymore.

PR #3542, opened 2016-12-19
Remove solitude auth check from monitor Bug 1311724

jasonthomas:
Before we decommission solitude we need to remove this monitor.

r?


ddurst:
r+

PR #3543, opened 2016-12-19
Remove solitude reference from view and template

jasonthomas:
I missed this in https://github.com/mozilla/zamboni/pull/3542

r?


ddurst:
r+

